PR #118, opened 2026-05-02
fix(skills): unbreak skills.sh autocomplete in Add-skill dialog

ljagiello:
## Summary

Two bugs in the skills.sh autocomplete when adding a skill via the Add-skill dialog.

### 1. Clicking a suggestion below the dialog box closed the modal
The `onmousedown` handler ran `pickSuggestion` synchronously, which unmounted the `<ul>` before the user lifted the mouse. `mouseup` then landed on whatever was now under the cursor — for items rendered visually below the dialog, that was an element outside the dialog, and Melt UI's debounced `interactOutside` detection treated it as an outside-click and closed the modal.

**Fix:** split the handler — `onmousedown` only prevents focus loss, `onclick` runs `pickSuggestion`. The original button stays the event target across mousedown/mouseup/click. Once it detaches, Melt's `isValidEvent` short-circuits (target no longer in document) and the dialog stays open.

### 2. Typing `owner/partial` (e.g. `ljagiello/c`) returned no suggestions
skills.sh search is free-text over skill names — it doesn't honour `owner/partial` as a prefix filter, and short suffixes get buried by unrelated matches.

**Fix:** when the query contains a slash, send the owner (first segment) to skills.sh so the response is scoped to that owner, then narrow client-side by the full query prefix.

Same fixes applied to the inline install row on the workspace skills page (`/platform/[workspaceId]/skills`) for consistency.

## Test plan
- [x] `ljagiello` → 8 ctf-skills suggestions appear
- [x] Click bottom item (`solve-challenge`, rendered visually below dialog box) → input populates, modal stays open
- [x] Click any item visually inside dialog box → input populates, modal stays open (no regression)
- [x] `ljagiello/c` → 8 ctf-skills suggestions (single char after slash)
- [x] `ljagiello/x` → empty list (no ljagiello skill matches)
- [x] `ljagiello/ctf-skills/ctf-cry` (3 segments) → no suggestions, treated as direct ref
- [x] All scenarios verified in Chrome against a running daemon at localhost:8080

### DIFF
--- a/tools/agent-playground/src/lib/components/skills/skills-sh-import.svelte
+++ b/tools/agent-playground/src/lib/components/skills/skills-sh-import.svelte
@@ -42,17 +42,32 @@
     }, 200);
   }
 
+  /**
+   * skills.sh search is free-text over skill names — it doesn't honour
+   * `owner/partial` as a prefix filter, and short suffixes like `c` get
+   * buried by unrelated matches. Once the user types a slash, send the
+   * owner (first segment) to skills.sh so the response is scoped to
+   * that owner, then narrow client-side by the full query prefix.
+   */
+  const segs = $derived(searchQuery.split("/").filter(Boolean));
+  const searchTerm = $derived(segs[0] ?? "");
   const suggestionsQuery = createQuery(() =>
     queryOptions({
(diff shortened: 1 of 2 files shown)


basedfriday:
LGTM

PR #165, opened 2026-05-03
fix(cli): load launcher .env so compiled binary finds daemon on correct port

Vpr99:
## What

The compiled `friday` binary (deno-compile'd from `apps/atlas-cli`) starts with a fresh `process.env`. The launcher writes `FRIDAYD_URL=http://localhost:18080` to `~/.friday/local/.env` and imports it into child processes, but a standalone binary launched from the shell never sees those variables. Every command routing through `getAtlasDaemonUrl()` falls back to the hardcoded `:8080`, producing `Connection refused (os error 61)` when the daemon is actually on `:18080`.

## Changes

- `apps/atlas-cli/src/cli.ts`: Load `~/.friday/local/.env` into `process.env` at the entry point, before any command runs. Only injects keys not already present — shell exports still win.
- `apps/atlas-cli/package.json`: Add `dotenv` dependency (replacing `@std/dotenv` per codebase preference).

## Verified

- `deno run ./apps/atlas-cli/src/otel-bootstrap.ts workspace ls` → connects to `:18080`, lists workspaces ✅
- Old installed binary (`~/.friday/local/bin/friday workspace ls`) → still fails on `:8080` ✅ (confirms bug is in the build)
- Compiled binary (`deno compile …`) → works on `:18080` ✅
- Shell export override (`FRIDAYD_URL=http://localhost:9999`) → wins over `.env` ✅

## Files Changed

- `apps/atlas-cli/src/cli.ts`
- `apps/atlas-cli/package.json`

### DIFF
--- a/apps/atlas-cli/package.json
+++ b/apps/atlas-cli/package.json
@@ -13,6 +13,7 @@
     "@atlas/oapi-client": "workspace:*",
     "@atlas/storage": "workspace:*",
     "@atlas/utils": "workspace:*",
+    "dotenv": "^16.4.7",
     "@inkjs/ui": "^2.0.0",
     "@std/dotenv": "npm:@jsr/std__dotenv@^0.225.5",
     "@std/yaml": "npm:@jsr/std__yaml@^1.0.10",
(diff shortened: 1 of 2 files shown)


basedfriday:
You might want to base this into my branch.

PR #176, opened 2026-05-04
fix(artifacts): persist correct mime for agent-created files

### LINKED ISSUE #173: Skill/markdown attachments saved as application/octet-stream, download as .bin
## Summary

Markdown files (e.g. `SKILL.md`) registered via `artifacts_create` from a chat session are persisted with `mimeType: application/octet-stream` instead of `text/markdown`. The chat UI shows an `octet-stream` badge, and clicking **Download** or **Open** lands the file as `SKILL.bin` (e.g. `~/Downloads/SKILL (1).bin`).

Reproduced in chat `chat_e9ZqAjYdCz` while a skill-authoring agent wrote `workspace-sparring SKILL.md` to scratch and called `artifacts_create`.

| Field | Observed | Expected |
| ----- | -------- | -------- |
| Stored mimeType | `application/octet-stream` | `text/markdown` |
| Badge in artifact card | `octet-stream` | `markdown` (or similar) |
| Downloaded filename | `workspace-sparring SKILL.bin` | `workspace-sparring SKILL.md` |

## Repro

1. In a chat session, have the agent write a markdown file to scratch (e.g. `write_file("SKILL.md", "...")`).
2. Have the agent call `artifacts_create({ path: "SKILL.md", title: "...", summary: "..." })`.
3. In the chat UI, click **Download** (or **Open**) on the artifact card.
4. Observe the file lands as `*.bin`.

Any text artifact (`.md`, `.txt`, `.json`, `.csv`, `.yaml`, …) reproduces — anything without binary magic bytes.

## Root cause

The bug is in the agent-side artifact creation path; the storage layer's fallback chain then turns a missing hint into a wrong-but-frozen MIME type.

### The chain

**1. `artifacts_create` never sends a `mimeType` hint** — `packages/system/agents/workspace-chat/tools/artifact-tools.ts:152-157`

```ts
const data = {
  type: "file" as const,
  content: base64,
  contentEncoding: "base64" as const,
  originalName: basename(path),
  // ← mimeType missing, even though basename(path) is right here
};
```

The tool has the filename in hand but never infers from it.

**2. Storage falls back to magic-byte sniffing → octet-stream for text** — `packages/core/src/artifacts/jetstream-adapter.ts:70-78`

```ts
let mimeType: string;
if (input.mimeType) {
  mimeType = input.mimeType;
} else {
  const sniffed = await fileTypeFromBuffer(bytes);
  mimeType = sniffed?.mime ?? "application/octet-stream";
}
```

`fileTypeFromBuffer` only detects formats with binary magic bytes (PDF, PNG, JPEG, ZIP, …). Markdown/JSON/CSV/YAML have none, so `sniffed` is `undefined` and the record is persisted with `application/octet-stream`.

**3. Download route faithfully serves `.bin`** — `apps/atlasd/routes/artifacts.ts:765-838`

The route calls `deriveDownloadFilename({ mimeType, originalName, title })`. Because `extFromMime("application/octet-stream")` returns `"bin"` (`packages/core/src/artifacts/file-upload.ts:206`), `deriveDownloadFilename` (lines 255-272) **rewrites** the original `.md` extension to `.bin`, then sets:

- `Content-Type: application/octet-stream`
- `Content-Disposition: attachment; filename="workspace-sparring SKILL.bin"`

The browser obeys.

**4. UI badge** — `tools/agent-playground/src/lib/components/chat/artifact-card.svelte:177-178` shows `octet-stream` because it reads the (incorrect) stored mime. Same root cause, just the visible symptom.

### Why PR #169 didn't catch this

#169 ("Chat attachment lifecycle: scrubber + parse_artifact + persistence drain") added base64 magic-byte sniffing in the scrubber for **binary** artifacts coming from MCP tools (Gmail PDFs, images). Text artifacts authored directly via `artifacts_create` were never in scope, and they have no magic bytes to sniff anyway.

### What's already correct

- `packages/core/src/artifacts/file-upload.ts:38-63` — `EXTENSION_TO_MIME` map already maps `.md → text/markdown`, plus `.txt`, `.json`, `.csv`, etc.
- `packages/core/src/artifacts/file-upload.ts:274-279` — `getValidatedMimeType(fileName)` exists and returns the right thing. It's just **never called** from `artifacts_create`.
- `deriveDownloadFilename` and the download route are doing exactly what they're told; they're not buggy, they're being fed bad data.

## Proposed fix (two layers)

**1. Root cause — set the hint at the caller** (`packages/system/agents/workspace-chat/tools/artifact-tools.ts:152-157`):

```ts
import { getValidatedMimeType } from "@atlas/core/artifacts/file-upload";

const filename = basename(path);
const inferredMime = getValidatedMimeType(filename);
const data = {
  type: "file" as const,
  content: base64,
  contentEncoding: "base64" as const,
  originalName: filename,
  ...(inferredMime ? { mimeType: inferredMime } : {}),
};
```

**2. Defense-in-depth — fall back to extension inference in the storage layer** (`packages/core/src/artifacts/jetstream-adapter.ts:70-78`):

```ts
let mimeType: string;
if (input.mimeType) {
  mimeType = input.mimeType;
} else {
  const sniffed = await fileTypeFromBuffer(bytes);
  if (sniffed?.mime) {
    mimeType = sniffed.mime;
  } else if (input.originalName) {
    mimeType = getValidatedMimeType(input.originalName) ?? "application/octet-stream";
  } else {
    mimeType = "application/octet-stream";
  }
}
```

This makes any future caller that forgets `mimeType` still produce the right result whenever the filename has a known extension.

## Migration / existing artifacts

Existing artifact rows already have `mimeType: application/octet-stream` baked in. Two options:

1. **Leave them.** Acceptable if the bug is recent (post-#169) and few skill artifacts exist.
2. **Backfill.** One-shot script: for each artifact where `mimeType === "application/octet-stream"` AND `originalName` has a known extension in `EXTENSION_TO_MIME`, update `mimeType`. Low-risk; `originalName` is preserved verbatim.

Recommend (2) if the count is non-trivial; the script is a few lines.

## Tests to add

- `artifacts_create` with a `.md` scratch file → stored `mimeType` is `text/markdown`.
- `artifacts_create` with `.txt`, `.json`, `.csv` → matching mime types.
- `artifacts_create` with an unknown extension (`.xyz`) → falls back to `application/octet-stream` (and download still serves a sensible filename).
- `materializeBlob` unit test: input with `originalName` ending in `.md` and no `mimeType` → returns `text/markdown`.

## File:line summary

| Role | Path | Lines |
| ---- | ---- | ----- |
| Bug — caller misses hint | `packages/system/agents/workspace-chat/tools/artifact-tools.ts` | 152-157 |
| Bug — storage default | `packages/core/src/artifacts/jetstream-adapter.ts` | 70-78 |
| Helper (unused by caller) | `packages/core/src/artifacts/file-upload.ts` | 38-63, 274-279 |
| Download route (correct, fed bad data) | `apps/atlasd/routes/artifacts.ts` | 765-838 |
| Filename derivation | `packages/core/src/artifacts/file-upload.ts` | 206, 229-272 |
| UI badge (symptom) | `tools/agent-playground/src/lib/components/chat/artifact-card.svelte` | 177-178 |

## Related

- #169 — Chat attachment lifecycle: scrubber + parse_artifact + persistence drain (introduced the surrounding lifecycle work; this case slipped through).

Vpr99:
Closes #173.

## Summary

Agent-side `artifacts_create` was registering scratch files with no `mimeType` hint. The storage layer's magic-byte sniff (`fileTypeFromBuffer`) is binary-only — text formats (md/json/csv/yaml/html/source code/...) have no magic bytes, so records persisted as `application/octet-stream`. The download route then rewrote `originalName`'s extension to `.bin`, so a `SKILL.md` artifact downloaded as `SKILL.bin`.

## Root cause

The original bug was a wrong default in `deriveDownloadFilename`: it computed the extension from the stored `mimeType` and overwrote `originalName`'s extension to match. That made every stored mime an implicit contract — it had to round-trip cleanly through `extFromMime` or the user got a renamed file. The `extFromMime` fallback uses a `[a-z0-9]+` regex on the mime's tail, which silently maps any hyphenated mime tail to `bin`. So even after layer-1 mime inference, ~12 of 22 inferred mimes would still rewrite incorrectly: `text/x-typescript` → `script.bin`, `text/javascript` → `app.javascript`, `text/x-python` → `main.bin`, etc.

Two-layer band-aids (caller-side inference + an octet-stream guard) didn't fix the contract — they just narrowed the failure window.

## Fix

**Flip the trust hierarchy in `deriveDownloadFilename`.** `originalName` is the source of truth when its extension agrees with the stored `mimeType` (or when the mime is `application/octet-stream`, meaning we have no better signal). The only unconditional override is the scrubber path: when the scrubber stamps `<tool>-<ts>.bin` before sniffing the mime, we swap `.bin` for the real extension once known. If the stored MIME proves the extension is wrong (e.g. PNG bytes uploaded as `misnamed.txt`), we rewrite to match the MIME.

Result: filename round-trip is automatic for every correctly-named text artifact, including ones that don't round-trip through `extFromMime`. Mismatched binary uploads still get corrected.

**Collapse the two extension→mime maps into one with a flag.** `EXTENSION_TO_MIME` is now `Map<string, { mime, uploadable }>`. `uploadable: true` entries gate `getValidatedMimeType` and `ALLOWED_EXTENSIONS`; `uploadable: false` entries are agent-side inferences only — markup/source/config text formats. Adding such an extension does NOT widen the UI upload allowlist. One source of truth, one flag, no duplicate map to keep in sync.

**Only infer text-ish MIME at `artifacts_create` time.** The caller-side helper now stamps `mimeType` only when `isTextMimeType(inferredMime)` is true. Binary/image/audio files are left for the storage layer's magic-byte sniff so filename alone can't mislabel them.

**Sandbox HTML artifact previews.** Because `.html` is now inferred as `text/html` and rendered inline, both the playground iframe (`sandbox=""`) and the daemon content route (`Content-Security-Policy: sandbox; default-src 'none'; img-src data: blob:; style-src 'unsafe-inline'`) are locked down so agent-written HTML can't execute as same-origin active content.

## Manual testing

To reproduce on your machine:

1. Run the playground against this branch: `deno task dev:playground`.
2. Open a chat, send: *"Write me a brief short story in a markdown file and call artifacts_create on it."*
3. The agent will run `write_file` → `artifacts_create` → `display_artifact`. Observe the artifact card in the chat:
   - **Badge** should say `markdown`, not `octet-stream`.
   - **Click Download** — file lands as `<title>.md`, not `<title>.bin`.
   - **Click Open** — opens with the correct viewer for markdown rather than triggering a binary download.
4. Repeat with other formats:
   - `.html` (e.g. *"write a one-page HTML status report and create an artifact"*) — badge says `html`, downloads as `.html`, previews inline in a sandboxed iframe.
   - `.ts` / `.py` / `.toml` etc. — badges show generic labels (`x-typescript`, `x-python`, `x-toml`) but downloads land with the right extension. *(This is the case Ken's review caught — `.ts` previously downloaded as `.bin` despite the layer-1 inference.)*
   - An extension outside the map (e.g. `.foo`) — badge says `octet-stream` (storage layer fell through), and the **download still preserves `.foo`** because the new logic trusts `originalName` when there is no better MIME signal.
5. Sanity-check the stored mime via the daemon API:
   ```sh
   curl -s http://localhost:8080/api/artifacts/<artifactId> | jq '.artifact.data | {mimeType, originalName}'
   ```
   Expected for a markdown file: `{ "mimeType": "text/markdown", "originalName": "story.md" }`.

## Test Plan

- [x] Unit test: `artifacts_create` infers `text/markdown` for `.md` filename
- [x] Unit test: `artifacts_create` infers `text/html` for `.html` filename (extension outside upload allowlist)
- [x] Unit test: `artifacts_create` omits `mimeType` for unknown extension (storage falls back to magic-byte sniff)
- [x] Unit test: `artifacts_create` omits `mimeType` for binary extensions (storage sniffs bytes instead of trusting filename)
- [x] Parametrized round-trip test: `deriveDownloadFilename` preserves `originalName` for every entry in the inferred map (25 cases — `.ts`, `.py`, `.go`, `.rs`, `.sh`, `.bash`, `.sql`, `.toml`, `.tsv`, `.js`, `.jsx`, `.mjs`, `.cjs`, `.tsx`, `.html`, `.htm`, `.xml`, `.svg`, `.css`, `.ini`, `.conf`, `.log`, `.md`, `.csv`, `.json`)
- [x] Unit test: `deriveDownloadFilename` rewrites `.bin` originalName when real mime known (no scrubber-path regression)
- [x] Unit test: `deriveDownloadFilename` rewrites proven-mismatched extension (e.g. PNG stored with `.txt` originalName → `.png`)
- [x] Unit test: `deriveDownloadFilename` appends extension when `originalName` has none
- [x] Unit test: `deriveDownloadFilename` falls back to `<title>.<ext>` when no `originalName`
- [x] Unit test: `inferMimeFromFilename` returns mime for both uploadable and agent-only extensions; `undefined` for unknown
- [x] Unit test: `getValidatedMimeType` returns `undefined` for non-uploadable extensions (gates UI uploads)
- [x] Unit test: `ALLOWED_EXTENSIONS` excludes agent-only inferred extensions
- [x] Route test: `text/html` artifact content response carries `Content-Security-Policy: sandbox` header
- [x] All 958 tests in related packages continue to pass
- [x] `deno check` clean across modified files + downstream callers (`tools/agent-playground/src/lib/upload.ts` adjusted for new map shape)
- [x] `svelte-check` clean (0 errors) in playground after adding `sandbox=""` to HTML iframe

### DIFF
--- a/apps/atlasd/routes/artifacts.test.ts
+++ b/apps/atlasd/routes/artifacts.test.ts
@@ -773,6 +773,39 @@ describe("Content endpoint", () => {
     expect(htmlCsp).toContain("allow-scripts");
   });
 
+  it("sandboxes XHTML artifact content the same as HTML", async () => {
+    // XHTML is rendered as a document and executes `<script>` exactly like
+    // text/html. Without the sandbox the embedded script would run
+    // same-origin against the daemon.
+    const createResponse = await artifactsApp.request("/", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        title: "XHTML report",
+        summary: "XHTML report with active content",
+        data: {
+          type: "file",
+          content:
+            '<?xml version="1.0"?><html xmlns="http://www.w3.org/1999/xhtml"><body><script>parent.document.body.dataset.pwned = "true"</script></body></html>',
+          mimeType: "application/xhtml+xml",
+          originalName: "report.xhtml",
+        },
+      }),
+    });
+    expect(createResponse.status).toEqual(201);
+    const { artifact } = ArtifactResponseSchema.parse(await createResponse.json());
+
+    const contentResponse = await artifactsApp.request(`/${artifact.id}/content`, {
+      method: "GET",
+    });
+
+    expect(contentResponse.status).toEqual(200);
+    expect(contentResponse.headers.get("Content-Type")).toEqual("application/xhtml+xml");
+    const xhtmlCsp = contentResponse.headers.get("Content-Security-Policy") ?? "";
+    expect(xhtmlCsp).toContain("sandbox");
+    expect(xhtmlCsp).toContain("allow-scripts");
+  });
+
   it("sandboxes SVG artifact content (X-Content-Type-Options does not stop SVG scripts)", async () => {
     // SVG served same-origin can execute embedded `<script>` regardless
     // of `nosniff`. The CSP sandbox directive is the only thing that
--- a/apps/atlasd/routes/artifacts.test.ts
+++ b/apps/atlasd/routes/artifacts.test.ts
@@ -767,6 +767,37 @@ describe("Content endpoint", () => {
     expect(contentResponse.headers.get("Content-Type")).toEqual("text/html");
     expect(contentResponse.headers.get("Content-Security-Policy")).toContain("sandbox");
   });
+
+  it("sandboxes SVG artifact content (X-Content-Type-Options does not stop SVG scripts)", async () => {
+    // SVG served same-origin can execute embedded `<script>` regardless
+    // of `nosniff`. The CSP sandbox directive is the only thing that
+    // blocks it — ship it on `image/svg+xml` for parity with text/html.
+    const createResponse = await artifactsApp.request("/", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        title: "SVG icon",
+        summary: "SVG with embedded script that should be sandboxed",
+        data: {
+          type: "file",
+          content:
+            '<svg xmlns="http://www.w3.org/2000/svg"><script>parent.document.body.dataset.pwned = "true"</script></svg>',
+          mimeType: "image/svg+xml",
+          originalName: "icon.svg",
+        },
+      }),
+    });
+    expect(createResponse.status).toEqual(201);
+    const { artifact } = ArtifactResponseSchema.parse(await createResponse.json());
+
+    const contentResponse = await artifactsApp.request(`/${artifact.id}/content`, {
+      method: "GET",
+    });
+
+    expect(contentResponse.status).toEqual(200);
+    expect(contentResponse.headers.get("Content-Type")).toEqual("image/svg+xml");
+    expect(contentResponse.headers.get("Content-Security-Policy")).toContain("sandbox");
+  });
 });
 
 describe("PDF upload integration", () => {
--- a/apps/atlasd/routes/artifacts.test.ts
+++ b/apps/atlasd/routes/artifacts.test.ts
@@ -765,7 +765,12 @@ describe("Content endpoint", () => {
 
     expect(contentResponse.status).toEqual(200);
     expect(contentResponse.headers.get("Content-Type")).toEqual("text/html");
-    expect(contentResponse.headers.get("Content-Security-Policy")).toContain("sandbox");
+    // HTML gets `allow-scripts` so legit agent-rendered pages (Leaflet,
+    // charts, embedded viewers) can run. The opaque-origin sandbox is
+    // what isolates them from the parent.
+    const htmlCsp = contentResponse.headers.get("Content-Security-Policy") ?? "";
+    expect(htmlCsp).toContain("sandbox");
+    expect(htmlCsp).toContain("allow-scripts");
   });
 
   it("sandboxes SVG artifact content (X-Content-Type-Options does not stop SVG scripts)", async () => {
@@ -796,7 +801,11 @@ describe("Content endpoint", () => {
 
     expect(contentResponse.status).toEqual(200);
     expect(contentResponse.headers.get("Content-Type")).toEqual("image/svg+xml");
-    expect(contentResponse.headers.get("Content-Security-Policy")).toContain("sandbox");
+    // SVG renders as an image — no `<script>` should ever execute.
+    // Sandbox stays on, but `allow-scripts` is intentionally absent.
+    const svgCsp = contentResponse.headers.get("Content-Security-Policy") ?? "";
+    expect(svgCsp).toContain("sandbox");
+    expect(svgCsp).not.toContain("allow-scripts");
   });
 });
 
--- a/apps/atlasd/routes/artifacts.test.ts
+++ b/apps/atlasd/routes/artifacts.test.ts
@@ -740,6 +740,33 @@ describe("Content endpoint", () => {
     expect(cd.split(";")[0]?.trim()).toEqual("attachment");
     expect(cd).toContain('filename="data.json"');
   });
+
+  it("sandboxes inline HTML artifact content", async () => {
+    const createResponse = await artifactsApp.request("/", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        title: "HTML report",
+        summary: "HTML report with active content",
+        data: {
+          type: "file",
+          content: "<script>parent.document.body.dataset.pwned = 'true'</script>",
+          mimeType: "text/html",
+          originalName: "report.html",
+        },
+      }),
+    });
+    expect(createResponse.status).toEqual(201);
+    const { artifact } = ArtifactResponseSchema.parse(await createResponse.json());
+
+    const contentResponse = await artifactsApp.request(`/${artifact.id}/content`, {
+      method: "GET",
+    });
+
+    expect(contentResponse.status).toEqual(200);
+    expect(contentResponse.headers.get("Content-Type")).toEqual("text/html");
+    expect(contentResponse.headers.get("Content-Security-Policy")).toContain("sandbox");
+  });
 });
 
 describe("PDF upload integration", () => {
--- a/apps/atlasd/routes/artifacts.test.ts
+++ b/apps/atlasd/routes/artifacts.test.ts
@@ -806,6 +806,78 @@ describe("Content endpoint", () => {
     expect(xhtmlCsp).toContain("allow-scripts");
   });
 
+  it("sandboxes HTML artifact content even when mime carries a charset parameter", async () => {
+    // Storage adapters that round-trip text mimes with a charset (e.g.
+    // `text/html; charset=utf-8`) used to bypass the sandbox because the
+    // route keyed the CSP gate off raw `mimeType` equality. The browser
+    // still parses such responses as HTML, so the sandbox MUST apply.
+    const createResponse = await artifactsApp.request("/", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        title: "HTML with charset",
+        summary: "HTML report whose stored mime carries a charset parameter",
+        data: {
+          type: "file",
+          content: "<script>parent.document.body.dataset.pwned = 'true'</script>",
+          mimeType: "text/html; charset=utf-8",
+          originalName: "report.html",
+        },
+      }),
+    });
+    expect(createResponse.status).toEqual(201);
+    const { artifact } = ArtifactResponseSchema.parse(await createResponse.json());
+
+    const contentResponse = await artifactsApp.request(`/${artifact.id}/content`, {
+      method: "GET",
+    });
+
+    expect(contentResponse.status).toEqual(200);
+    // Response Content-Type preserves the charset for the browser.
+    expect(contentResponse.headers.get("Content-Type")).toEqual("text/html; charset=utf-8");
+    const csp = contentResponse.headers.get("Content-Security-Policy") ?? "";
+    expect(csp).toContain("sandbox");
+    expect(csp).toContain("allow-scripts");
+  });
+
+  it("does not set a CSP for benign mimes (PNG)", async () => {
+    // Locks the contract that the CSP sandbox is mime-keyed: ordinary
+    // image/audio/document responses must not inherit the active-content
+    // restrictions, otherwise a future refactor that broadens the gate
+    // would ship without breaking the HTML/SVG tests.
+    const pngBytes = new Uint8Array([
+      0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a, 0x00, 0x00, 0x00, 0x0d, 0x49, 0x48, 0x44,
+      0x52, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x01, 0x08, 0x06, 0x00, 0x00, 0x00, 0x1f,
+      0x15, 0xc4, 0x89, 0x00, 0x00, 0x00, 0x0d, 0x49, 0x44, 0x41, 0x54, 0x78, 0x9c, 0x62, 0x00,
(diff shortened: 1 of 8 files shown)


packages/core/src/artifacts/file-upload.ts
```diff
@@ -280,3 +286,55 @@ export function getValidatedMimeType(fileName: string): string | undefined {
...
+/**
+ * Additional extension→mime mappings for *agent-side inference only*.
+ *
+ * Deliberately separate from `EXTENSION_TO_MIME`, which doubles as the
+ * UI upload allowlist (`ALLOWED_EXTENSIONS`). Agents commonly write
+ * source code, markup, and config files to scratch and register them as
+ * artifacts; without an inferred mime they persist as
+ * `application/octet-stream` and download as `.bin`.
+ */
+const INFERRED_TEXT_EXTENSION_TO_MIME = new Map([
```
LissaGreense: most of these mimes don't round-trip through `extFromMime`, so `script.ts` still downloads as `script.bin`, `script.js` as `script.javascript`, `server.log` as `server.txt`. The octet-stream guard doesn't catch it since the stored mime isn't octet-stream anymore. Can we add the matching entries to `MIME_TO_EXT`?

packages/core/src/artifacts/file-upload.ts
```diff
@@ -237,46 +301,79 @@ export function extFromMime(mimeType: string): string {
...
+ * the stored mime carries no better information.
  */
 export function deriveDownloadFilename(opts: {
   mimeType: string;
   originalName?: string;
   title: string;
 }): string {
-  const ext = extFromMime(opts.mimeType);
+  // Strip mime parameters (`; charset=…`) up front so the equality
+  // check against the filename-inferred mime survives storage adapters
```
Vpr99, line 333: **INFO** — Scrubber .bin path is a no-op when mime is octet-stream — worth a comment

**Category:** correctness

The `.bin` early-return fires unconditionally before the octet-stream check. If `baseMime` is `application/octet-stream`, `extFromMime` returns `'bin'` and the replacement is `tool.bin → tool.bin` — correct, but non-obvious. A one-line comment like `// extFromMime returns 'bin' for octet-stream, so this is a no-op when mime is unknown — scrubber file keeps .bin` would help the next reader.

packages/core/src/artifacts/file-upload.ts
```diff
@@ -29,37 +29,82 @@ export const LEGACY_FORMAT_ERRORS = new Map([
...
+  [".jpeg", { mime: "image/jpeg", uploadable: true }],
+  [".webp", { mime: "image/webp", uploadable: true }],
+  [".gif", { mime: "image/gif", uploadable: true }],
+  // Uploadable: audio
+  [".mp3", { mime: "audio/mpeg", uploadable: true }],
+  [".mp4", { mime: "audio/mp4", uploadable: true }],
+  [".m4a", { mime: "audio/x-m4a", uploadable: true }],
+  [".wav", { mime: "audio/wav", uploadable: true }],
+  [".webm", { mime: "audio/webm", uploadable: true }],
+  [".ogg", { mime: "audio/ogg", uploadable: true }],
```
Vpr99, line 79: **WARNING** — SVG EXTENSION_TO_MIME entry silently depends on TEXT_MIME_EXACT — coupling not documented

**Category:** correctness

`artifact-tools.ts` stamps `mimeType` only when `isTextMimeType(inferredMime)` is true. `image/svg+xml` is included in `TEXT_MIME_EXACT` (added in this PR) so the CSP sandbox fires correctly. But the coupling is implicit: someone adding a new active-content mime to EXTENSION_TO_MIME in the future may forget to also add it to TEXT_MIME_EXACT, causing it to silently fall through to the storage sniff, store as `octet-stream`, and miss the sandbox. A comment on the `.svg` entry pointing to the TEXT_MIME_EXACT dependency would close this gap.

```suggestion
// Also in TEXT_MIME_EXACT — isTextMimeType must return true so artifact-tools stamps
// the mime at create time; without it the /content CSP sandbox key misses and
// an SVG <script> executes same-origin.
```

packages/core/src/artifacts/file-upload.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  // Round-trip lock: every mime the agent-side inference can stamp on
+  // an artifact must result in a download filename that preserves the
+  // original extension. This blocks the regression Ken's reviewer
+  // surfaced — ~12 of 22 inferred mimes failed to round-trip through
+  // extFromMime under the old logic.
+  it.each([
+    ["text/html", "page.html"],
+    ["text/html", "page.htm"],
+    ["application/xml", "feed.xml"],
```
Vpr99, line 100: **INFO** — charset test only covers text/markdown — add a hyphenated mime case

**Category:** test-coverage

The charset-parameter test covers `notes.md` + `text/markdown; charset=utf-8`. Hyphenated mimes like `text/x-typescript; charset=utf-8` are the most likely to be normalized with a charset by a storage adapter, and they're exactly where the `.split(';')[0]` fix matters most. Adding one case for `script.ts` + `text/x-typescript; charset=utf-8` would lock in that behavior.

apps/atlasd/routes/artifacts.test.ts
```diff
@@ -740,6 +740,64 @@ describe("Content endpoint", () => {
...
+    const { artifact } = ArtifactResponseSchema.parse(await createResponse.json());
+
+    const contentResponse = await artifactsApp.request(`/${artifact.id}/content`, {
+      method: "GET",
+    });
+
+    expect(contentResponse.status).toEqual(200);
+    expect(contentResponse.headers.get("Content-Type")).toEqual("image/svg+xml");
+    expect(contentResponse.headers.get("Content-Security-Policy")).toContain("sandbox");
+  });
```
Vpr99, line 882: **INFO** — Sandbox tests assert toContain('sandbox') — consider asserting the full CSP string

**Category:** security

Both sandbox tests check `toContain('sandbox')` but don't assert the full policy value. A tighter `toBe('sandbox; default-src \'none\'; img-src data: blob:; style-src \'unsafe-inline\'')` would catch accidental future weakening of the policy (e.g. an `allow-scripts` addition). Low stakes since the route logic is nearby, but cheap to harden.

tools/agent-playground/src/lib/upload.ts
```diff
@@ -47,7 +47,7 @@ export function validateFile(file: File): { valid: true } | { valid: false; erro
     return { valid: false, error: `${ext.slice(1).toUpperCase()} files must be under ${maxMB}MB.` };
   }
 
-  const mimeForExt = EXTENSION_TO_MIME.get(ext);
+  const mimeForExt = EXTENSION_TO_MIME.get(ext)?.mime;
```
Vpr99, line 50: **INFO** — One-liner .mime accessor has no dedicated unit test

**Category:** coverage

The `EXTENSION_TO_MIME.get(ext)?.mime` change is correct and covered indirectly by ALLOWED_EXTENSIONS tests, but a direct unit test for `validateFile` with an image extension would lock in the behavior and make the refactor safer against future map shape changes.

Vpr99 (commented):
## Code Review

**Verdict:** APPROVE

### Summary

Solid fix — the trust-hierarchy flip in `deriveDownloadFilename` is the right design. `originalName` as source of truth for text artifacts is cleaner than the old MIME-driven extension rewrite, the scrubber `.bin` repair path still works, and the `uploadable` flag elegantly collapses two maps into one without widening the UI upload allowlist.

All 8 files reviewed across the full diff:
- `file-upload.ts` — core logic change + EXTENSION_TO_MIME refactor ✅
- `file-upload.test.ts` (new) — good parametrized coverage of 25 round-trip cases and charset-parameter edge case ✅
- `artifact-tools.ts` — minimal, correct caller-side change ✅
- `artifact-tools.test.ts` (new) — covers the four key cases including the SVG CSP path ✅
- `artifacts.ts` (route) — CSP sandbox for `text/html` and `image/svg+xml` ✅
- `artifacts.test.ts` — two new sandbox tests ✅
- `artifact-card.svelte` — `sandbox=""` on the playground iframe ✅
- `upload.ts` — `.mime` accessor fix ✅

A few observations inline — mostly docs/test hardening, one worth flagging (the `image/svg+xml` → `TEXT_MIME_EXACT` dependency chain).

---

> 5 findings: 5 inline

---

*Automated review by Friday*

ljagiello (approved):

PR #181, opened 2026-05-04
fix(chat): sanitize chatId in JetStream KV key

LissaGreense:
## Summary

- Telegram (and any `<provider>:<id>`-shaped) chats blew up on first message: NATS KV's `validateKey` regex `/^[-/=.\w]+$/` rejects `:`, so writing metadata threw `Error: invalid key: …`. The agent then loaded empty history and crashed in `stream-text` with `AI_InvalidPromptError: messages must not be empty`, and the assistant reply also failed to persist.
- Stream names already routed through `sanitizeForStreamName` — the KV key path was the asymmetric gap. Both now flow through one `sanitizeKeyComponent` helper, and the workspace prefix used by `listChatsByWorkspace` gets the same treatment so listings still resolve.
- The original chatId stays in `Chat.id` (only the lookup key is lossy), so reads return the same id the caller passed in. NATS subjects tolerate `:`, so the message stream subject is unchanged — only the KV path needed the fix.

## Test Plan

- [x] `deno task test packages/core/src/chat/storage.test.ts` — added a regression test creating a `telegram:<uuid>` chat and exercising `appendMessage` / `getChat` / `listChatsByWorkspace`.
- [x] `deno task test packages/core` — 636 tests pass.
- [x] `deno task test apps/atlasd/src/chat-sdk/` — 124 tests pass.
- [x] `deno check` on the modified files — clean.
- [x] Manual: send a Telegram message into a workspace and confirm the agent reply persists (no `chat_sdk_append_message_failed` / `Invalid prompt: messages must not be empty` in the daemon log).

### DIFF
--- a/packages/core/src/chat/jetstream-backend.ts
+++ b/packages/core/src/chat/jetstream-backend.ts
@@ -70,12 +70,22 @@ export interface ChatStreamLimits {
 
 const SAFE_NAME_RE = /[^A-Za-z0-9_-]/g;
 
-function sanitizeForStreamName(s: string): string {
+/**
+ * Sanitize a chat-id / workspace-id component for use in NATS-derived keys.
+ *
+ * Same rules used for both stream names and KV keys: nats.js's KV `validateKey`
+ * accepts `/^[-/=.\w]+$/` (no `:`), and stream names allow only word chars,
+ * `-`, `_`. The component IDs that flow through here can contain colons (e.g.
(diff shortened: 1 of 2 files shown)


basedfriday:
@LissaGreense qq, good catch. Just to confirm, this was just affecting communicators?

LissaGreense:
> @LissaGreense qq, good catch. Just to confirm, this was just affecting communicators?

Yes! In communicators chat_id depends on provider~

PR #182, opened 2026-05-05
fix(atlasd): decouple cascade execution from signal delivery

basedfriday:
## Summary

- Decouple cascade execution from signal delivery so a slow cascade on
  one workspace stops blocking every other workspace's cron / HTTP
  signal — restores the per-workspace parallelism that was lost when
  signal routing consolidated onto a single JetStream workQueue
  consumer.
- Add a per-signal `concurrency` policy (`skip` default,
  `queue` / `concurrent` / `replace`) on `WorkspaceSignalConfig`,
  applied uniformly to cron, HTTP, and cross-cascade triggers.
- Surface cascade backlog state via a new `INSTANCE_EVENTS` JetStream
  stream + an `/api/instance/events` SSE endpoint so UIs don't poll.

## Background

Crons stopped firing jobs reliably across workspaces. Live repro on
the local daemon showed:

```
$ nats consumer info SIGNALS atlasd-signals
Outstanding Acks: 12 / 256
Unprocessed Messages: 8
Last delivery: 4m12s ago
Waiting Pulls: 0
```

A workspace whose cascade ran ~3 minutes monopolised the consumer; a
trivial no-op cron on a different workspace fired into JetStream but
never produced a session. `nats consumer info` confirmed the consumer
loop hadn't pulled in 4+ minutes — it was wedged inside
`await this.dispatch(envelope)`.

Audit traced the regression to the JetStream-migration PR. Pre-
migration, each workspace had its own NATS subscription and ran its
own drain loop, so workspaces were independent in flight. The
migration consolidated that into one durable workQueue consumer with
a single drain loop that synchronously awaited each cascade — losing
the per-workspace fan-out.

## Approach

Two-stream pipeline:

```
cron / HTTP / cross-cascade emit
    │  publishSignal()
    ▼
SIGNALS stream  (delivery durability)
    │
    ▼ thin forwarder, ack
SignalConsumer → publishCascade()
    │
    ▼
CASCADES stream  (dispatch buffer)
    │
    ▼ applies per-signal concurrency policy,
      kicks off cascade as background Promise (NOT awaited),
      acks on policy decision
CascadeConsumer
    │
    ▼
triggerWorkspaceSignal → runtime.processSignal (FSM engine, unchanged)
```

The `SignalConsumer` becomes a thin forwarder. The `CascadeConsumer`
holds the in-flight registry, applies the `concurrency` policy
(skip / queue / concurrent / replace), and dispatches cascades as
background Promises so a slow cascade on one key never stalls
delivery of the next envelope.

`max_ack_pending=32` on CASCADES caps total in-flight cascades
(env-tunable via `FRIDAY_CASCADE_CONCURRENCY`). Chat dispatches via
`signalToStream` → `triggerSignalWithSession` directly, so chat
bypasses the cap by architecture — interactive flows aren't subject
to background back-pressure.

`max_deliver=1` on CASCADES means a daemon crash mid-cascade leaves
the session as `interrupted` via the existing `markInterruptedSessions`
sweep on next startup; cascades are not auto-replayed at the queue
level. Deliberate behaviour change vs pre-cascade-split — replays of
crashed cascades are likely to fail again the same way and burn ack
budget.

### Per-signal concurrency policy

| Policy | Behaviour | Correlated-caller response |
|---|---|---|
| `skip` (default) | Drop new envelope if a cascade for the same `(workspace, signal)` is in flight | `ok=false, reason=skipped-duplicate` |
| `queue` | Per-key keyed mutex; envelopes run serially in arrival order | Caller awaits up to its own timeout |
| `concurrent` | No overlap guard; full parallelism | Each gets its own response |
| `replace` | Abort the in-flight via existing `activeAbortControllers`; start the new one | Cancelled caller gets `ok=false, reason=cancelled-by-newer-trigger` |

Lives next to `schedule` and `onMissed` on `WorkspaceSignalConfig` —
orthogonal to `onMissed` (downtime catch-up policy).

### INSTANCE_EVENTS stream

Modeled on `WORKSPACE_EVENTS`. Subjects `instance.>`, retention
`Limits`, 7-day `max_age`. Carries:

- `cascade.queue_saturated` — backlog crossed `>0.5 * cap`
- `cascade.queue_drained` — backlog fell back below the threshold
- `cascade.queue_timeout` — envelope sat past `FRIDAY_CASCADE_QUEUE_TIMEOUT` (default 5 min) before pickup
- `cascade.replaced` — `replace` policy aborted an in-flight cascade

`GET /api/instance/events?stream=true` is an SSE feed; consumers
don't poll. `?since=<seq>&type=<filter>` gives replay over the stream
for late joiners and reload-after-disconnect.

## Other fixes folded in

- **`onError` capture in streamText** — the AI SDK's
  `NoOutputGeneratedError` is thrown without `cause` set; the
  underlying API error only reaches the `onError` callback. Adding
  the callback so `createErrorCause` sees the real cause replaces the
  generic "No output generated" with the actual
  rate-limit / auth / model-deprecated reason.
- **Stricter MCP `readyUrl` check** — reject 5xx as not-ready. Servers
  that bind their port before completing init (workspace-mcp during
  OAuth bootstrap is the canonical case) answered GET with 5xx until
  fully ready; the prior "any non-throw counts" rule let
  half-initialised servers pass.
- **Drop domain failures from error to warn** — a single LLM API
  rejection produced two error-level log lines (`fsm-engine` +
  workspace runtime) plus a warn from the cascade dispatcher. The
  inner two are now warn — operators scanning at error level see
  infra-level failures, not every misconfigured `model:` id.
- **`hasActiveSessionsForSignal` removed from runtime** — its only
  caller was the dispatch-time short-circuit at `triggerWorkspaceSignal`,
  which the cascade decoupling moved into `CascadeConsumer`'s per-key
  registry.
- **`cascadeConsumer.getStats()` wired into `/api/daemon/status`** —
  surfaces `inFlight`, `cap`, `saturated` for ops visibility.
- **Synthetic `cascade.queue_drained` on consumer startup** — a
  previous daemon that crashed while saturated leaves a hanging
  saturated event with no matching drained; emitting drained on every
  startup closes the gap so UIs don't see a permanent saturated
  indicator.
- **`nc.flush()` after SSE subscribe** — `subscribe()` returns before
  the broker registers; flushing forces a PING/PONG so the SSE
  handshake honestly means "from now on, you get every event."
- **System-skill docs** — `writing-workspace-signals` documents the
  new `concurrency` policy alongside `onMissed`;
  `friday-cli/references/http.md` documents the new
  `cascadeConsumer` block on `/api/daemon/status` and the
  `/api/instance/events` routes.

## Migration

None needed.

- `concurrency` is optional with `skip` default that matches old
  behaviour. Existing `workspace.yml`s don't set it, get `skip`
  automatically.
- `CASCADES` and `INSTANCE_EVENTS` are CREATE-IF-MISSING; first
  daemon start with the new code auto-creates them.
- SIGNALS durable consumer name and config unchanged. Pre-restart
  in-flight messages are forwarded through CASCADES on redelivery.

## Test plan

- [x] New unit / integration tests in
      `apps/atlasd/src/cascade-stream.test.ts` cover all four
      policies, queue-timeout, correlated-caller responses, stream
      chunk forwarding, and a head-of-line decoupling assertion (slow
      cascade on workspace A; fast no-op on workspace B; B finishes
      within 200ms of fire regardless of A).
- [x] Existing `signal-stream.test.ts` updated to assert the new
      "publish-and-ack" forwarder behaviour.
- [x] `deno task typecheck` — 0 errors across 8225 files.
- [x] `deno task knip` — clean, no findings.
- [x] Full repo test suite — 376 test files, 5054 tests passing.
- [x] Live verification on a multi-workspace daemon: while one
      workspace runs a 3-minute LLM cascade, a no-op cron on a
      separate workspace gets sessions on every tick — same time
      window, no head-of-line. Three consecutive end-to-end
      completions on the second workspace within minutes of the fix
      landing.

### DIFF
--- a/apps/atlasd/src/atlas-daemon.ts
+++ b/apps/atlasd/src/atlas-daemon.ts
@@ -2557,6 +2557,7 @@ export class AtlasDaemon {
 
   getStatus() {
     const cronStats = this.cronManager?.getStats();
+    const cascadeStats = this.cascadeConsumer?.getStats();
 
     return {
       activeWorkspaces: this.runtimes.size,
@@ -2564,6 +2565,7 @@ export class AtlasDaemon {
       cronManager: cronStats
         ? { isActive: this.cronManager?.isRunning || false, ...cronStats }
         : null,
+      cascadeConsumer: cascadeStats ?? null,
       migrations: this.migrationStatus,
       configuration: {
         maxConcurrentWorkspaces: this.options.maxConcurrentWorkspaces,
--- a/apps/atlasd/src/atlas-daemon.ts
+++ b/apps/atlasd/src/atlas-daemon.ts
@@ -710,7 +710,10 @@ export class AtlasDaemon {
           const manager = this.getWorkspaceManager();
           const cfg = await manager.getWorkspaceConfig(workspaceId);
           const sig = cfg?.workspace?.signals?.[signalId];
-          return (sig?.concurrency ?? "skip") as ConcurrencyPolicy;
+          // `sig.concurrency` is typed `ConcurrencyPolicy | undefined`
+          // through the WorkspaceSignalConfig schema in @atlas/config —
+          // no cast needed.
+          return sig?.concurrency ?? "skip";
         } catch {
           return "skip";
         }
--- a/apps/atlasd/src/atlas-daemon.ts
+++ b/apps/atlasd/src/atlas-daemon.ts
@@ -61,6 +61,7 @@ import { configRoutes } from "../routes/config.ts";
 import { cronRoutes } from "../routes/cron.ts";
 import { daemonApp } from "../routes/daemon.ts";
 import { healthRoutes } from "../routes/health.ts";
+import { instanceEventsRoutes } from "../routes/instance-events.ts";
 import { jobsRoutes } from "../routes/jobs.ts";
 import { linkRoutes } from "../routes/link.ts";
 import { mcpRegistryRouter } from "../routes/mcp-registry.ts";
@@ -81,6 +82,12 @@ import { workspacesRoutes } from "../routes/workspaces/index.ts";
 import { integrationRoutes } from "../routes/workspaces/integrations.ts";
 import { mcpRoutes } from "../routes/workspaces/mcp.ts";
 import { CapabilityHandlerRegistry } from "./capability-handlers.ts";
+import {
+  CascadeConsumer,
+  type ConcurrencyPolicy,
+  ensureCascadesStream,
+  publishCascade,
+} from "./cascade-stream.ts";
 import { CHAT_PROVIDERS, type PlatformCredentials } from "./chat-sdk/adapter-factory.ts";
 import { broadcastJobOutput } from "./chat-sdk/broadcast.ts";
 import {
@@ -94,6 +101,7 @@ import { createFSMBroadcastNotifier } from "./chat-sdk/fsm-broadcast-adapter.ts"
 import { ChatTurnRegistry } from "./chat-turn-registry.ts";
 import { DiscordGatewayService } from "./discord-gateway-service.ts";
 import { createApp } from "./factory.ts";
+import { ensureInstanceEventsStream } from "./instance-events.ts";
 import { getAllMigrations } from "./migrations/index.ts";
 import { NatsManager } from "./nats-manager.ts";
 import { ProcessAgentExecutor } from "./process-agent-executor.ts";
@@ -225,6 +233,7 @@ export class AtlasDaemon {
   private capabilityRegistry: CapabilityHandlerRegistry | null = null;
   private processAgentExecutor: ProcessAgentExecutor | null = null;
   private signalConsumer: SignalConsumer | null = null;
+  private cascadeConsumer: CascadeConsumer | null = null;
   private toolWorkers: ToolWorker[] = [];
   private cronManager: CronManager | null = null;
   private workspaceManager: WorkspaceManager | null = null;
@@ -415,6 +424,19 @@ export class AtlasDaemon {
       // signals are work units, not long-term history.
     });
 
+    // CASCADES stream — dispatch buffer between SignalConsumer (forwarder)
+    // and CascadeConsumer (executor). 1h max_age because cascades that
+    // sit longer than that are queue-timeout territory anyway. The
+    // CascadeConsumer enforces its own per-envelope queue-timeout
+    // (FRIDAY_CASCADE_QUEUE_TIMEOUT, default 5min).
+    await ensureCascadesStream(nc);
+
+    // INSTANCE_EVENTS — instance-wide operational feed, consumed via SSE
+    // by /api/instance/events. Used today for cascade backlog / replace /
+    // queue-timeout signals; intentionally open to other instance-level
+    // event types (daemon, health, …) without a stream split.
+    await ensureInstanceEventsStream(nc);
+
     // Wire chat storage to JetStream + eagerly create the CHATS KV bucket
     // so the first cold read doesn't pay the create cost.
     initChatStorage(nc, {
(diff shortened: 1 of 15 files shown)


apps/atlasd/src/cascade-stream.ts
```diff
@@ -0,0 +1,607 @@
...
+  ) {
+    this.name = opts.name ?? "atlasd-cascades";
+    this.expiresMs = Math.max(1000, opts.expiresMs ?? 10_000);
+    this.batchSize = opts.batchSize ?? 16;
+    this.maxAckPending = opts.maxAckPending ?? DEFAULT_MAX_ACK_PENDING;
+    this.queueTimeoutMs = opts.queueTimeoutMs ?? DEFAULT_QUEUE_TIMEOUT_MS;
+    // ack_wait should comfortably exceed the longest legitimate cascade.
+    // 30min covers most LLM jobs. A cascade outlasting this will be
+    // redelivered, but maxDeliver=1 means it's term'd on second receipt.
+    this.ackWaitNs = 30 * 60 * 1_000_000_000;
```
Vpr99: A race condition that my agent noticed:

> `CascadeConsumer` acks messages **before** `runCascade` completes (by design — decoupling). If the daemon crashes after ack but before the cascade finishes, the message is gone (`max_deliver=1`). However, if the daemon crashes **while** a message is being processed but **before** the ack is sent, the message sits unacked on the broker. With `ack_wait = 30 min`, the broker will not redeliver that message for 30 minutes. With `max_deliver=1`, when it finally is redelivered, it is immediately `term`'d — the cascade is lost and the session is left `interrupted`.
>
> The 30-minute blind spot is catastrophic for a queue that is supposed to drain continuously. The comment says "ack_wait should comfortably exceed the longest legitimate cascade," but because ack happens on policy decision (milliseconds after receipt), `ack_wait` is **only** relevant for crash-recovery redelivery. It should be short — e.g., 10–30 seconds — so the broker quickly frees the slot for the next consumer or marks it as redeliverable.

apps/atlasd/src/cascade-stream.ts
```diff
@@ -0,0 +1,607 @@
...
+import {
+  type SignalEnvelope,
+  SignalEnvelopeSchema,
+  type SignalResponse,
+  signalResponseSubject,
+  signalStreamSubject,
+} from "./signal-stream.ts";
+
+/** Per-signal concurrency policy read from workspace.yml. */
+export type ConcurrencyPolicy = "skip" | "queue" | "concurrent" | "replace";
```
Vpr99: This is redelcared from `packages/config/signals.ts` and should likely be referenced from there!

apps/atlasd/src/atlas-daemon.ts
```diff
@@ -665,30 +700,31 @@ export class AtlasDaemon {
...
-      {
-        maxAckPending: jsCfg.consumer.maxAckPending.value,
-        maxDeliver: jsCfg.consumer.maxDeliver.value,
-        ackWaitNs: jsCfg.consumer.ackWaitNs.value,
+      async (workspaceId, signalId): Promise<ConcurrencyPolicy> => {
+        try {
+          const manager = this.getWorkspaceManager();
+          const cfg = await manager.getWorkspaceConfig(workspaceId);
+          const sig = cfg?.workspace?.signals?.[signalId];
+          return (sig?.concurrency ?? "skip") as ConcurrencyPolicy;
```
Vpr99: Type assertion here - I'd suggest either:
a) ideally ensuring that `sig` is correctly typed with `concurrency: ConcurrencyPolicy`.
b) if that's impossible, then pulling in `ConcurrencyPolicySchema` and zod parsing in-situ.

apps/atlasd/src/cascade-stream.ts
basedfriday: You're right — `ack_wait=30min` was a misread of what the knob controls. Since we ack on policy decision (parse → resolvePolicy → ack, all in milliseconds), `ack_wait` only bounds the *crash-recovery* window for the small receive→ack gap, not cascade duration. With `max_ack_pending=32`, a daemon crash during peak would have stalled the next consumer for the full 30 min before the broker reclaimed those slots.

Shortened to 30s in `798c67b`. With `max_deliver=1` the redelivery is term'd anyway — the shortening just unsticks the consumer faster. Updated the comment to spell out the correct semantics so the next person doesn't make the same mistake.

apps/atlasd/src/cascade-stream.ts
basedfriday: Fixed in `798c67b` — dropped the local literal redeclaration. Both `cascade-stream.ts` and `atlas-daemon.ts` now import `ConcurrencyPolicy` directly from `@atlas/config`. Tried re-exporting through `cascade-stream.ts` first, but pushing the discriminated-union inference through this module tripped TS2589 elsewhere; direct import is cleaner.

apps/atlasd/src/atlas-daemon.ts
basedfriday: Fixed in `c5f6433` — went with option (a). `sig.concurrency` is already typed `ConcurrencyPolicy | undefined` through the `WorkspaceSignalConfig` schema (now that the field is on `BaseSignalConfigSchema` in `@atlas/config`), so the cast was redundant. Return-type inference lands at `ConcurrencyPolicy` cleanly.

ljagiello (approved):

PR #191, opened 2026-05-05
fix(mcp-registry): tolerate unknown registryType + drop bad entries instead of 502

ljagiello:
## Summary
- `/api/mcp-registry/search` was returning **502 `{\"error\":\"Search failed\"}`** for queries that matched a server with a package using a `registryType` outside our closed Zod enum (e.g. `nuget`). The upstream spec defines `registryType` as an open string, so a single unrecognized value tanked the entire response. Reproduced live with `?q=reddit` — 16 valid servers were being hidden behind one `nuget` entry (`io.github.mkerchenski/reddit-ads-mcp`).
- Loosened `registryType` to `z.string()` to match the upstream spec.
- Made `search()` parse each entry independently — malformed entries are dropped and logged with their server name, instead of failing the whole call. Structural defense against future upstream schema drift.

## Test plan
- [x] `vitest run packages/core/src/mcp-registry/` — 251/251 pass
- [x] `deno check` clean, biome clean
- [x] Live smoke test against `registry.modelcontextprotocol.io`: `?search=reddit` now returns all 16 servers including the `nuget` one
- [ ] Manual: hit `/api/daemon/api/mcp-registry/search?q=reddit&limit=20` in the running daemon and confirm 200 with full result set

### DIFF
--- a/packages/core/src/mcp-registry/upstream-client.test.ts
+++ b/packages/core/src/mcp-registry/upstream-client.test.ts
@@ -196,6 +196,48 @@ describe("MCPUpstreamClient", () => {
 
       expect(result.servers).toHaveLength(1);
       expect(result.servers[0]?.server.name).toEqual("io.github.test/good-server");
+      expect(result.dropped).toEqual(2);
+    });
+
+    it("reports zero dropped when all entries are valid", async () => {
+      const mockResponse = {
+        servers: [
+          {
+            server: {
+              $schema: "https://schema.modelcontextprotocol.io/server/2025-04-18.json",
+              name: "io.github.test/server",
+              description: "Valid",
+              version: "1.0.0",
+            },
+            _meta: {
+              "io.modelcontextprotocol.registry/official": {
+                status: "active",
+                statusChangedAt: "2025-01-01T00:00:00Z",
+                publishedAt: "2025-01-01T00:00:00Z",
+                updatedAt: "2025-01-01T00:00:00Z",
+                isLatest: true,
+              },
+            },
+          },
+        ],
+      };
+
+      const mockFetch: typeof fetch = (
+        _input: URL | RequestInfo,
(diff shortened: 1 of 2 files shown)


basedfriday:
## Review

**Verdict:** ship it. Targeted fix, good rationale documented in-source, tests cover both the schema relaxation and the per-entry isolation.

### Correctness
- Translator usage (`translator.ts:200,328,337,345`) is all equality-based — widening `registryType` to `z.string()` is safe; unknown values fall through to the existing "transport not supported" rejection at `translator.ts:379`. Reasonable degradation.
- `fetchLatest` correctly kept strict whole-entry parsing (single-entry endpoint should still fail loudly).
- Test additions match the change.

### Suggestions (non-blocking)
1. **`UpstreamSearchResponseSchema` is now dead for parsing** (`upstream-client.ts:139-142`) — only used as a return-type alias. Either drop it and inline `{ servers: UpstreamServerEntry[] }`, or add a comment noting it's documentation-only. Otherwise readers wonder why two near-identical schemas coexist.
2. **Hoist `NameProbeSchema` to module scope.** It's redefined inside `search()` on every call; the rest of the schemas live at module level — consistency wins.
3. **Log noise risk:** if upstream drift becomes chronic you'll get N near-identical "dropped malformed entry" warnings per search. The aggregate `info` log helps. Flag for later if it shows up — not worth changing now.
4. **Test gap (minor):** the new "drops malformed entries" test doesn't assert `dropped` was logged or that the count is 2. The kept-set assertion is what callers see, so this is optional.

### Risks
Low. Schema widening is a strict subset → superset change for valid inputs. No callers rely on the narrow type (verified via grep over `mcp-registry/`). Per-entry `safeParse` is negligible at typical `limit` values. No security implications — `registryType` is consumed via equality, not interpolated.

Issue #193, opened 2026-05-05
stream_event_dropped log spam from non-web chat adapters


basedfriday:
## Summary

`createMessageHandler` in `apps/atlasd/src/chat-sdk/chat-sdk-instance.ts` calls `streamRegistry.appendEvent(chatId, chunk, ownBuffer)` for every client-safe chunk. For non-web adapters (Slack, Telegram, Teams, etc.) the registry has no buffer for the chatId — `appendEvent` returns false and the handler logs `stream_event_dropped` at warn level for every chunk.

A multi-paragraph Slack reply produces hundreds of these warnings per turn, drowning out the *real* drops (cross-turn rejection, buffer overflow) the log line is meant to surface.

## Reproduction

Send any chat message that triggers a streamed reply through a non-web adapter and grep `~/.atlas/logs/global.log` for `stream_event_dropped`. Every text-delta gets a warn line.

## Suggested fix

Skip the `appendEvent` call entirely when `ownBuffer` is undefined — those adapters never use the registry, so the call is just a guaranteed-false roundtrip + log emit:

```ts
(chunk: unknown) => {
  if (!ownBuffer) return;            // non-registry adapter path
  if (isClientSafeEvent(chunk)) {
    const appended = streamRegistry.appendEvent(chatId, chunk, ownBuffer);
    if (!appended) {
      logger.warn(\"stream_event_dropped\", { chatId, chunkType: ... });
    }
  }
}
```

Net effect: warn logs only fire for actual cross-turn drops, which is what the line was designed to flag.

## Related

- `apps/atlasd/src/chat-sdk/chat-sdk-instance.ts:899-911`

basedfriday:
Related: #194 (merged 2026-05-05).

That PR added an identity check to `appendEvent` so the warn line now reliably indicates a real drop (cross-turn rejection / buffer mismatch) rather than firing on benign cleanup. The non-web adapter case described here is unaffected — verified at `chat-sdk-instance.ts:899-911`, the suggested early exit (`if (!ownBuffer) return;`) was not added, so Slack/Telegram/Teams turns still emit a warn per client-safe chunk.

Now that the surrounding noise is gone, this spam is the dominant source of false positives on the line, making the fix higher-value.

Issue #192, opened 2026-05-05
Race: chat handler can capture wrong buffer if follow-up POST lands during subscribe window


basedfriday:
## Summary

`createMessageHandler` in `apps/atlasd/src/chat-sdk/chat-sdk-instance.ts` captures `ownBuffer = streamRegistry.getStream(chatId)` at line 888, but two awaits run before that capture: `thread.subscribe()` (812) and `ChatStorage.appendMessage()` (841). If a follow-up POST arrives in that window, the handler captures the *next* turn's buffer instead of its own, and stale chunks leak into the new turn's stream.

This is the residual race left after #<previous-fix-PR>, which addressed the broader bug where ANY late event from a stale turn could leak. The fix narrowed the window from "anytime during the turn" to "during the ~10–50ms subscribe + appendMessage window", but didn't close it.

## Reproduction

1. POST 1 → adapter creates buffer A → `processMessage` enqueues msg → handler 1 awaits `thread.subscribe()` (yields)
2. POST 2 arrives within the await window → `chatTurnRegistry.replace(chatId)` aborts turn 1 → adapter creates buffer B (replaces A) → `processMessage` enqueues msg 2
3. Handler 1 resumes → `getStream(chatId)` returns B → `ownBuffer = B` (wrong)
4. Handler 1's chunks → `appendEvent(chatId, chunk, B)` → identity matches → leak into B

## Why it matters

User-driven aborts have ~1s human reaction delay so they never hit this window. But automated/scripted clients, or any flow that immediately resends after a 4xx, can race here.

## Suggested fix

Stash `turnBuffer` on `Message.raw` in `atlas-web-adapter.ts:233`, alongside the existing `abortSignal` and `foregroundWorkspaceIds` fields. Read it back in the handler instead of calling `getStream(chatId)`. Three small edits, no architectural change — the precedent is already there.

## Related

- `apps/atlasd/src/chat-sdk/chat-sdk-instance.ts:880-913`
- `apps/atlasd/src/chat-sdk/atlas-web-adapter.ts:233-278`

basedfriday:
Related: #194 (merged 2026-05-05).

[...]

PR #194, opened 2026-05-05
fix(atlasd): prevent cross-turn event leak in chat stream registry

basedfriday:
## Summary

Two cooperating bugs in the chat stream pipeline let a stopped turn's late events corrupt the next turn's stream, surfacing in the UI as a "text delta error":

1. **Cleanup killed the wrong buffer.** `chat-sdk-instance.ts` ran `streamRegistry.finishStream(chatId)` in its finally without an identity check. If a follow-up POST had already created a fresh buffer for the same chatId, the prior turn's cleanup flipped the new buffer to `active=false` — every subsequent text-delta from the new turn was then dropped at the registry and the UI went silent.

2. **`appendEvent` was chatId-keyed only.** Late events from an aborted turn (subprocess emit between abort and SIGTERM, NATS reconnect-buffer flush, `streamSub` drain) landed in whatever buffer was current — including the next turn's, where they arrived without a matching `text-start` and tripped the AI SDK's UI-message protocol validator.

The handler now captures `ownBuffer` once at turn start and threads it through both the write path (`appendEvent(chatId, chunk, ownBuffer)`) and the cleanup path (`finishStreamIfCurrent(chatId, ownBuffer)`). Putting the identity check at the registry write boundary means the guarantee holds even when an upstream layer fails to honor cancellation cleanly.

## Live-daemon QA

Repro: send a chat message, wait 0.8s, send a follow-up to abort it, subscribe to SSE for the chat.

| | events delivered | text content | violations |
|---|---|---|---|
| pre-fix | 6 | none | stream cut to `[DONE]` mid-flight |
| post-fix | 14 | full `text-start → text-delta → text-end` | 0 |

## Test plan

- [x] `apps/atlasd/src/stream-registry.test.ts` — 2 new cases covering identity drop + identity match
- [x] `apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts` — 2 new cases covering cross-turn isolation + cleanup non-interference
- [x] All 3 new tests fail against the pre-fix code; all pass with the fix
- [x] Full repo: typecheck (0 errors), knip (clean), tests (909 passed | 1 skipped)
- [x] Live daemon QA verified end-to-end via SSE subscriber comparison

## Known residual / followups

- **#192** — narrow ~10-50ms race between adapter `createStream` and handler `getStream` capture. Strict improvement, not introduced here. Closing it requires threading `turnBuffer` through `Message.raw`.
- **#193** — `stream_event_dropped` log spam from non-web adapters (Slack/Telegram). Pre-existing, surfaced more clearly now that real drops are quieter.

### DIFF
--- a/apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
+++ b/apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
@@ -136,6 +136,74 @@ describe("createMessageHandler", () => {
     expect(buffer?.active).toBe(false);
   });
 
+  it("drops events from a stale turn that emits after a follow-up replaced its buffer", async () => {
+    // Cross-turn isolation: turn 1's late producer (e.g. a NATS chunk that
+    // arrived after `streamSub.unsubscribe()` had already drained one more
+    // message off the wire) must not bleed into turn 2's buffer. Without the
+    // identity check at appendEvent's expectedBuffer parameter, turn 1's
+    // text-delta would land in turn 2's buffer mid-stream — UI sees a
+    // text-delta with no matching text-start and the AI SDK throws.
+    const registry = new StreamRegistry();
+
+    let captureChunkCallback: ((chunk: unknown) => void) | undefined;
+    const triggerFn = vi.fn(
+      (
+        _signalName: string,
+        _payload: Record<string, unknown>,
+        _streamId: string,
+        onStreamEvent: (chunk: unknown) => void,
+      ) => {
+        captureChunkCallback = onStreamEvent;
+        return Promise.resolve({ sessionId: "sess-1" });
+      },
+    );
+
+    const handler = createMessageHandler("ws-test", triggerFn, registry);
+    const thread = makeThread("chat-cross");
+    const turn1Buffer = registry.createStream("chat-cross");
+
+    await handler(thread, makeMessage({ threadId: "chat-cross" }));
+
+    // Simulate a follow-up turn replacing the buffer mid-flight, then a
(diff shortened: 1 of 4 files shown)


apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
```diff
@@ -136,6 +136,74 @@ describe("createMessageHandler", () => {
...
+      },
+    );
+
+    const handler = createMessageHandler("ws-test", triggerFn, registry);
+    const thread = makeThread("chat-cross");
+    const turn1Buffer = registry.createStream("chat-cross");
+
+    await handler(thread, makeMessage({ threadId: "chat-cross" }));
+
+    // Simulate a follow-up turn replacing the buffer mid-flight, then a
```
basedfriday, line 167: nit: The comment "turn 1's buffer was closed by createStream replacement" is slightly off. `createStream` doesn't close the old buffer, it replaces the map entry and the old buffer becomes unreachable. The buffer's `active` flag stays true until something calls `finishStream` on it. The assertion still works because the buffer captured in `turn1Buffer` was already replaced when `captureChunkCallback` fired, so events didn't land in it. Consider: "turn 1's buffer was replaced by createStream, so the callback references a stale entry".

basedfriday (commented):
Clean fix for a subtle race. The identity-check approach at the registry boundary is the right call - it guards against stale producers regardless of whether upstream layers honor cancellation cleanly. Tests cover both failure modes (cross-turn leak and cleanup interference) and the PR description has good live QA evidence. One minor comment wording nit, otherwise LGTM.

PR #195, opened 2026-05-05
fix(atlasd): cache + prewarm MCP tool probe to avoid cold-start timeout

ljagiello:
## Summary

When a user adds a custom MCP server backed by `npx`/`uvx` and clicks "list tools" right after, the first probe times out for ~5s before the next click succeeds. Logs show the cause:

```
07:27:39.559  MCP stdio connection attempt for "reddit-mcp"
07:27:45.446  MCP stdio connected (5887ms — npx had to download the package)
07:27:45.447  MCP tool probe failed — aborted due to timeout
07:27:58.295  retry → connected in 541ms (npm cache warm) → 16 tools
```

The probe has a hardcoded 5s `AbortSignal.timeout`, just below the cold-install duration. Bumping the timeout would "fix" it but trades a brief wrong error for a long wrong wait. Two better levers:

- **Cache** the tool list, keyed by `serverId` + stable hash of `configTemplate` (1h TTL, success-only). Most repeat clicks become instant cache hits. Invalidated on update / delete.
- **Prewarm** in the background after add / install / custom-add / update with a generous 60s timeout so the npm download finishes before the user clicks.

The PR went through four review rounds that addressed: foreground/prewarm dedup (no duplicate `npx` spawn), credential-disconnect handling (don't cache `[]`), classified prewarm failures (no misleading "still starting up"), retryable response shape, and a regression caught at QA where `createMCPTools` silently drops servers that fail to start (typo'd npm package, MCPStartupError) — those now surface as a real connect-phase error instead of an empty tool list.

## QA — end-to-end verification

Tested against a live dev daemon + playground in Chrome. All four cases verified via the UI's Connection Test panel.

| Scenario | Setup | Expected | Result |
|---|---|---|---|
| **Cache hit on second click** | Add `reddit-mcp` (npx package), wait for prewarm, click "Test Connection" twice | 2nd click instant from cache, no daemon log | ✅ 2nd GET **1ms** (HTTP 200, 16 tools) |
| **Dedup during cold install** | Add `@modelcontextprotocol/server-everything` (uncached), click "Test Connection" immediately while prewarm is still cold-installing | GET awaits in-flight prewarm, no duplicate `npx` spawned | ✅ Cold install 3812ms; GET completes at the same instant (3797ms), single subprocess; subsequent GET 1ms cache hit |
| **Credential disconnect (auth)** | Add MCP with `envVars: [{key: "REDDIT_CLIENT_ID"}]`, no credential connected, click "Test Connection" | UI shows "Authentication failed" with verbatim Link error | ✅ "Authentication failed: No default credential set for cred-test-mcp. Go to Settings > Connections to pick one." (no nested templating) |
| **Typo'd command (connect)** | Add MCP with `args: ["-y", "@nonexistent-org-12345/no-such-mcp-package"]`, click "Test Connection" | UI shows "Connection failed" with real error, not empty tool list | ✅ After fix in `88aca050` — "Connection failed: MCP server 'typo-mcp' failed to start or expose tools" |

### Bug QA caught (and fixed in this PR)

The unit tests mocked `createMCPTools` directly with `{tools: {x: ...}, disconnected: []}`. But the real failure mode for connection errors is `{tools: {}, disconnected: []}` — the server is silently *dropped*, not pushed to `disconnected`. `probeAndExtract` was caching that as `{ok: true, tools: []}` for an hour. The fix in `88aca050` adds an empty-tools guard after the disconnected check; classifier maps the generic Error to `phase: "connect"`. Regression test included.

## Test plan

- [x] `deno task test apps/atlasd/routes/mcp-registry.test.ts` — **80/80 pass** (10 new + 70 pre-existing)
- [x] `deno task typecheck` — 0 errors across 8228 files
- [x] E2E QA via Chrome browser automation against live dev stack (4 scenarios above)

### DIFF
--- a/apps/atlasd/routes/mcp-registry.test.ts
+++ b/apps/atlasd/routes/mcp-registry.test.ts
@@ -78,10 +78,12 @@ beforeEach(() => {
   mockSearch.mockReset();
   mockCreateMCPTools.mockReset();
   mockStreamText.mockReset();
+  _resetCacheForTest();
 });
 
 // Import AFTER mock setup (vi.mock is hoisted, but this makes intent clear)
 const { mcpRegistryRouter } = await import("./mcp-registry.ts");
+const { _resetCacheForTest, _flushPrewarmsForTest } = await import("./mcp-tool-cache.ts");
 
 /** Build a Hono app that wraps the MCP registry router with a partial mock app context. */
 function createWrappedRouter(context: Record<string, unknown>) {
@@ -2090,6 +2092,131 @@ describe("MCP Registry Routes", () => {
       const body = ToolProbeSuccessSchema.parse(await res.json());
       expect(body.tools).toHaveLength(1);
       expect(body.tools[0]).toEqual({ name: "bare-tool", description: undefined });
+    });
+
+    it("serves a cached tool list on the second probe without re-invoking createMCPTools", async () => {
+      const entry = createTestEntry("probe-cache-hit");
+      // Seed without prewarm side effects: prewarm uses createMCPTools too.
+      // Mock first to count both prewarm + probe calls.
+      mockCreateMCPTools.mockResolvedValue({
+        tools: { "cached-tool": { description: "cached" } },
+        dispose: vi.fn().mockResolvedValue(undefined),
+      });
+
+      await mcpRegistryRouter.request("/", {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ entry }),
+      });
+      await _flushPrewarmsForTest();
+      const callsAfterAdd = mockCreateMCPTools.mock.calls.length;
+
+      const first = await mcpRegistryRouter.request(`/${entry.id}/tools`);
+      expect(first.status).toBe(200);
+      const firstBody = ToolProbeSuccessSchema.parse(await first.json());
+      expect(firstBody.tools[0]?.name).toBe("cached-tool");
+
+      const callsAfterFirst = mockCreateMCPTools.mock.calls.length;
+      // Either prewarm filled the cache (first probe is a hit, no new call) or
+      // prewarm was racy and the foreground probe ran. Either way, a *second*
+      // /tools call must be a cache hit.
+      const second = await mcpRegistryRouter.request(`/${entry.id}/tools`);
+      expect(second.status).toBe(200);
+      const secondBody = ToolProbeSuccessSchema.parse(await second.json());
+      expect(secondBody.tools[0]?.name).toBe("cached-tool");
+
+      expect(mockCreateMCPTools.mock.calls.length).toBe(callsAfterFirst);
+      // Sanity: at least the prewarm fired on add.
+      expect(callsAfterAdd).toBeGreaterThanOrEqual(1);
+    });
+
+    it("does not cache failed probes — retry sees a fresh probe", async () => {
+      const entry = createTestEntry("probe-no-cache-on-fail");
+      // Prewarm fails (cold install timeout simulation).
+      mockCreateMCPTools.mockRejectedValueOnce(new Error("ECONNREFUSED 127.0.0.1:8080"));
+
+      await mcpRegistryRouter.request("/", {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ entry }),
+      });
+      await _flushPrewarmsForTest();
+
+      // First foreground probe fails too.
+      mockCreateMCPTools.mockRejectedValueOnce(new Error("ECONNREFUSED 127.0.0.1:8080"));
+      const failed = await mcpRegistryRouter.request(`/${entry.id}/tools`);
+      const failedBody = ToolProbeErrorSchema.parse(await failed.json());
+      expect(failedBody.ok).toBe(false);
+
+      // Now succeed — cache must not contain the prior failure.
+      mockCreateMCPTools.mockResolvedValueOnce({
+        tools: { "recovered-tool": { description: "ok" } },
+        dispose: vi.fn().mockResolvedValue(undefined),
+      });
+      const ok = await mcpRegistryRouter.request(`/${entry.id}/tools`);
+      const okBody = ToolProbeSuccessSchema.parse(await ok.json());
+      expect(okBody.tools[0]?.name).toBe("recovered-tool");
+    });
+  });
+
+  // ═══════════════════════════════════════════════════════════════════════
+  // Prewarm — POST mutations populate the tool cache
+  // ═══════════════════════════════════════════════════════════════════════
+
+  describe("prewarm on add", () => {
+    it("invokes createMCPTools after POST /", async () => {
+      mockCreateMCPTools.mockResolvedValue({
+        tools: { "warmed-tool": { description: "warmed" } },
+        dispose: vi.fn().mockResolvedValue(undefined),
+      });
+
+      const entry = createTestEntry("prewarm-create");
+      const res = await mcpRegistryRouter.request("/", {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ entry }),
+      });
+      expect(res.status).toBe(201);
+
+      await _flushPrewarmsForTest();
+      expect(mockCreateMCPTools).toHaveBeenCalled();
+    });
+
+    it("invalidates cache on DELETE /:id", async () => {
+      const entry = createTestEntry("prewarm-delete");
+      mockCreateMCPTools.mockResolvedValue({
+        tools: { "v1-tool": { description: "v1" } },
+        dispose: vi.fn().mockResolvedValue(undefined),
+      });
+      await mcpRegistryRouter.request("/", {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ entry }),
+      });
+      await _flushPrewarmsForTest();
+
+      const del = await mcpRegistryRouter.request(`/${entry.id}`, { method: "DELETE" });
+      expect(del.status).toBe(204);
+
+      // Re-add same id with a different config; must not serve stale cached tools.
+      const reAdded = {
+        ...entry,
+        configTemplate: { transport: { type: "stdio", command: "echo", args: ["different"] } },
+      };
+      mockCreateMCPTools.mockResolvedValue({
+        tools: { "v2-tool": { description: "v2" } },
+        dispose: vi.fn().mockResolvedValue(undefined),
+      });
+      await mcpRegistryRouter.request("/", {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ entry: reAdded }),
+      });
+      await _flushPrewarmsForTest();
+
+      const probe = await mcpRegistryRouter.request(`/${entry.id}/tools`);
+      const body = ToolProbeSuccessSchema.parse(await probe.json());
+      expect(body.tools[0]?.name).toBe("v2-tool");
     });
   });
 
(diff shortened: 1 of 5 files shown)


basedfriday (commented):
test

apps/atlasd/routes/mcp-registry.ts
```diff
@@ -658,6 +588,9 @@ export const mcpRegistryRouter = daemonFactory
           return c.json({ error: "Server was modified concurrently." }, 409);
         }
 
+        invalidateCache(updatedEntry.id);
+        prewarmTools(updatedEntry.id, updatedEntry.configTemplate, logger);
```
basedfriday, line 592: **Stale-prewarm race on update.** `prewarmTools` dedups by `serverId` only — if a v1 prewarm is still in flight when this update fires, the call returns the existing v1 promise and never starts a v2 prewarm.

After the update settles:
- v1 prewarm eventually resolves and writes `{configHash: v1, tools: v1Tools}` to the cache.
- A subsequent GET checks `getCachedTools(id, v2)` — `configHash` mismatch returns null. Good.
- `getInFlightPrewarm(id)` returns undefined (v1 finished). No dedup wait.
- Falls through to the foreground 5s probe — exactly the cold-start path this PR is trying to avoid.

Narrow window (user has to update mid-prewarm), but it silently degrades the prewarm guarantee precisely when configs change. Consider tracking the in-flight's configHash and replacing when it differs, or having the update path explicitly evict the in-flight before kicking off a fresh prewarm.

**Worth doing:** Probably yes — small change (one extra map keyed by config hash, or carry the hash on the promise). But low severity; merge-shouldn't-block.

apps/atlasd/routes/mcp-registry.ts
```diff
@@ -691,6 +624,7 @@ export const mcpRegistryRouter = daemonFactory
       }
 
       await adapter.delete(id);
+      invalidateCache(id);
```
basedfriday, line 634: **DELETE doesn't cancel an in-flight prewarm.** If a prewarm started moments ago is still running when this fires, the prewarm's `try` block will eventually call `putCachedTools` *after* the `invalidateCache(id)` here, leaving stale tools in the map for a deleted server.

Not user-visible in practice: a re-add with identical config would hit the stale entry but get correct tools; a re-add with different config gets `configHash` mismatch and re-probes. Without re-add, the entry sits there until TTL expires (combined with the TTL-no-evict issue on the cache module, it sits forever until process restart).

**Worth doing:** No — too narrow to justify wiring an AbortController through prewarm. Worth a one-line comment here noting the deliberate leak window so a future reader doesn't re-discover it as a bug.

apps/atlasd/routes/mcp-tool-cache.ts
```diff
@@ -0,0 +1,266 @@
...
+    }
+    return val;
+  });
+}
+
+export function getCachedTools(serverId: string, config: MCPServerConfig): CachedTool[] | null {
+  const entry = cache.get(serverId);
+  if (!entry) return null;
+  if (entry.configHash !== hashConfig(config)) return null;
+  if (Date.now() - entry.cachedAt > CACHE_TTL_MS) return null;
```
basedfriday: **TTL-expired entries are returned-as-null but never deleted.** Cache grows monotonically with stale entries until `invalidateCache` is called or the process restarts. Bound is the count of distinct servers ever added, so not catastrophic, but worth a `cache.delete(serverId)` next to the null return on line 62.

```ts
if (Date.now() - entry.cachedAt > CACHE_TTL_MS) {
  cache.delete(serverId);
  return null;
}
```

**Worth doing:** Yes — one line, straightforward.

apps/atlasd/routes/mcp-tool-cache.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  // Match by `name` as well as `instanceof` — tests that drive the prewarm
+  // through a deferred mock can produce instances whose prototype chain
+  // disagrees with this module's class reference (vitest module-graph
+  // quirk). The constructors all set `name` to a stable string.
+  const credName =
+    inner instanceof Error &&
+    (inner.name === "LinkCredentialNotFoundError" ||
+      inner.name === "LinkCredentialExpiredError" ||
+      inner.name === "NoDefaultCredentialError");
```
basedfriday: **The name-string fallback is the test wagging the production dog.** The comment attributes this to a vitest module-graph quirk, but the fix belongs in the test setup, not in the production classifier. If `instanceof` works reliably in production (which it does — single module graph, no module duplication), the test should be re-importing the same module instances rather than producing prototype-mismatched errors.

Likely cause: `vi.mock("@atlas/core/mcp-registry/credential-resolver", ...)` in the test file replaces the module, so when `mcp-tool-cache.ts` imports the resolver's classes, it gets *mocked* class refs while the test code constructs errors via the real classes (or vice-versa). Routing the test imports through the same mocked module — or using `vi.importActual` for the error classes specifically — would let `instanceof` pass on its own.

The extra branch isn't harmful (it widens classification, doesn't narrow), but it papers over a real test-setup smell. Future readers will copy this pattern.

**Worth doing:** Yes — but as a separate cleanup PR. Don't block this on it.

apps/atlasd/routes/mcp-tool-cache.ts
```diff
@@ -0,0 +1,266 @@
...
+  }
+  // createMCPTools also silently drops servers that fail to start (connect
+  // error, MCPStartupError, list-tools timeout) — they're warn-logged but
+  // don't appear in `disconnected`. For a single-server probe, an empty
+  // `tools` map with no `disconnected` entry means *this* probe failed
+  // silently. Throw so the classifier surfaces a real error rather than
+  // caching `[]` for an hour.
+  if (Object.keys(result.tools).length === 0) {
+    await result.dispose();
+    throw new Error(`MCP server "${serverId}" failed to start or expose tools`);
```
basedfriday: **Empty-tools throw conflates two failure modes.** A server that legitimately registers zero tools (rare, but legal per the MCP spec) is now permanently unloadable — every probe throws. And the message "failed to start or expose tools" is ambiguous: it covers both startup crash *and* list-tools timeout, which the classifier elsewhere distinguishes as `phase: connect` vs `phase: tools`.

For the QA scenario this fixes (typo'd npm package, MCPStartupError silently dropped), the right phase really is `connect`. So the catch-all-classified-as-connect is correct *for the bug being fixed*. It's just imprecise for adjacent failures.

Not blocking, but consider a dedicated error class (`MCPProbeNoToolsError`) so the classifier can pick the right phase, and so legitimately-zero-tool servers can be allow-listed if that ever comes up.

**Worth doing:** No — speculative for unobserved cases. Fine as-is.

apps/atlasd/routes/mcp-registry.test.ts
```diff
@@ -2091,6 +2103,395 @@ describe("MCP Registry Routes", () => {
...
+      // and `await adapter.get(id)` (in-memory KV, normally <1ms) and reach
+      // `await Promise.race(...)` BEFORE the prewarm settles. Otherwise the
+      // IIFE's microtasks (dispose → throw → catch → classify → finally →
+      // delete) finish first, evicting `inFlightPrewarm` before the handler
+      // consults it; the GET then falls through to the foreground probe
+      // path, which still classifies disconnect to "auth" via mockResolved
+      // carry-over OR fails on the exhausted `mockImplementationOnce`. Use
+      // 250ms to stay well clear of GC pauses / loaded-CI jitter — failure
+      // mode (`expected "auth" got "connect"`) doesn't point at timing.
+      await new Promise((resolve) => setTimeout(resolve, 250));
```
basedfriday, line 2431: The 250ms timing dependency here is well-commented but still load-bearing on CI scheduling. Two pragmatic options if this ever flakes:

1. Replace the wall-clock wait with a synchronization signal — expose a hook from the route handler (or a shared queueMicrotask cycle) that the test can `await` to know the GET has reached `Promise.race`. Brittle to refactor.
2. Keep the wait but bump to a vi.advanceTimersByTime + fake timers setup. More setup cost but determinism.

Not blocking — the comment makes the failure mode self-explanatory. Just flagging that this is the test most likely to flake under CI load.

basedfriday (commented):
Solid PR — the cache+prewarm design is the right shape, classified prewarm errors are a clean improvement, and the empty-tools regression caught at QA is a great find. A few non-blocking observations on edge-case races and one small leak. Tests pass locally on the branch (79/79 — PR description says 80/80, minor discrepancy).

Note: I couldn't verify against the running daemon — it's serving from a different worktree, not this branch. Static review + local test run only.

basedfriday:
**Live verification (PR 195 branch checked out, fresh daemon):**

```
POST /api/mcp-registry/custom { command: "uvx", args: ["mcp-server-time"] }
  → 201 in 58ms (non-blocking, prewarm dispatched)

GET /api/mcp-registry/live-test-time/tools  (1s after add)
  → 200, full tool list, 39ms

GET /api/mcp-registry/live-test-time/tools  (immediately after)
  → 200, full tool list, 39ms (cache hit)
```

Happy path works as advertised. uvx was already cached locally so I didn't get to exercise the cold-install path or the retryable hint live; the test suite covers those (79/79 passing on this branch). The update-race I flagged is real on inspection but I couldn't trigger it with warm uvx — would need a genuinely cold package + race timing to repro live.

(Sorry about the empty `"test"` review I left earlier verifying the API endpoint — couldn't delete it once submitted.)

apps/atlasd/routes/mcp-registry.ts
ljagiello: Fixed in 89d2bbd4b — `inFlightPrewarm` now stores `{configHash, promise, marker}`. Same-config callers dedupe; different-config replaces the slot. The v1 IIFE keeps running but its `finally` only evicts if `get(serverId)?.marker === marker` so it doesn 't clobber a racing v2 entry. `getInFlightPrewarm(serverId, config)` filters by hash so the GET handler doesn 't await a stale prewarm. Test: `prewarmTools dedupes same-config callers but starts a fresh probe when config changes`.

apps/atlasd/routes/mcp-registry.ts
ljagiello: Comment added in 89d2bbd4b documenting the deliberate leak window — agreed not worth wiring an AbortController.

apps/atlasd/routes/mcp-tool-cache.ts
ljagiello: Fixed in 89d2bbd4b — `cache.delete(serverId)` next to the null return on TTL expiry. Test: `evicts the entry on TTL expiry — getCachedTools returns null and the map is clean` (uses `vi.useFakeTimers()`).

apps/atlasd/routes/mcp-tool-cache.ts
ljagiello: Fixed in 2d2e15376 — fully removed the name-string fallback. You were right that the original need (an instance-mismatch from a deferred-mock test) is gone — current tests all throw via production code, so `instanceof` matches reliably and the production classifier no longer needs to know about a test concern. 81/81 pass with the fallback stripped.

apps/atlasd/routes/mcp-tool-cache.ts
ljagiello: Done in 2d2e15376 — added a dedicated `MCPProbeNoToolsError` class. `probeAndExtract` throws an instance, and the classifier has an explicit branch routing it to `phase: "connect"`. User-visible behavior unchanged (same message, same phase) but callers can now distinguish silent-drop probe failures from real network connect errors via `instanceof` if that ever matters.

apps/atlasd/routes/mcp-registry.test.ts
ljagiello: Acknowledged — left as-is per your own triage (not blocking; comment makes the failure mode self-explanatory). If it flakes under CI load we 'll come back with the sync-signal or fake-timers approach.

PR #198, opened 2026-05-05
HTML Chat exports

Vpr99:
# Chat Export

One-click export of any chat to a self-contained `.zip`. The bundle opens in any modern browser with no JS runtime, no Friday account, and no backend dependency. Light/dark mode tracks OS preference; tool bursts and reasoning traces collapse and expand natively; HTML and PDF artifacts render inline; images appear where they were sent.

## What's in the export

```
friday-chat-{chatId}.zip
├── index.html                         — full transcript, fully styled, no JS
├── chat.json                          — raw message data (userId stripped)
└── assets/artifacts/{id}/{filename}   — every artifact referenced in the chat
```

The HTML references artifacts via relative paths inside the zip, so the recipient extracts and double-clicks `index.html` — no further setup.

## How it works

The export reuses the **same Svelte components** the live UI mounts (`ChatMessageList`, `ToolCallCard`, `DelegateToolCard`, `ArtifactCard`). Render runs through SvelteKit's normal SSR pipeline; the orchestrator packs the result into a zip alongside artifact bytes.

```
GET /platform/:wsId/chat/:chatId/export        (SvelteKit +server.ts)
   │
   ├─ event.fetch('/.../export/preview')
   │     │  +page.server.ts → daemon GET /chat/:id?full=true + /artifacts?chatId=
   │     │  +page.svelte    → setExportContext({ artifacts, resolveUrl })
   │     │  +page.ts        → csr = false
   │     │  svelte.config   → kit.inlineStyleThreshold: Infinity
   │     ↓
   │     self-contained HTML (all CSS inline, no scripts, no <link>)
   │
   ├─ parallel: GET /artifacts/:id/content for each artifact
   ├─ JSZip pack → index.html + chat.json (userId stripped) + assets/...
   ▼
application/zip
```

### `ExportContext` — the bridge between live and static modes

A small Svelte context carries the two things that differ between live and export rendering:

```ts
type ExportContext = {
  artifacts: ReadonlyMap<string, ArtifactPrefetch>;
  resolveUrl: (id: string) => string;  // → "assets/artifacts/<id>/<filename>"
};
```

- **Live UI** never sets the context. `getExportContext()` returns `undefined`. `ArtifactCard` fetches metadata from the daemon, copy-to-clipboard buttons render, `ResizeObserver` scales iframes — all unchanged.
- **Export preview** sets the context with the prefetched artifact map. `ArtifactCard` resolves synchronously, never calls `fetch()`, and rewrites artifact URLs to relative zip paths. Copy buttons are hidden (no JS to drive them). Iframes render at 1:1.

One component tree, two consumption modes — divergence between live and exported output is structurally impossible.

### Tool bursts are native `<details>`

`chat-message-list.svelte` and `delegate-tool-card.svelte` previously used `<button onclick> + $state Map` to track open/closed. They now use `<details open={initialOpen}><summary>...`:

- Works in the export without JS.
- Keyboard, ARIA, and screen-reader behavior improve in the live UI.
- Auto-expand of a `connect_service` burst that arrives mid-stream is preserved by a live-only `$effect` that flips `el.open = true` when the call needs user action. Effects are no-ops during SSR, so the export sees only the initial state.

### Daemon API surface

The daemon stays a pure data source. The only API change:

- `GET /api/workspaces/:wsId/chat/:chatId?full=true` — query param skips the existing `messages.slice(-100)` trim used by the live rehydrate path. The export needs every message.

SvelteKit owns the export entrypoint; the daemon stays a read-only data source. (An earlier exploration of a daemon-side `GET /chat/:chatId/export` route exists on a separate branch and is not part of this PR.)

### Failure modes

- **Preview render fails (non-2xx)** — orchestrator forwards the upstream status verbatim (e.g. `404` for missing chat).
- **Render exceeds 10s** — `AbortController` cancels the preview fetch; route returns `504 { error: "Chat too large to export" }`.
- **Single artifact byte fetch fails** — server-side warn, that entry is omitted from the zip, the rest of the export is intact. The HTML keeps the relative link, so the recipient sees one broken download — better than no export.
- **Artifact list fetch fails** — render proceeds with an empty list; cards surface a clear "missing from export context" placeholder instead of spinning a loader the static HTML can never resolve.

### App-shell & SSR safety

- The root `+layout.svelte` strips its sidebar, command palette, and `QueryClientProvider` when the route ends with `/export/preview`. The exported HTML carries chat content only.
- `DaemonGate` guards its loading branches with `browser &&` so SSR passes children through (otherwise the gate would freeze on "Connecting to daemon..." since `loading` defaults to true and only flips after a client-side fetch).
- Markdown is rendered through `markdownToHTMLSafe` in the chat list so any user-provided HTML in messages is sanitized before it lands in the bundle.
- `chat.json` strips `userId` at the zip-write boundary — account-ownership PII shouldn't leak in shared exports.

## Test plan

- **`ArtifactCard`** (export mode) — renders synchronously from the prefetch map, does not call `fetch()`, routes URLs through `ctx.resolveUrl()`, surfaces a clear error when the referenced `artifactId` is missing.
- **`ToolCallCard`** (export mode) — copy-button wrapper and button are omitted; with no context, both render. JSON output renders in both modes.
- **Preview page** — `load()` returns chat + messages + `ArtifactPrefetch[]` on the happy path, propagates `404` from the daemon, degrades to an empty artifact list when the artifact endpoint errors. Page render contains user/assistant bubbles + the chat title and never references `/api/daemon/`.
- **Export route** — decode the streamed zip and assert it contains `index.html`, `chat.json` (with `userId` stripped), and `assets/artifacts/{id}/{filename}` entries; a single failed artifact byte fetch omits that entry without failing the export; non-2xx preview status forwards verbatim.

Manual QA: exported a real chat containing markdown, images, reasoning traces, tool bursts, and an HTML map artifact. Opened `index.html` from the zip in a clean browser profile — light/dark mode tracks OS preference, tool bursts expand/collapse natively, the map loads tiles correctly.


basedfriday (commented):
Backend-impact review (HTTP only — no NATS / messaging changes touched).

**1. Unbounded read on `?full=true`.** `validateAtlasUIMessages(messages)` Zod-parses the entire array in one shot on the daemon. For chats with thousands of messages or heavy tool bursts this is O(MB) memory + nontrivial CPU. The SvelteKit orchestrator has a 10s `AbortController` on the preview fetch, but the daemon handler doesn't honor request-abort — it'll keep validating after the client gives up. Low severity since atlasd is loopback-only, but a hard cap (e.g. 10k messages → 413) would be cheap insurance.

**2. Auth model unchanged but worth naming.** `GET /:chatId` does no per-user scoping in the handler — `ChatStorage.getChat(chatId, workspaceId)` is workspace-scoped only, and per-user auth is presumably enforced upstream. `?full=true` doesn't weaken this; anyone who could read the trimmed view can now read the full view. Intentional, just calling it out.

**3. PR description drift.** The body says *"the previous `GET /chat/:chatId/export` daemon route is retired"* but the diff contains no such deletion in `apps/atlasd/`. Either it was removed in an earlier PR or the description is stale — worth a clarifying edit so future archaeology isn't confused.

**4. userId strip lives in SvelteKit, not the daemon.** `chat.json` strips `userId` at the zip-write boundary in `+server.ts`. The daemon endpoint still returns it on the wire (unchanged behavior). Fine since atlasd is loopback, but anyone proxying or logging daemon responses still sees it. Not a regression — noting for completeness.

Test coverage on the new query param is solid (absent / `true` / `false` / `"1"` / `"yes"` / `"TRUE"` / empty all asserted). RPC-client fix in `workspace-chat.agent.ts` is the only caller — verified by grep.

Architecture (one component tree, two consumption modes via Svelte context) is clean. Safe to merge from a backend-impact perspective; (1) and (3) are the only items I'd want the author to look at.

tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/preview/preview.test.ts
```diff
@@ -0,0 +1,319 @@
...
+  const Stub = mod.default;
+  const proxy = new Proxy({}, { get: () => Stub });
+  return {
+    Icons: proxy,
+    IconSmall: proxy,
+    Button: Stub,
+    DropdownMenu: proxy,
+    MarkdownRendered: Stub,
+    markdownToHTML: (text: string) => text,
+    markdownToHTMLSafe: (text: string) => text,
```
Comment (ljagiello) [inline tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/preview/preview.test.ts:45]: **Sanitization isn't actually exercised here.** The `@atlas/ui` mock replaces `markdownToHTMLSafe` with `(text) => text`, so the test that asserts user message text is rendered passes regardless of whether the real sanitizer ran.

The export bundle is opened directly in any browser by a recipient told to trust it — a sanitization regression here becomes stored XSS in a file shipped out of band. This is the highest-leverage missing assertion in the PR.

**Suggestion:** add at least one test that runs the real `markdownToHTMLSafe` against `<script>alert(1)</script>` / `` and asserts the payload is escaped or stripped in the rendered HTML. If `@atlas/ui` is too heavy to import in this test file, consider re-exporting just the sanitizer from `packages/core` so it can be exercised directly.

tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/+server.ts
```diff
@@ -0,0 +1,203 @@
...
+    } else {
+      const summary = artifacts[idx];
+      const reason = result.reason instanceof Error ? result.reason.message : String(result.reason);
+      console.warn(
+        `[chat-export] artifact byte fetch failed for ${summary?.id ?? "<unknown>"}: ${reason}`,
+      );
+    }
+  });
+
+  const zipBytes = await zip.generateAsync({ type: "arraybuffer" });
```
Comment (ljagiello) [inline tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/+server.ts:300]: **No upper bound on zip size — full chat + every artifact buffered in RAM.**

The 10s `EXPORT_TIMEOUT_MS` only wraps the preview render. After that:
1. `Promise.allSettled` over every artifact runs with no concurrency limit (line 143).
2. Each `await res.arrayBuffer()` lands in RAM.
3. `zip.generateAsync({ type: "arraybuffer" })` materializes the entire zip in RAM before the response starts.

A chat with ~50× 10MB image artifacts is 500MB+ resident. Local-only today, but the binary ships in Friday Studio.

**Suggestion:** at minimum, per-artifact byte ceiling (skip + warn above N MB) and an aggregate ceiling — both can reuse the existing single-failure semantics. Streaming via `generateInternalStream` + `ReadableStream` is the longer-term fix but optional for now.

tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/+server.ts
```diff
@@ -0,0 +1,203 @@
...
+  const previewUrl = `/platform/${wsPath}/chat/${chatPath}/export/preview`;
+  const chatUrl = `/api/daemon/api/workspaces/${wsPath}/chat/${chatPath}?full=true`;
+  const artifactsUrl = `/api/daemon/api/artifacts?chatId=${chatPath}`;
+
+  // Race the preview render against a 10s ceiling. We abort the inner
+  // fetch on timeout so the SvelteKit handler chain can clean up, and
+  // return 504 to the caller with a JSON body rather than a half-rendered
+  // HTML stub.
+  const previewController = new AbortController();
+  const previewTimer = setTimeout(() => previewController.abort(), EXPORT_TIMEOUT_MS);
```
Comment (ljagiello) [inline tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/+server.ts:82]: **Only the preview fetch is timed out, and `event.request.signal` isn't propagated anywhere.**

If the daemon hangs after sending the preview response, the chat-fetch, artifacts-fetch, and every per-artifact byte fetch run unbounded. If the user closes the browser tab, the in-flight artifact fetches keep running because `event.request.signal` isn't forwarded to any of the downstream `event.fetch` calls.

**Suggestion:** plumb `event.request.signal` (combined with an outer export-wide timeout via `AbortSignal.any([...])`) into all `event.fetch` calls.

tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/+server.ts
```diff
@@ -0,0 +1,203 @@
...
+      };
+    }),
+  );
+
+  const zip = new JSZip();
+  zip.file("index.html", html);
+  // Strip `userId` from the exported chat — it's account-ownership PII and
+  // shared exports shouldn't leak it. The schema validates the wire shape
+  // upstream; the strip happens at the zip-write boundary.
+  const { userId: _userId, ...chatWithoutUserId } = chatParsed.data.chat;
```
Comment (ljagiello) [inline tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/+server.ts:277]: **The "PII strip" framing oversells what's actually scrubbed.**

This only removes top-level `userId`. `messages[*].metadata` still carries `agentId`/`sessionId`, `systemPromptContext` is forwarded verbatim, and tool-call inputs/outputs frequently embed local file paths (`/Users/<name>/...`), credentials in URL params, etc.

**Suggestion:** either narrow the doc-comment ("removes account-ownership ID; transcript and tool I/O are exported verbatim — recipients see what you saw") or do a real scrub of metadata IDs and surface a "this export contains your prompt + tool I/O" warning at the trigger point. The current half-stripped state with confident framing is the worst of both.

tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/+server.ts
```diff
@@ -0,0 +1,203 @@
...
+import JSZip from "jszip";
+import { z } from "zod";
+
+/**
+ * Hard ceiling on the preview render. The preview load function fans out
+ * to the daemon and the artifact list endpoint, so a hung upstream there
+ * could pin this route indefinitely. 10s mirrors the prior daemon route's
+ * `EXPORT_TIMEOUT_MS` for recipient-facing parity.
+ */
+const EXPORT_TIMEOUT_MS = 10_000;
```
Comment (ljagiello) [inline tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/+server.ts:16]: **The 504 timeout branch and the catch-block 502 are real defensive code with no test.** A future refactor that inverts the abort condition or swallows the error silently would not fail any test.

**Suggestion:** `vi.useFakeTimers()` + a fetch stub that rejects with `AbortError` when the signal fires gives a deterministic <10ms test for the 504 path. Same harness covers the 502 fallback.

tools/agent-playground/svelte.config.js
```diff
@@ -4,7 +4,10 @@ import { vitePreprocess } from "@sveltejs/vite-plugin-svelte";
 /** @type {import('@sveltejs/kit').Config} */
 const config = {
   preprocess: vitePreprocess(),
-  kit: { adapter: adapter({ fallback: "index.html" }) },
+  kit: {
+    adapter: adapter({ fallback: "index.html" }),
+    inlineStyleThreshold: Infinity,
```
ljagiello, line 15: **Global setting — every SSR'd page in the playground now ships its full CSS inlined into every response.** No shared `.css`, no cross-navigation caching. For the export route this is required; for the rest of the app it's a regression (Sidebar, CommandPalette, UpdateBanner styles inlined into every navigation).

**Needs decision:** if the playground stays local-only forever, this is fine — document it. If it ever serves over a network, move CSS inlining into a per-route HTML rewriter inside the export `+server.ts`, restoring per-page CSS caching for the rest of the app.

tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/preview/+page.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * `ssr = true` is non-negotiable: the playground's root `+layout.ts`
+ * exports `ssr = false`, which makes every other route render as the SPA
+ * fallback. The export pipeline only works when this leaf opts back into
+ * SSR — without it `inlineStyleThreshold: Infinity` has nothing to inline
+ * and the response is the 980-byte SPA shell.
+ *
+ * `prerender = false` because chat data is dynamic per workspace + chatId.
+ */
+export const csr = false;
```
Comment (ljagiello) [inline tools/agent-playground/src/routes/platform/[workspaceId]/chat/[chatId]/export/preview/+page.ts:15]: **`csr = false` and `inlineStyleThreshold: Infinity` are the two settings the entire feature rests on, and neither has automated coverage.** The preview test self-flags this and punts to manual browser QA. A SvelteKit upgrade or someone deleting this line thinking it's dead code silently breaks the export.

**Suggestion:** in `server.test.ts`, after the existing zip decode + `index.html` extraction, add `expect(html).not.toContain('<link rel="stylesheet"')` and `expect(html).not.toMatch(/<script[\s>]/i)` (or document the exact set of allowed script tags and assert that set). The harness is already there — just one assertion away.

packages/core/src/chat/export/render.ts
```diff
@@ -435,3 +449,152 @@ export function flattenToolCalls(calls: ToolCallDisplay[]): Map<string, ToolCall
...
+      "mediaType" in part && typeof part.mediaType === "string" ? part.mediaType : "image/png";
+    if (!mediaType.startsWith("image/")) continue;
+    const filename =
+      "filename" in part && typeof part.filename === "string" ? part.filename : undefined;
+    imgs.push({ url: part.url, mediaType, filename });
+  }
+  return imgs;
+}
+
+const TIME_FMT = new Intl.DateTimeFormat(undefined, { hour: "numeric", minute: "2-digit" });
```
ljagiello: **Timestamps render in the renderer's local timezone with no UTC marker.** The export is rendered on the sender's machine, then opened on the recipient's machine. PST sender + JST recipient = silently wrong wall-clock with no indication. The "same calendar day" branch (`12:31 PM` only) is worst because "today" is the renderer's today, not the reader's.

**Suggestion:** either always emit UTC (`timeZone: "UTC"` + a `Z`/`UTC` suffix), or always show the date (`Apr 20, 11:31 PM`) regardless of the same-day branch.

ljagiello (commented):
Code review — verdict **Needs Work**. Architecture is sound (ExportContext, native `<details>`, daemon-as-pure-data-source, the `@atlas/core/chat/export/*` extraction). Test quality on the pure-function side is exemplary (zip is actually decoded and inspected, `?full` matrix covers boundary IDs). Inline comments below flag the items worth addressing before merge — none are show-stoppers. Most important: the markdown sanitizer is mocked to identity in the preview test, so the security boundary that prevents stored XSS in the shared HTML bundle has no real assertion behind it.

Full review doc: `docs/reviews/2026-05-05-chat-export.md`.

PR #199, opened 2026-05-05
fix(fsm-engine): interpolate {{inputs.x}} in agent action prompts

odk-:
## Summary
- Friday workspaces use `agent` actions exclusively, but only the `llm` action path ran `interpolatePromptPlaceholders` (in `buildContextPrompt`). Agent action prompts passed `{{inputs.description}}` through to the agent verbatim while the values sat in the appended `## Input` block — the agent then either ignored the placeholders or refused.
- Added the same interpolation to `WorkspaceRuntime.executeAgent` (against `fsmContext.input`) so both action types behave consistently.
- Extended the placeholder regex to support the Liquid/Jinja-style `| default: 'fallback'` filter (`{{inputs.style | default: 'classic SNES/GBA style'}}`), with single- and double-quoted literals; the early-return on missing `prepareResult` is dropped so default-only placeholders still render.
- Exported `interpolatePromptPlaceholders` from `@atlas/fsm-engine` so the workspace runtime can import it.

## Test plan
- [x] `deno task test packages/fsm-engine/tests/prompt-interpolation.test.ts` — 15 passed (5 new cases for the default filter)
- [x] `deno task test packages/fsm-engine/tests/` — 162 passed
- [x] `deno task test packages/workspace/` — 273 passed
- [x] `deno task test apps/atlasd/src/agent-helpers.test.ts` — 22 passed
- [x] `deno check packages/fsm-engine/mod.ts packages/workspace/src/runtime.ts apps/atlasd/src/agent-helpers.ts`
- [x] Manual: trigger a Friday workspace agent prompt that references `{{inputs.description}}` / `{{inputs.style | default: '...'}}` and confirm the values are substituted in the prompt sent to the agent

### DIFF
--- a/packages/fsm-engine/fsm-engine.ts
+++ b/packages/fsm-engine/fsm-engine.ts
@@ -102,8 +102,12 @@ type PrepareResult = z.infer<typeof PrepareResultSchema>;
  *
  * A Liquid-style `| default: 'fallback'` filter is supported on the placeholder
  * itself (`{{inputs.style | default: 'classic'}}`); it kicks in when the path
- * resolves to undefined/null so optional fields don't have to be set on every
- * invocation. Both single- and double-quoted fallback literals are accepted.
+ * resolves to undefined, null, or empty string, matching Liquid's default
+ * semantics. UI forms commonly submit "" for unfilled optional fields, and
+ * substituting an empty string instead of the fallback would make the filter
+ * useless for its primary use case. Both single- and double-quoted fallback
+ * literals are accepted. Numbers/booleans (including 0/false) are not treated
+ * as missing — those are legitimate values, not absence.
  *
  * Uses only properties already exposed via `PrepareResult` (`task`, `config`,
  * `artifactRefs`); does not reach into ambient FSM state, so there's no way
@@ -136,7 +140,11 @@ export function interpolatePromptPlaceholders(
       cursor = (cursor as Record<string, unknown>)[segment];
     }
     if (cursor === undefined || cursor === null) return fallback ?? original;
-    if (typeof cursor === "string") return cursor;
+    // Empty strings are treated as missing for the `default:` filter (Liquid
+    // convention). Without this, an explicit `default: 'classic'` would do
+    // nothing for the very case authors reach for it — a form field that
+    // arrives as "" rather than absent.
+    if (typeof cursor === "string") return cursor === "" ? (fallback ?? cursor) : cursor;
     if (typeof cursor === "number" || typeof cursor === "boolean") return String(cursor);
     return JSON.stringify(cursor);
   });
--- a/packages/fsm-engine/fsm-engine.ts
+++ b/packages/fsm-engine/fsm-engine.ts
@@ -100,6 +100,11 @@ type PrepareResult = z.infer<typeof PrepareResultSchema>;
  * unresolved placeholders are kept verbatim so typos stay visible during
  * authoring rather than silently rendering as empty strings.
  *
+ * A Liquid-style `| default: 'fallback'` filter is supported on the placeholder
+ * itself (`{{inputs.style | default: 'classic'}}`); it kicks in when the path
+ * resolves to undefined/null so optional fields don't have to be set on every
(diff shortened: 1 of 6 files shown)


basedfriday:
@odk- can you add a set of evals for this?

basedfriday:
## Daemon-side validation

Spun up a minimal workspace (`type: llm` agent, FSM with one `agent` action whose prompt referenced `{{inputs.marker}}` and `{{inputs.style | default: 'neutral and concise'}}`) and triggered three signals against the local daemon to exercise the `runtime.ts:executeAgent` path this PR changes.

| Case | Payload | Agent's echo line | Result |
|---|---|---|---|
| Both inputs | `{marker:"ALPHA-7", style:"neon"}` | `TOPIC_ECHO=ALPHA-7 STYLE_ECHO=neon` | ✓ Substitution ran |
| Marker only | `{marker:"BRAVO-9"}` | `TOPIC_ECHO=BRAVO-9 STYLE_ECHO=neutral and concise` | ✓ Default fired on missing key |
| Empty style | `{marker:"CHARLIE-3", style:""}` | (see below) | ✓ Default fired on `""` (Liquid convention) |

End-to-end behavior matches the unit tests on the actual FSM call site.

## Side-finding worth flagging — empty-string + appended `## Input` block disagree

In case 3 the LLM saw the resolved prompt with `"neutral and concise"` (proves the empty-string-as-missing rule fired). But `buildAgentPrompt` at `apps/atlasd/src/agent-helpers.ts:150` *also* appends:

```
## Input
{ "config": { "marker": "CHARLIE-3", "style": "" } }
```

…with the original raw config — `style: ""` and all. The two signals contradict and the LLM noticed:

> *"I need to write a tagline for 'CHARLIE-3' in a neutral and concise style. However, I notice that the 'style' field in the config is empty (`'style': ''`)."*

It still produced a usable tagline, but the echo line came back as `TOPIC_ECHO=CHARLIE-3 STYLE_ECHO=` (empty). Same shape would happen for any field where an author declares `default:` and the upstream form posts an empty string — the prompt body is correct, the Input block is contradictory.

## Recommendation

**Not a blocker for this PR.** The primary use cases the PR cites (knowledge-base save_entry, workspace-chat agent prompts referencing `{{inputs.content}}`) all involve non-empty inputs, so this corner doesn't bite them. The Liquid convention is the right call for `default:` and unit tests pin it.

**Possible follow-ups, in priority order:**

1. *(small, optional)* In `buildAgentPrompt`, strip empty-string and `null` values from the JSON serialized into the `## Input` block. Mirrors the same "absence" semantics the default filter uses, removes the contradiction, and is a one-line `JSON.stringify` replacer.
2. *(larger, defer)* Have `interpolatePromptPlaceholders` return both the rendered string AND a resolved-config object the runtime can use as the source for the `## Input` block.
3. *(docs)* Add a one-line note to the `interpolatePromptPlaceholders` JSDoc that the `default:` filter changes only the prompt body, not the appended input section.

## Eval suite added — `tools/evals/agents/prompt-interpolation/`

Folded a real-LLM eval into the PR. **5 cases** against Haiku 4.5, `temperature: 0`:

- `simple-substitution` — single `{{inputs.x}}` resolved
- `dotted-path` — nested object via dotted access
- `multi-placeholder` — multiple placeholders all resolved
- `default-filter-fallback` — default fires on missing input
- `default-filter-override` — explicit value beats default

Each case asserts the function-level contract on the rendered prompt (`expectInPrompt` / `expectNotInPrompt`) before scoring the LLM response on three dimensions: no `{{` leak, no missing-input refusal markers, and addresses-the-data keyword check. **15/15 perfect scores across runs.**

**Honest scope note** (in the file's docstring): the eval calls `interpolatePromptPlaceholders` directly, so it does NOT catch a regression that bypasses `runtime.ts:executeAgent`. That end-to-end coverage is the daemon-side smoke test above — not in CI.

Two extra unit-test cases added alongside: apostrophe in double-quoted default (`default: "it's me"`) and explicit empty fallback (`default: ''`).

Pre-push triplet (typecheck / knip / tests): all green.

PR #201, opened 2026-05-05
fix(workspace-chat): block phantom bundled-agent refs at every layer

ljagiello:
## Summary

Workspace-chat (the LLM meta-agent that authors `workspace.yml`) emitted `type: atlas, agent: "email"` after the email bundled agent was deleted in `11667073e`. The atlas-daemon at `apps/atlasd/src/atlas-daemon.ts:1430` threw `Base agent not found: email`, the catch block only logged at `error` level, and any FSM step delegating to the agent stalled silently in production.

Four-layer defense:

1. **Strip stale `email` references** from skill markdown, `list_capabilities` and `upsert_agent` tool descriptions, the Gmail MCP `constraints` blurb, and `references/agent-types.md`. Pure deletions on LLM-priming surfaces.
2. **New validator pass** `checkBundledAgentRefs` in `packages/core/src/mcp-registry/config-validator.ts` hard-fails any `type: atlas` agent whose `agent` field doesn't resolve in `bundledAgentsRegistry`. New issue code `unknown_bundled_agent`. Empty strings flow through to the lookup (don't short-circuit on length — that was the silent-failure shape this catches).
3. **Static drift test** at `packages/system/skills/workspace-api/skill-content.test.ts` scans skill markdown + `prompt.txt` + `list_capabilities.ts` + `upsert-tools.ts` for literal `agent: "<id>"` references qualified by a preceding `type: atlas` declaration, and pins each id to `bundledAgentsRegistry`. Fails CI on stale skill content the next time a bundled agent is added or removed.
4. **Daemon load path** now collects agent-registration failures and surfaces them via the existing `metadata.lastError` / `lastErrorAt` / `failureCount` workspace metadata (same pattern as the cascade-failure handler at `atlas-daemon.ts:691-697`). Workspace status preserved (no flip to inactive) so partial-failure tolerance is unchanged — only the visibility of the broken state changes.

## Test plan

- [x] **Unit tests** — 6 new validator cases (`unknown_bundled_agent`) + 5 priming-file scans = 37/37 passing
- [x] **Mutation-tested** — injected `agent: "email"` into `SKILL.md` → drift test fails at the right line; reverted the empty-string fix → unit test fails at the right case
- [x] **API end-to-end** — POST `qa-phantom-test` workspace.yml to `/api/workspaces/:id/lint` → returns `hard_fail` with `unknown_bundled_agent` issues for both `email` and `snowflake-analyst` (Yena's actual production shape)
- [x] **Daemon load-path end-to-end** — triggered signal on a phantom-agent workspace → `metadata.lastError = "Agent registration: 1 failure(s) — flash-report-email-sender (atlas → email): Base agent not found: email. Edit workspace.yml or run validate_workspace to fix."` populated with ISO `lastErrorAt` and `failureCount: 1`
- [x] **Behavioral eval** — `tools/evals/agents/workspace-chat/agent-type-default.eval.ts` `send-daily-report-email` case + universal `no-phantom-atlas-agent` check across all cases. **20/20 PASS** against `claude-sonnet-4-20250514` at `temperature: 0`
- [x] **Typecheck** — 0 errors across 8227 files

## Notes

- The eval check accepts any of: bare `serverId` in agent tools, `<serverId>/<toolName>` in agent tools, or `enable_mcp_server` call — all three are valid MCP wiring shapes the runtime accepts.
- All three test surfaces (validator, drift test, eval) source from `bundledAgentsRegistry` (with `browser`/`research` aliases) for consistency.
- Diff is intentionally surgical — priming-surface changes are deletion-only; no replacement examples or editorial prose added.


basedfriday:
## Review — block vs. cleanup

**The PR already does both.** Layer 1 of the four-layer defense is pure cleanup — `email` is *removed* from `SKILL.md`, `references/agent-types.md`, `list-capabilities.ts`, `bundled-agent-default.eval.ts`, and the Gmail MCP `constraints` blurb. The title undersells this; something like `fix(workspace-chat): remove stale email refs and block phantom bundled-agent ids` is closer to what landed.

The blocker layers are not redundant with cleanup — they protect against threat models cleanup can't reach:

| Threat | Caught by |
|---|---|
| Stale priming surface (today's `email` leak) | **cleanup** (layer 1) |
| Future stale priming surface when an agent is added/removed | **drift test** (layer 3) — fails CI, forces future cleanup |
| LLM hallucinating a never-existed id (`snowflake-analyst` in Yena's repro) | **validator** (layer 2) + **eval** |
| Hand-edited workspace.yml | **validator** (layer 2) |
| Anything that bypasses validation | **daemon metadata** (layer 4) — note: surfaces, doesn't block |

Cleanup alone would not have caught Yena's incident: her workspace.yml had `agent: "snowflake-analyst"` alongside `agent: "email"`, and the former was never a bundled agent — no amount of priming-surface cleanup catches a pure hallucination. You need the validator.

---

### Issue worth fixing in this PR

**`packages/core/src/mcp-registry/config-validator.ts:298`**

```ts
if (baseId in bundledAgentsRegistry) continue;
```

The `in` operator walks the prototype chain. `agent: "toString"`, `agent: "constructor"`, `agent: "hasOwnProperty"` all silently pass validation against a plain object. The drift test (`known.has(id)`) and eval check (`knownBundledIds.has(...)`) both use `Set.has`, which is safe. Recommend `Object.hasOwn(bundledAgentsRegistry, baseId)` or build a `Set` once at module load. Edge case, but it's the same shape of bug the whole pass exists to prevent.

### Minor

- **`failureCount` increments per *load attempt*, not per *unique failure*** (`atlas-daemon.ts:1485`). A daemon restart on the same broken workspace.yml bumps the counter again. Probably fine, but the field name suggests "how broken is this thing" while the value is "how many times have we tried to load it." Worth a one-line comment or a rename like `loadAttemptFailures`.

The block-vs-cleanup framing is the right one and the PR lands on the right side of it: cleanup where you can, block where you must, drift-test to keep cleanup honest. LGTM after the `in` → `Object.hasOwn` fix.

packages/system/skills/workspace-api/skill-content.test.ts
```diff
@@ -28,3 +29,99 @@ describe("workspace-api SKILL.md content", () => {
...
+ * system prompt — and asserts that every literal `agent: "<id>"` reference
+ * resolves to a real entry in `bundledAgentsRegistry`. It is the cheapest
+ * possible safety net: when a bundled agent is added or removed, this test
+ * fails fast and points at the file that needs to be updated.
+ */
+describe("LLM-priming surfaces only reference real bundled agents", () => {
+  // Files the workspace-chat LLM sees, either via load_skill, the system
+  // prompt, or as a tool description string. New files added under this
+  // umbrella should be appended here.
+  const PRIMING_FILES = [
```
Vpr99: Should we also include all the system skills in this list? Agents noticed that `assets/example-jobs-pipeline.yml` and `example-kb-workspace.yml` also include references to agents.

PR #203, opened 2026-05-05
fix(env): write .env values unquoted, strip quotes on read

odk-:
## Summary

The Settings UI at `/settings` saved env vars through `PUT /api/config/env`, which used `@std/dotenv`'s `stringify` to write the file. That helper wraps any value containing a non-word char (e.g. `-` in `sk-ant-foo`) in single quotes, so a saved API key landed on disk as `ANTHROPIC_API_KEY='sk-ant-foo'`. The Go launcher's `loadDotEnv` reads `.env` raw (no quote stripping) and forwards every entry into spawned services' `Environment`, so agents ended up with `ANTHROPIC_API_KEY` literally containing `'sk-ant-foo'` (quotes and all) — every provider call returned 401.

- `apps/atlasd/routes/config.ts` — replace the `@std/dotenv` `stringify` with a local serializer that only quotes when truly necessary (whitespace, `#`, `$`, leading quote, newline, backslash). API keys, URLs, etc. now write unquoted, matching the Tauri installer's `render_env_lines`.
- `tools/friday-launcher/project.go` — strip a single layer of matching surrounding quotes in `loadDotEnv` (defense in depth: pre-fix `.env` files and hand-edited dotenv-style quoting now work without requiring a re-save through the UI).
- Tests: 10 new vitest cases (`apps/atlasd/routes/config.test.ts`) covering on-disk shape and `PUT → GET` round-trip, plus a Go regression guard (`TestImportDotEnvIntoProcessEnv_StripsSurroundingQuotes`).

## Test plan

- [x] `deno task test apps/atlasd/routes/config.test.ts` — 10/10 pass
- [x] `deno task test apps/atlasd/routes/` — 607/607 pass (no regressions)
- [x] `go test ./tools/friday-launcher/...` — all pass incl. new regression guard
- [x] `deno check apps/atlasd/routes/config.ts` clean
- [x] Manual: save an Anthropic API key via the Settings UI and verify the daemon's spawned agents authenticate successfully

### DIFF
--- a/apps/atlasd/routes/config.test.ts
+++ b/apps/atlasd/routes/config.test.ts
@@ -115,11 +115,30 @@ describe("PUT /env on-disk format", () => {
     expect(onDisk).toBe(`MSG="it's fine"`);
   });
 
-  test("multi-line value uses double quotes with escaped newlines", async () => {
-    await putEnv({ BLOCK: "line1\nline2" });
+  test("newline-bearing values are rejected at the boundary", async () => {
+    // The Go launcher's unquoteEnvValue strips outer quotes only; it
+    // doesn't expand `\n`, so KEY="line1\nline2" would reach spawned
+    // services with a literal backslash-n. Reject up front instead.
+    const app = createApp();
+    const res = await app.request("/env", {
+      method: "PUT",
+      headers: { "Content-Type": "application/json" },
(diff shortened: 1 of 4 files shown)


LissaGreense (commented):
Great work Michal! I have some small comments here~

LissaGreense (approved):
rn looks good for me!

PR #204, opened 2026-05-05
MCP Catalog -  Details improvements

### LINKED ISSUE #200: MCP Catalog Details Typography and Layout
The MCP details page has some issues with browsing and management:

- font sizes are very small and hard to read
- it's hard to scan different values like types of credentials and their status
- configuration keys can overlap the text value due to max width constraints

davidwoolf:
- Increases font sizes for MCP Catalog Servers (the details page)
- Condenses badges to be cleaner and more direct
- Standardizes sections to be more scannable
- Moves up "Test Connection" to be more visible and quickly testable
- Migrates old css variables to the new color system

Fixes #200


tools/agent-playground/src/lib/components/mcp/mcp-server-detail.svelte
```diff
@@ -388,6 +320,51 @@
     scrollbar-width: thin;
   }
 
+  .actions-bar {
+    align-self: flex-end;
+    display: flex;
+    align-items: center;
+    flex-shrink: 0;
+    gap: var(--size-3);
+    position: absolute;
```
basedfriday, line 345: This is `position: absolute` but I don't see a `position: relative` on `.detail-pane`, the `<article>`, or `.content` in `list-detail` — so the actions cluster will end up positioning against the viewport (or whatever positioned ancestor it finds higher up), not the article header. The `align-self: flex-end` on line 324 also stops doing anything once the element is taken out of flow.

Adding `position: relative` to `.detail-pane` (or `<article>`) and dropping the `align-self` should sort it.

tools/agent-playground/src/lib/components/mcp/mcp-credentials-panel.svelte
```diff
@@ -88,6 +95,10 @@
     discovery.providers.length > 0 || discovery.idRefs.length > 0,
   );
 
+  $effect.pre(() => {
+    hasContent = hasCredentialRefs;
```
basedfriday: This round-trip feels inverted: the child computes `hasCredentialRefs` from its own props and writes it back to the parent through a `$bindable` + `$effect.pre`, just so the parent can decide whether to render the `<h3>Credentials</h3>` wrapper around it. Writing to a bindable from an effect is generally the shape you want to avoid in Svelte 5, and the parent ends up with a hidden-but-mounted section plus a duplicate copy of the same predicate.

Could you either lift `hasCredentialRefs` to the parent (it's a pure derivation from `server.configTemplate.env`), or have the panel render its own `<h3>` when it has content? Either way drops the bind and the effect.

packages/ui/src/lib/button.svelte
```diff
@@ -154,52 +130,62 @@
...
-      z-index: 1;
+    .append {
+      opacity: 0.4;
+      transition: all 150ms ease;
+      transform: translate3d(0, 0, 0);
+    }
+
+    /* Variants */
+    &.variant-primary {
+      --button-shadow-outline-color: rgb(238 238 239 / 0.5);
```
basedfriday: Heads up — the original button had a `@media (prefers-color-scheme: dark)` block that set `--button-shadow-outline-color` to a fully transparent value, so the outline-shadow was invisible in dark mode. That override is gone now but the light-mode value here applies in both modes, so every primary button picks up a faint pale ring in dark mode. Pretty sure that wasn't intended.

Either restore the dark-mode override, swap to a `light-dark()` value, or drop `--button-shadow-outline-color` entirely if the ring isn't wanted anywhere.

packages/ui/src/lib/simple-table.svelte
```diff
@@ -0,0 +1,56 @@
...
+    padding-inline-start: 0;
+  }
+
+  table :global(th:last-child),
+  table :global(td:last-child) {
+    padding-inline-end: 0;
+  }
+
+  table :global(tr *:last-child),
+  table :global(tr *:last-child) {
```
basedfriday: Same selector listed twice — was one of these meant to be `tr *:first-child` to clear the inline-start border on the leading cell?

tools/agent-playground/src/lib/components/mcp/mcp-catalog-tree.svelte
```diff
@@ -81,20 +81,18 @@
...
 
   // ---------------------------------------------------------------------------
   // Helpers
   // ---------------------------------------------------------------------------
 
-  function transportBadge(server: (typeof allServers)[number]): string {
-    return server.configTemplate.transport?.type ?? "unknown";
-  }
-
   function securityColor(rating: string | undefined): string {
```
basedfriday, line 90: Looks like dead code now: `securityColor()` has no callers (the `.security-dot` was removed). Same story with `let searchFocused = $state(false)` on line 48 plus the `class:focused` binding (line 112) and the `onfocus`/`onblur` handlers (lines 120–121) — the rule that used `.focused` got replaced with `:focus-within` in the CSS. All of it can come out.

tools/agent-playground/src/lib/components/mcp/mcp-catalog-tree.svelte
```diff
@@ -232,149 +161,95 @@
...
+    transition: color 150ms ease;
   }
 
   .tree-item:hover {
-    background-color: var(--highlight);
-    opacity: 1;
+    color: var(--text-bright);
   }
 
   .tree-item.active {
```
basedfriday, line 231: Tiny one — this rule was overriding a base `.tree-item { opacity: 0.85 }` that got removed in this PR, so it's now a no-op.

packages/ui/src/lib/colors.css
```diff
@@ -53,6 +53,9 @@
 
   --green-300: rgb(15 164 40);
   --green-600: rgb(59 232 146);
+
+  --purple-300: rgb(101 4 255);
+  --purple-600: rgb(172 120 255);
```
basedfriday, line 58: Nit: this line uses 2-space indent in an otherwise 4-space block, and the file is missing a trailing newline.

basedfriday (commented):
Overall direction looks good — the typography bump and the cleaner badge/header treatment read well, and the workspace-usage rewrite (dropping the `createQuery`-inside-`{#each}` pattern for `createQueries`) is a real win. A handful of things below; the biggest two are the actions-bar positioning and the credentials-panel bind/effect handshake.

A few more I didn't drop inline because they're not pinned to a single line:

**Color migration is only partway done.** The PR description says it migrates old CSS variables to the new color system, but most of the CSS in the touched MCP components still references `--color-text`, `--color-surface-{1,2,3}`, `--color-border-1`, `--color-success/warning/accent/error` (e.g. `mcp-test-chat.svelte`, `mcp-credentials-panel.svelte`, `mcp-registry-import.svelte`, `mcp-server-detail.svelte`, `mcp-catalog-tree.svelte`). `button.svelte` itself mixes new (`--text-bright`, `--text-faded`) with old (`outline: 1px solid var(--color-text)`, hover `color-mix(... var(--color-text) 5%)`). Either finish the migration in the files this PR already touches, or trim the description so it doesn't oversell — half-migrated styling will be harder to reconcile later.

**Two table primitives in one PR.** This introduces `SimpleTable` for the 2-column required-config table in `mcp-server-detail`, and at the same time the new `mcp-credentials-list.svelte` reaches for the TanStack-backed `Table.Root` for a flat list of 1–3 credentials with no sort/filter/page. Worth picking one direction so `@atlas/ui` doesn't end up with two overlapping tables.

**'Built-in' → 'Bundled'** in `mcp-server-utils.ts` is fine on its own, but other agent UI surfaces (`agent-index-sidebar.svelte`, `agent-reference-panel.svelte`) still say 'Built-in'. Probably worth aligning in a follow-up rather than here.

Net: solid refactor, just needs a couple of cleanups before it ships.

tools/agent-playground/src/lib/components/mcp/mcp-server-detail.svelte
davidwoolf: align-self and flex: 0 were pointless, positioning is fine (for now) due to the placement. However, I want to encode this in a bigger Page component in a future refactor.

packages/ui/src/lib/button.svelte
davidwoolf: Good callout, these aren't being used anymore, removed.

packages/ui/src/lib/simple-table.svelte
davidwoolf: ah yep! copy and paste error, should just be one item!

tools/agent-playground/src/lib/components/mcp/mcp-catalog-tree.svelte
davidwoolf: Great catch! This reminds me that knip is not currently finding unused functions @Vpr99

tools/agent-playground/src/lib/components/mcp/mcp-catalog-tree.svelte
davidwoolf: So this is correct, the opacity change was removed and we apply a brighter text on hover for some affordance.

tools/agent-playground/src/lib/components/mcp/mcp-credentials-panel.svelte
davidwoolf: Yeah this was over engineered. Fixed with a simple css :has selector

PR #205, opened 2026-05-05
docs(readme,setup): rewrite README, expand setup-dev-env.sh into a full bootstrap

basedfriday:
## Summary

- **README rewrite.** Chat-first framing — workspaces are how you ship a chat pattern, not the only way in. `hellofriday.ai` repositioned as the packaged distribution of the same daemon + playground (not "the offering" the README is about). Prereq table split into "you install" (Node, git) and "the setup script handles" (Deno, Go, uv, nats-server, agent-browser, Python + SDK), with explicit min versions matching `*_MIN` constants in the script. Outdated SQLite/Postgres phrasing corrected to JetStream now that `storage.db` / `activity.db` / `skills.db` have been migrated/deleted.
- **`scripts/setup-dev-env.sh` expanded into a one-shot bootstrap with version detection.** Keeps your existing toolchain wherever it satisfies the min; only auto-installs missing or stale tools.

## What the script does now

1. **Preflight** — Xcode CLT (macOS), Node ≥ 24, git. Fails fast with install hints if missing.
2. **Deno** — curl install (or `brew upgrade deno` for Homebrew installs).
3. **Go** — Homebrew on macOS; hard-fail on Linux (no sudo / asdf clobber).
4. **uv** — astral.sh curl (or `brew upgrade uv` for Homebrew installs).
5. **nats-server** — pinned to `v2.12.8` (matches Dockerfile). Homebrew on macOS, otherwise `go install` + `~/.local/bin` symlink so re-runs are idempotent without shell-profile edits.
6. **agent-browser** — `npm i -g agent-browser` + Chromium fetch. Soft-fails to a `SKIPPED_TOOLS` summary so users see what's disabled rather than aborting the whole bootstrap on a system-Node permissions issue.
7. **Stale Python `friday-agent-sdk` cleanup** — strips editable installs that shadow the uv-managed SDK; deduped to avoid double-uninstalling when `python3` and `python3.12` resolve to the same brew binary.
8. **Daemon envfile** — writes `FRIDAY_UV_PATH`, `UV_PYTHON_INSTALL_DIR`, `UV_CACHE_DIR`, `FRIDAY_AGENT_SDK_VERSION`, and now `FRIDAY_JETSTREAM_STORE_DIR=$home/jetstream` (identical to the daemon's current default in `apps/atlasd/src/nats-manager.ts`, pinned for parity with the studio-installer launcher).
9. **Pre-warm uv cache** — Python 3.12 + pinned `friday-agent-sdk` wheel so the first user-agent spawn isn't a cold start.

## For @ljagiello

The script now writes `FRIDAY_JETSTREAM_STORE_DIR=<friday_home>/jetstream` to the daemon `.env`. The studio-installer launcher should emit the same env var at runtime so installer-mode and dev-mode agree on JetStream data location (rather than falling through to nats-server's `$TMPDIR/nats/jetstream` default). Pair this with the existing uv path emission in `apps/studio-installer/src-tauri/src/commands/prewarm_agent_sdk.rs`. There's an inline `@ljagiello` comment at the env-write site as a reminder.

## Test plan

- [x] `bash -n scripts/setup-dev-env.sh` — clean
- [x] `version_ge` correctness verified against six pairs (`2.7.4 vs 2.7.0`, `2.6.0 vs 2.7.0`, `2.7.0 vs 2.7.0`, `1.26.1 vs 1.26.0`, `1.25.0 vs 1.26.0`, `0.11.5 vs 0.11.0`) — all pass
- [x] `STALE_PYTHONS` dedup verified with synthetic `(python3, python3, python3.12)` — collapses to two
- [x] `deno task typecheck` — 0 errors (8230 files)
- [x] `deno run -A npm:knip` — clean
- [ ] `deno task test` — 5109 pass / 2 fail (3 test files). Failures are unrelated to this change: `apps/link/src/routes/communicator.test.ts` and `apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts` are NATS test-server flakes (network race in `packages/core/src/test-utils/nats-test-server.ts:52`); `packages/workspace/src/__tests__/standing-orders-bootstrap.test.ts` is a separate workspace test. This PR only touches `README.md` and `scripts/setup-dev-env.sh` — no TS code changed.
- [ ] Manual: fresh-machine run on macOS to confirm no-op on a tool-complete laptop (please test before merge)
- [ ] Manual: confirm Linux dev path still surfaces clear errors when Go / brew unavailable

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,38 +3,102 @@
 [![CI](https://github.com/friday-platform/friday-studio/actions/workflows/test.yml/badge.svg)](https://github.com/friday-platform/friday-studio/actions)
 [![Discord](https://img.shields.io/badge/Discord-Join-5865F2?logo=discord&logoColor=white)](https://discord.gg/uczJyp5FMH)
 
-AI agent orchestration platform. Workspaces run autonomous agents triggered by
-signals (HTTP, cron).
+Friday turns natural-language asks into repeatable AI-powered workflows.
+You describe what you want in chat — *"every morning, triage my inbox,
+draft replies, file real asks as Linear tickets"* — and the result is a
+`workspace.yml` you can read, version, share, and run on schedule. Run it
+once, or run it forever.
 
-The daemon manages workspace lifecycles — each workspace defines agents,
-signals, and jobs in a `workspace.yml`. Signals arrive (HTTP, CLI, cron), the
-daemon routes them to the workspace runtime, which spawns sessions where agents
-execute with MCP tool access.
+**Why both a chat and a YAML?** Prompts decay. Run the same prompt tomorrow
+and the output drifts; bump the model and your old contracts shift with
+it. Friday keeps both halves of the loop: **conversation** is how you
+build, **configuration** is how you ship — agents, MCP tools, signals
+(HTTP, cron, Slack, email, webhook), and the jobs that wire them together.
+Read the file. Version it. Hand it to a teammate and it runs the same on
+their machine.
+
+Two surfaces, one runtime:
+
+- **Chat in the playground.** Friday picks up your installed skills, MCP
+  servers, memory, and OAuth credentials, and actually drives your tools
+  instead of describing what it would do.
+- **Run autonomously.** Capture the pattern in a `workspace.yml`, bind it
+  to a signal, and the daemon runs it observably and locally on schedule.
+
+It's for developers and operators who want agents in production —
+reviewing PRs at 3am, draining their inbox before standup, watching
+competitor pricing, summarizing a Fathom call into a Linear ticket —
+without wiring queues, secret stores, schedulers, MCP plumbing, and
+provider clients from scratch every time.
+
+> **Prefer the packaged version?** **[hellofriday.ai](https://hellofriday.ai)**
+> ships the same daemon and playground as a one-click installer for macOS
+> and Windows, with the launcher, dependencies, and tray UI bundled. The
+> rest of this README is for working in-tree.
+
+## Architecture
+
+| | What it is | When you use it |
+| --- | --- | --- |
+| **`atlasd` (daemon)** | Headless runtime — HTTP API on `:8080`, workspace lifecycle, signal router, session state, credentials, JetStream message bus. | Production. CI. Anything you'd `systemd`-ify. |
+| **Agent Playground** | SvelteKit web UI on `:5200` — chat with Friday, author and run agents, inspect every session step-by-step, manage skills/MCP servers/schedules/memory/credentials, browse the workspace marketplace. | Day-to-day. The playground is what you actually look at. |
+
+The desktop installer ships both behind a tray icon. In-tree, `deno task
+dev:playground` runs both side by side with hot reload.
 
 ## Quickstart
 
 ### Prerequisites
 
-| Tool | Version | Why |
-| --- | --- | --- |
-| [Deno](https://deno.com/) | `2.7.0+` | Runs the daemon, CLI, and TypeScript packages |
-| [Go](https://go.dev/) | `1.26+` | Builds the Go services under `tools/` (pty-server, webhook-tunnel, friday-launcher) |
-| [Node.js](https://nodejs.org/) | `24+` | Needed for `npx`, Vite, and the web playground |
-| [uv](https://docs.astral.sh/uv/) | `0.11+` | Provisions the managed Python that user agents run under (auto-installed by `setup-dev-env.sh` if absent) |
-| [git](https://git-scm.com/) | any recent | — |
-| Docker (optional) | any recent | Alternative path: run the full stack with `docker compose up` |
+The daemon orchestrates several runtimes — Friday agents can be authored
+in TypeScript *or* Python, can drive a real browser, and fan out work over
+a message bus. Install the four below yourself, then `scripts/setup-dev-env.sh`
+handles the rest.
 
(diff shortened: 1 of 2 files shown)


scripts/setup-dev-env.sh
```diff
@@ -144,19 +388,30 @@ upsert_env() {
...
+    upsert_env "FRIDAY_AGENT_SDK_VERSION"     "$PINNED_SDK_VERSION"
+    # JetStream store dir: identical to the daemon's current default
+    # (apps/atlasd/src/nats-manager.ts → join(getFridayHome(), "jetstream")),
+    # but pinned in .env so the value is immune to default-changes in the
+    # daemon and visible in `cat ~/.atlas/.env` for operators debugging.
+    # @ljagiello — the studio-installer launcher should emit this same env
+    # var at runtime so installer-mode and dev-mode keep JetStream data
+    # under friday_home rather than $TMPDIR. Pair this with the existing
+    # uv path emission in the launcher.
+    upsert_env "FRIDAY_JETSTREAM_STORE_DIR"   "$home/jetstream"
```
basedfriday, line 452: @ljagiello — the studio-installer launcher should emit this same `FRIDAY_JETSTREAM_STORE_DIR=<friday_home>/jetstream` at runtime so installer-mode matches the dev-mode this script sets up. Pair with the existing uv path emission in `apps/studio-installer/src-tauri/src/commands/prewarm_agent_sdk.rs`. Without it, installer-mode falls through to nats-server's `$TMPDIR/nats/jetstream` default and JetStream data lands outside `friday_home`.

README.md
```diff
@@ -3,276 +3,342 @@
...
+> **Prefer the packaged version?** **[hellofriday.ai](https://hellofriday.ai)**
+> ships the same daemon and playground as a one-click installer for macOS
+> and Windows, with the launcher, dependencies, and tray UI bundled. The
+> rest of this README is for working in-tree.
+
+## Architecture
+
+| | What it is | When you use it |
+| --- | --- | --- |
+| **`atlasd` (daemon)** | Headless runtime — HTTP API on `:8080`, workspace lifecycle, signal router, session state, credentials, JetStream message bus. | Production. CI. Anything you'd `systemd`-ify. |
```
ljagiello: Minor accuracy: `atlasd` doesn't own credentials — `apps/link/` is the credential/OAuth service, and the Project Structure block (line ~402) attributes it correctly. Suggest dropping `credentials,` from this row (or rephrasing to "credential consumer") to keep the two sections consistent.

scripts/setup-dev-env.sh
```diff
@@ -62,61 +184,184 @@ detect_home() {
...
-    # The installer adds to ~/.local/bin; pick it up for this script's session.
-    export PATH="$HOME/.local/bin:$PATH"
-fi
+echo "→ Pinned SDK:     friday-agent-sdk==$PINNED_SDK_VERSION"
+echo "→ Pinned NATS:    nats-server $PINNED_NATS_VERSION"
+
+# ── 1. Ensure Deno + workspace deps ─────────────────────────────────────────
+case "$(check_min_version deno "$DENO_MIN"; echo $?)" in
+    0) echo "→ deno:          $(command -v deno) ($(get_tool_version deno))" ;;
+    1) if brew_install_or_upgrade deno; then
```
ljagiello: Subtle footgun for mixed-toolchain machines: if a user has a curl-installed deno that's stale and Homebrew is available, this branch calls `brew_install_or_upgrade deno`, which falls into `brew install deno` (since `brew list deno` is false). Result is two denos on disk, with `$PATH` order picking the winner — the "stale" curl install is bypassed rather than upgraded.

The PR description says "keeps your existing toolchain wherever it satisfies the min," which is true for satisfying installs but not for stale curl installs.

Suggest gating the brew path on whether the existing binary is under `$(brew --prefix)/bin`; otherwise prefer the in-place `deno upgrade` / `uv self update` path. Same applies to the uv case-1 branch below.

ljagiello (commented):
Reviewed the README + bootstrap. Verdict: **clean** — no critical findings. Two important nits inline (`brew` install can shadow a curl-installed deno/uv; "credentials" is misattributed to atlasd in the architecture table) and one decision point: the `@ljagiello` `FRIDAY_JETSTREAM_STORE_DIR` parity action — is that landing here or as a follow-up?

Full review notes: `docs/reviews/2026-05-05-readme-setup-bootstrap.md`.

README.md
basedfriday: Agreed — fixed in 86d94ea. Dropped `credentials` from the daemon row; the project-structure block continues to attribute `link/` correctly.

scripts/setup-dev-env.sh
basedfriday: Good catch — fixed in 86d94ea. Added an `is_brew_managed` helper that resolves the binary on PATH and checks against `$(brew --prefix)`. Stale-branch routing is now:

- **deno / uv** — brew-managed → `brew upgrade`; otherwise → `deno upgrade` / `uv self update` (in-place, no shadow).
- **go / nats-server** — brew-managed → `brew upgrade`; otherwise → fail-fast asking the user to upgrade via their existing installer. These have no self-update, so the only alternative would be silently shadowing, which is worse than a clear stop.

Verified the helper against six synthetic paths (Apple-Silicon brew prefix, Cellar paths, `~/.deno/bin`, `/usr/local/bin`, `~/.local/bin`, `/usr/local/go/bin`) — all classify correctly.

PR #206, opened 2026-05-05
chat handler race fix + workspace/instance SSE feeds + cascade banner

### LINKED ISSUE #183: UI consumer for INSTANCE_EVENTS SSE feed (cascade backlog badge / banner)
Follow-up from #182.

The cascade decoupling PR ships the `INSTANCE_EVENTS` JetStream
stream + the `GET /api/instance/events?stream=true` SSE endpoint.
What's missing is a front-end consumer that surfaces the cascade
backlog state in the playground UI.

## Backend already in place

- Stream: `INSTANCE_EVENTS`, subjects `instance.>`, retention
  `Limits`, 7-day max_age.
- Event types currently emitted by `CascadeConsumer`:
  - `cascade.queue_saturated` — backlog crossed `>0.5 * cap`
  - `cascade.queue_drained` — backlog fell back below the threshold
  - `cascade.queue_timeout` — envelope sat past
    `FRIDAY_CASCADE_QUEUE_TIMEOUT` (default 5 min) before pickup
  - `cascade.replaced` — `replace` policy aborted an in-flight
    cascade
- HTTP routes (`apps/atlasd/routes/instance-events.ts`):
  - `?stream=true` — SSE feed; subscribes to the NATS subject
  - `?since=<seq>&type=<filter>&limit=<n>` — replay over the stream
- Daemon status: `GET /api/daemon/status.cascadeConsumer` —
  `{ inFlight, cap, saturated }` for point-in-time view.

## What this issue covers

A small front-end surface in `tools/agent-playground` that consumes
the SSE feed and renders cascade backlog state. Suggested shape:

1. **Persistent badge / banner** when `cascade.queue_saturated` is
   the latest cascade-backlog event without a matching drained.
   Clears on `cascade.queue_drained`.
2. **Toast / inline notice** for `cascade.queue_timeout` and
   `cascade.replaced` (rate-limited or batched if there's a flood).
3. **Optional `/schedules`-style page** listing recent instance
   events, fed by the replay endpoint with the SSE feed for live
   updates.

## Notes

- The SSE endpoint already does `nc.flush()` after subscribe, so
  the handshake is honest about \"from now on, you get every event.\"
- `INSTANCE_EVENTS` is `Limits` retention (not WorkQueue) — multiple
  UI clients can subscribe simultaneously without consuming each
  other's messages.
- The replay endpoint uses `streams.getMessage` with `AllowDirect:
  false` (default), so reads go to the leader and are read-your-
  writes consistent.
- Stale-saturated state across daemon restarts is already handled —
  `CascadeConsumer.start` emits a synthetic `cascade.queue_drained`
  with `inFlight: 0` so a previous daemon's hanging saturated event
  doesn't leak forward.

## Out of scope

- Migrating the existing `/schedules` page from polling
  `WORKSPACE_EVENTS` to the same SSE pattern. Worth doing
  eventually, but that's a refactor of an existing surface, not part
  of this issue.

### LINKED ISSUE #184: Migrate /schedules page from polling WORKSPACE_EVENTS to SSE
Follow-up from #182.

The cascade-decoupling PR introduced an SSE feed pattern for
`INSTANCE_EVENTS` (`/api/instance/events?stream=true` →
`apps/atlasd/routes/instance-events.ts`). The same pattern applies to
the existing `WORKSPACE_EVENTS` stream that drives the
`/schedules` page in `tools/agent-playground`.

## Current state

`tools/agent-playground/src/routes/schedules/+page.svelte` consumes
events via TanStack Query polling:

```ts
queryKey: [\"workspace-events\", \"schedule.missed\"],
// invalidated on operator action via queryClient.invalidateQueries
```

Each tab open polls the daemon at the query's stale-time interval —
fine when there are few clients but wasteful for what's a low-volume
event stream.

## Target

Mirror the `instance-events` shape:

1. Add `GET /api/workspaces/:workspaceId/events?stream=true` (or
   global `GET /api/events?stream=true`) that subscribes to
   `events.<workspaceId>.>` (or `events.>`) and forwards as SSE.
   `nc.flush()` after subscribe so the handshake is honest.
2. Existing routes (`GET /api/workspaces/:workspaceId/events`,
   `GET /api/events`) stay as the replay path for late joiners and
   reload-after-disconnect.
3. Front end: replace the TanStack polling query with an
   `EventSource` (or fetch-stream reader) that consumes the SSE feed
   and merges into the existing data shape. Existing
   fire/dismiss/group-action POSTs are unchanged.

## Watch out for

- **Operator-action lifecycle is sticky state, not events.** Manual
  `schedule.missed` events have a sidecar KV (`WORKSPACE_EVENT_STATE`
  in `apps/atlasd/src/workspace-events.ts`) that tracks
  `pending` / `fired` / `dismissed`. The stream entries are
  immutable; operator actions flip the KV, not republish. The SSE
  feed only carries new event publishes — the UI still needs the
  replay endpoint to resolve current state on first load. After
  that, listen for events AND optimistically update on
  fire/dismiss.
- **`WORKSPACE_EVENTS` retention is Limits / 30 days** (vs
  `INSTANCE_EVENTS` 7 days). Replay scans further back; pagination
  via the existing `limit` / `scanLimit` query params.
- **Self-heal write-on-read** at
  `workspace-events.ts:413-415` writes a pending KV record if the
  read finds an event without one. That logic still runs on every
  GET; the SSE path only needs to deliver new events, not run the
  heal. Keep the heal on the replay/list endpoint.

## Out of scope

- Adding new event types to `WORKSPACE_EVENTS`. The SSE endpoint
  surfaces whatever the stream carries; new types can be added in
  separate work.

### LINKED ISSUE #192: Race: chat handler can capture wrong buffer if follow-up POST lands during subscribe window

### LINKED ISSUE #193: stream_event_dropped log spam from non-web chat adapters

basedfriday:
## Summary

Closes four open issues across atlasd + agent-playground:

- **Closes #192** — chat handler captured `streamRegistry.getStream(chatId)` after two awaits, so a follow-up POST landing in that ~10–50ms window leaked late chunks from the prior turn into the next. The web adapter now stashes the buffer on `Message.raw.turnBuffer` at dispatch and the handler reads it back deterministically (boundary-validated via an `isStreamBuffer` structural type guard, not a cast).
- **Closes #193** — with capture now precise, non-web adapters (Slack/Telegram/Teams) — which never create a registry buffer — were still hitting `appendEvent` and tripping `stream_event_dropped` on every chunk. Early-exit the tap when `ownBuffer` is undefined.
- **Closes #184** — new `?stream=true` SSE branch on `/api/events` mirroring the existing `/api/instance/events` pattern; `/schedules` page drops its 60s polling and consumes the feed via `EventSource`. Replay endpoint stays for first-load hydration / KV-resolved state. EventSource open is gated on the replay query's `isSuccess` so SSE events landing mid-fetch aren't clobbered when TanStack writes the replay snapshot.
- **Closes #183** — new `CascadeStatusBanner` mounted in playground `+layout.svelte` consuming the existing `INSTANCE_EVENTS` SSE feed: persistent banner on `cascade.queue_saturated` (clears on drained), toasts for `cascade.queue_timeout` / `cascade.replaced`. Hydrates from replay before opening the live feed.

## Review-driven fixes folded in

- `CascadeStatusBanner` awaits hydrate before opening EventSource (closes snapshot-clobbers-fresher-SSE race).
- Both EventSources got `error` listeners (playground `console.*` allowed in dev tooling).
- `sub.unsubscribe()` added to for-await `finally` of both SSE routes — defense-in-depth for the controller-cancel-before-abort path.
- Core-vs-JetStream tradeoff documented in both SSE route comments.
- `as StreamBuffer` cast on `Message.raw.turnBuffer` replaced with `isStreamBuffer` structural type guard exported from `stream-registry.ts`.
- `/schedules` `$effect` gated on `eventsQuery.isSuccess` (mirrors hydrate-then-subscribe ordering from the cascade banner).
- Race regression test rewritten with proper timeline (turn-2 swap during turn 1's `await` window via `mockAppendMessage.mockImplementationOnce`) and a positive `appendEventSpy` assertion that the handler called `appendEvent` with `turn1Buffer` as `expectedBuffer` — directly tests the fix mechanism instead of a negative-only assertion that would have stayed green if the tap were disabled entirely.

## Deliberately deferred → #212

- NATS slow-consumer mitigation (silent message loss when an EventSource client throttles).
- Subscription-leak window before `start(controller)` runs in both SSE routes (`subscribe` runs before the `ReadableStream` is constructed; the abort listener registers inside `start`).
- Direct integration tests for the SSE backend branches (subscribe + flush + abort + finally cleanup).

All three pair well as one focused follow-up PR; gated on this surface shipping beyond the dev-only playground.

## Test plan

- [x] `deno task typecheck` — 0 errors (8237 files)
- [x] `npx knip` — clean (only pre-existing `$types` resolution warnings)
- [x] `deno task test` (full suite) — 5187 passed | 26 skipped (382 files). Includes the rewritten race regression test (`captures turnBuffer from message.raw, not via getStream, to defeat subscribe-window race`) and the wire-up identity assertion in `atlas-web-adapter.test.ts`.
- [x] **Live-daemon QA** — passed end-to-end. Headers correct on both SSE endpoints; both EventSources connect from the playground; banner hydrate precedes its EventSource open by ~14ms (await-before-SSE fix verified live); published `instance.cascade.queue_saturated` / `_drained` / `_timeout` / `_replaced` and `events.<ws>.schedule.missed` over NATS — banner renders/clears, toasts fire with correct formatting, `/schedules` auto-switches to Missed tab and renders new row with status=pending. Zero console errors. Full QA log in [this comment](https://github.com/friday-platform/friday-studio/pull/206#issuecomment-4384157963).

### DIFF
--- a/apps/atlasd/routes/workspace-events.ts
+++ b/apps/atlasd/routes/workspace-events.ts
@@ -26,6 +26,16 @@ const ParamSchema = z.object({ workspaceId: z.string() });
 const QuerySchema = z.object({ limit: z.coerce.number().int().positive().max(500).optional() });
 
 /**
+ * Adds `stream=true` for the SSE feed branch on top of the replay-only
+ * QuerySchema. Mirrors `apps/atlasd/routes/instance-events.ts` so the two
+ * event surfaces (instance-wide cascade events; workspace operational
+ * events) consume the same shape.
+ */
+const ListQuerySchema = QuerySchema.extend({ stream: z.coerce.boolean().optional() });
+
+const enc = new TextEncoder();
+
+/**
  * Body for `POST /api/events/fire` and `/dismiss`. The composite
  * (workspaceId, signalId, scheduledAt) uniquely identifies a manual
  * `schedule.missed` event — same shape used to derive the KV state
@@ -79,16 +89,90 @@ export const workspaceEventsRoutes = daemonFactory
  * Global feed — every workspace's recent events in one list. Drives
  * the top-level `/schedules` page; per-workspace pages use the
  * `/api/workspaces/:workspaceId/events` variant above.
+ *
+ * Two read modes on the same path:
+ *   - `?stream=true` — SSE feed; subscribes to `events.>` and forwards
+ *                      each entry as a `data:` frame. The /schedules
+ *                      UI listens here for live updates so it doesn't
+ *                      poll the replay endpoint every minute.
+ *   - default        — paginated replay (newest first) for first-load
+ *                      hydration and reload-after-disconnect. Resolves
+ *                      manual-event KV state and runs the pending
+ *                      self-heal — the SSE feed only forwards new
+ *                      publishes and skips that work.
  */
 export const eventsRoutes = daemonFactory
   .createApp()
-  .get("/", zValidator("query", QuerySchema), async (c) => {
-    const { limit } = c.req.valid("query");
+  .get("/", zValidator("query", ListQuerySchema), async (c) => {
+    const { limit, stream } = c.req.valid("query");
     const ctx = c.get("app");
     const nc = ctx.daemon.getNatsConnection();
     if (!nc) return c.json({ error: "NATS connection not ready" }, 503);
-    const events = await listAllWorkspaceEvents(nc, { ...(limit !== undefined ? { limit } : {}) });
-    return c.json({ events }, 200);
+
+    if (!stream) {
+      const events = await listAllWorkspaceEvents(nc, {
+        ...(limit !== undefined ? { limit } : {}),
+      });
+      return c.json({ events }, 200);
+    }
+
+    // SSE branch — subscribe to the raw NATS subject and forward each
+    // event JSON as a `data:` frame. Operator-action lifecycle (the
+    // KV-backed `status` / `pending` / `actionedAt` fields resolved by
+    // the replay path) is intentionally NOT joined here: stream
+    // entries are immutable, so the UI hydrates lifecycle from the
+    // replay endpoint at first load and treats the SSE feed as a feed
+    // of new publishes only. After a fire/dismiss, the UI updates
+    // optimistically — the KV write is the source of truth, the
+    // stream just carries the original event.
+    const sub = nc.subscribe(`events.>`);
+    // The `subscribe()` call returns before the broker actually
+    // registers the subscription. Without this flush, an event
+    // published in the window between subscribe-returns and
+    // broker-registers would be silently dropped. Once the flush
+    // resolves the subscription is live server-side and the SSE
+    // handshake honestly means "from now on, you get every event."
+    await nc.flush();
+
+    const body = new ReadableStream<Uint8Array>({
+      start(controller) {
+        c.req.raw.signal.addEventListener("abort", () => {
+          try {
+            sub.unsubscribe();
+          } catch {
+            // already gone
+          }
+          try {
+            controller.close();
+          } catch {
+            // already closed
+          }
+        });
+
+        void (async () => {
+          try {
+            for await (const msg of sub) {
+              const payload = msg.string();
+              controller.enqueue(enc.encode(`data: ${payload}\n\n`));
+            }
+          } catch {
+            // subscription closed — fall through to controller.close()
+          } finally {
+            try {
+              controller.close();
+            } catch {
+              // already closed
+            }
+          }
+        })();
+      },
+    });
+
+    return c.body(body, 200, {
+      "Content-Type": "text/event-stream",
+      "Cache-Control": "no-cache",
+      Connection: "keep-alive",
+    });
   })
   /**
    * Fire a pending manual `schedule.missed` event. Triggers the
--- a/apps/atlasd/routes/workspace-events.ts
+++ b/apps/atlasd/routes/workspace-events.ts
@@ -156,8 +156,18 @@ export const eventsRoutes = daemonFactory
               controller.enqueue(enc.encode(`data: ${payload}\n\n`));
             }
           } catch {
-            // subscription closed — fall through to controller.close()
+            // for-await exited early — most often because the SSE
+            // controller threw in `enqueue` after the consumer cancelled
+            // (e.g. browser closed the EventSource). The abort listener
+            // normally tears down the NATS subscription on its own, but
+            // controller-cancel can land first; defense-in-depth in the
+            // finally below ensures we never leak the subscription.
           } finally {
+            try {
+              sub.unsubscribe();
+            } catch {
+              // already gone — abort listener fired first
+            }
             try {
               controller.close();
             } catch {
--- a/apps/atlasd/routes/workspace-events.ts
+++ b/apps/atlasd/routes/workspace-events.ts
@@ -125,6 +125,15 @@ export const eventsRoutes = daemonFactory
     // of new publishes only. After a fire/dismiss, the UI updates
     // optimistically — the KV write is the source of truth, the
     // stream just carries the original event.
+    // Core NATS subscription, not a JetStream consumer — the daemon
+    // re-subscribes automatically on reconnect, but messages published
+    // while the daemon's NATS connection was disconnected are LOST.
+    // Acceptable here because the playground is a dev tool and the
+    // replay endpoint covers reload-after-disconnect; the operator-
+    // action POSTs invalidate the query and refetch on every flip.
+    // For a production-grade ops UI an ordered ephemeral push consumer
+    // (DeliverPolicy=New) would survive disconnects with replay, at
+    // the cost of one consumer per HTTP client.
     const sub = nc.subscribe(`events.>`);
     // The `subscribe()` call returns before the broker actually
     // registers the subscription. Without this flush, an event
--- a/apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
+++ b/apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
@@ -197,22 +197,41 @@ describe("createMessageHandler", () => {
       },
     );
     const handler = createMessageHandler("ws-test", triggerFn, registry);
+    const appendEventSpy = vi.spyOn(registry, "appendEvent");
 
     const turn1Buffer = registry.createStream("chat-race");
-    // Model the race: turn 2 replaces the chatId-keyed buffer BEFORE turn 1's
-    // handler runs (i.e. while turn 1 was still awaiting subscribe / append).
-    const turn2Buffer = registry.createStream("chat-race");
+
+    // Model the actual race window: turn 2 replaces the chatId-keyed buffer
+    // *during* turn 1's `await ChatStorage.appendMessage()` call, not before
+    // the handler starts. This is the production race — the swap happens
+    // between `thread.subscribe` and the buffer-capture line where the bug
+    // fixed here used to be visible.
+    let turn2Buffer: ReturnType<typeof registry.createStream> | undefined;
+    mockAppendMessage.mockImplementationOnce(() => {
+      turn2Buffer = registry.createStream("chat-race");
+      return Promise.resolve({ ok: true });
+    });
 
     const thread = makeThread("chat-race");
     await handler(thread, makeMessage({ threadId: "chat-race", raw: { turnBuffer: turn1Buffer } }));
 
-    // Turn 1's late chunk fires after turn 2 owns the chatId. Pre-fix this
-    // would have appended into turn2Buffer because getStream returned it;
-    // post-fix the handler used turn1Buffer from raw, and the identity
-    // check at appendEvent drops the chunk.
+    // Turn 1's late chunk fires after turn 2 owns the chatId.
     captureChunkCallback?.({ type: "text-delta", delta: "from-turn-1" });
 
-    expect(turn2Buffer.events).toEqual([]);
+    // Positive: the handler called `appendEvent` with `turn1Buffer` as the
+    // `expectedBuffer`. Pre-fix the third arg was `getStream(chatId)`, which
+    // would resolve to `turn2Buffer` by now. This catches a regression that
+    // disables the tap entirely (in which case the negative-only assertions
+    // below would still pass).
+    expect(appendEventSpy).toHaveBeenCalledWith(
+      "chat-race",
+      expect.objectContaining({ type: "text-delta" }),
+      turn1Buffer,
+    );
+    // Negative: the chunk lands nowhere — `appendEvent`'s identity check
+    // rejected it because the registry's current buffer is turn 2.
+    expect(turn1Buffer.events).toEqual([]);
+    expect(turn2Buffer?.events).toEqual([]);
   });
 
   it("does not finish a follow-up turn's buffer when the prior turn cleans up", async () => {
--- a/apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
+++ b/apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
@@ -102,11 +102,11 @@ describe("createMessageHandler", () => {
     };
 
     const thread = makeThread("chat-abc");
-    registry.createStream("chat-abc");
+    const turnBuffer = registry.createStream("chat-abc");
 
     await handler(
       thread,
-      makeMessage({ id: "m-1", text: "Hi", threadId: "chat-abc", raw: { datetime } }),
+      makeMessage({ id: "m-1", text: "Hi", threadId: "chat-abc", raw: { datetime, turnBuffer } }),
     );
 
     expect(thread.subscribe).toHaveBeenCalled();
@@ -162,7 +162,10 @@ describe("createMessageHandler", () => {
     const thread = makeThread("chat-cross");
     const turn1Buffer = registry.createStream("chat-cross");
 
-    await handler(thread, makeMessage({ threadId: "chat-cross" }));
+    await handler(
+      thread,
+      makeMessage({ threadId: "chat-cross", raw: { turnBuffer: turn1Buffer } }),
+    );
 
     // Simulate a follow-up turn replacing the buffer mid-flight, then a
     // late producer from turn 1 emits one more chunk.
@@ -171,6 +174,45 @@ describe("createMessageHandler", () => {
 
     expect(turn1Buffer.events).toEqual([]); // turn 1's buffer was closed by createStream replacement
     expect(turn2Buffer.events).toEqual([]); // and turn 1's stale chunk did NOT leak into turn 2
+  });
+
+  it("captures turnBuffer from message.raw, not via getStream, to defeat subscribe-window race", async () => {
+    // The race: handler awaits `thread.subscribe()` and `ChatStorage.appendMessage()`
+    // before capturing its buffer. If a follow-up POST replaces the chatId-keyed
+    // buffer in that window, `streamRegistry.getStream(chatId)` would hand back
+    // the next turn's buffer — turn 1's chunks then land under turn 2's identity
+    // and leak in as orphan deltas. Stashing the buffer on `Message.raw.turnBuffer`
+    // at dispatch makes capture deterministic.
+    const registry = new StreamRegistry();
+    let captureChunkCallback: ((chunk: unknown) => void) | undefined;
+    const triggerFn = vi.fn(
+      (
+        _signalName: string,
+        _payload: Record<string, unknown>,
+        _streamId: string,
+        onStreamEvent: (chunk: unknown) => void,
+      ) => {
+        captureChunkCallback = onStreamEvent;
+        return Promise.resolve({ sessionId: "sess-1" });
+      },
+    );
+    const handler = createMessageHandler("ws-test", triggerFn, registry);
+
+    const turn1Buffer = registry.createStream("chat-race");
+    // Model the race: turn 2 replaces the chatId-keyed buffer BEFORE turn 1's
+    // handler runs (i.e. while turn 1 was still awaiting subscribe / append).
+    const turn2Buffer = registry.createStream("chat-race");
+
+    const thread = makeThread("chat-race");
+    await handler(thread, makeMessage({ threadId: "chat-race", raw: { turnBuffer: turn1Buffer } }));
+
+    // Turn 1's late chunk fires after turn 2 owns the chatId. Pre-fix this
+    // would have appended into turn2Buffer because getStream returned it;
+    // post-fix the handler used turn1Buffer from raw, and the identity
+    // check at appendEvent drops the chunk.
+    captureChunkCallback?.({ type: "text-delta", delta: "from-turn-1" });
+
+    expect(turn2Buffer.events).toEqual([]);
   });
 
   it("does not finish a follow-up turn's buffer when the prior turn cleans up", async () => {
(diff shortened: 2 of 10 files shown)


apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
```diff
@@ -878,14 +878,23 @@ export function createMessageHandler(
...
+    // late events from this turn into it. Non-web adapters don't set
+    // `turnBuffer`, so `ownBuffer` stays undefined and the `appendEvent`
+    // tap below short-circuits, matching prior behavior.
+    const ownBuffer =
+      typeof message.raw === "object" &&
+      message.raw !== null &&
+      "turnBuffer" in message.raw &&
+      typeof message.raw.turnBuffer === "object" &&
+      message.raw.turnBuffer !== null
+        ? (message.raw.turnBuffer as StreamBuffer)
```
ljagiello: **Important — `as StreamBuffer` cast bypasses the actual type check.**

Runtime narrowing only verifies `typeof === "object" && !== null`. The `as StreamBuffer` then asserts arbitrary shape conformance on a field typed `unknown`. CLAUDE.md forbids `as any` and this is the same shape-shifting escape hatch under a different name.

Failure mode if a buggy adapter stuffs the wrong shape: identity check in `appendEvent` always mismatches → events silently dropped → debuggable but quiet.

Fix: add a structural check before the cast, e.g.:

```ts
"chatId" in message.raw.turnBuffer &&
"subscribers" in message.raw.turnBuffer &&
"active" in message.raw.turnBuffer
```

Or stash via a `WeakMap<Message, StreamBuffer>` on the daemon to avoid the `unknown` round-trip entirely.

tools/agent-playground/src/routes/schedules/+page.svelte
```diff
@@ -60,18 +61,77 @@
...
+  // re-stamps a sanitized `id`) and the SSE feed (which delivers the
+  // raw event without an `id`). Going through (workspaceId, signalId,
+  // scheduledAt) bypasses the format mismatch.
+  function naturalKey(e: ScheduleMissedEvent): string {
+    return `${e.workspaceId}:${e.signalId}:${e.scheduledAt}`;
+  }
+
+  // Live updates via SSE. EventSource auto-reconnects on transient
+  // disconnects; on reconnect we don't replay missed events here —
+  // operator action endpoints already invalidate the query and refetch
```
ljagiello: **Important — SSE events arriving before replay resolves are clobbered.**

The replay `createQuery` and this SSE `$effect` both start on mount in parallel. SSE events that land between mount and replay-resolve are merged via `setQueryData` (good). But when the replay's `queryFn` resolves, TanStack writes its result over the cache — losing any SSE-merged events that aren't in the snapshot's window (publish that landed after the JetStream backwards-walk read but before the SSE event reached this merge).

Mirrors the exact race the `CascadeStatusBanner` already solves by awaiting hydrate before opening EventSource. Cheapest fix here: same pattern — gate this `$effect` on `eventsQuery.isSuccess` (or await the query before `new EventSource(...)`).

apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
```diff
@@ -173,6 +176,45 @@ describe("createMessageHandler", () => {
...
+        _signalName: string,
+        _payload: Record<string, unknown>,
+        _streamId: string,
+        onStreamEvent: (chunk: unknown) => void,
+      ) => {
+        captureChunkCallback = onStreamEvent;
+        return Promise.resolve({ sessionId: "sess-1" });
+      },
+    );
+    const handler = createMessageHandler("ws-test", triggerFn, registry);
```
ljagiello, line 199: **Important — assertion is negative-only; admits silent-swallow regressions.**

The test only asserts `turn2Buffer.events).toEqual([])`. Where the chunk *did* go is never observed, so the test stays green if a future regression disables the `appendEvent` tap entirely (e.g. wrong-direction tweak of `if (!ownBuffer) return`). The fix's actual mechanism is the identity-check drop in `stream-registry.ts:216` and that mechanism isn't directly exercised by an assertion.

Either: spy on `streamRegistry.appendEvent` and assert it was called with `expectedBuffer === turn1Buffer`, or assert `turn1Buffer.events).toEqual([])` too (chunk landed nowhere → must have been dropped at the identity check).

apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
```diff
@@ -173,6 +176,45 @@ describe("createMessageHandler", () => {
...
+
+    const turn1Buffer = registry.createStream("chat-race");
+    // Model the race: turn 2 replaces the chatId-keyed buffer BEFORE turn 1's
+    // handler runs (i.e. while turn 1 was still awaiting subscribe / append).
+    const turn2Buffer = registry.createStream("chat-race");
+
+    const thread = makeThread("chat-race");
+    await handler(thread, makeMessage({ threadId: "chat-race", raw: { turnBuffer: turn1Buffer } }));
```
ljagiello, line 216: **Important — test timeline does not match the production race window.**

The bug fixed here is that turn 2 replaces the chatId-keyed buffer *between* `await thread.subscribe()` and `await ChatStorage.appendMessage()`. This test creates `turn2Buffer` *before* `await handler(...)`, so the replacement happens before turn 1's awaits even begin — a strictly earlier scenario.

A hypothetical impl that captured `getStream(chatId)` *before* the awaits (still racy in production but earlier than the actual bug) would also pass.

The adjacent `does not finish a follow-up turn's buffer` test (line ~218) uses the correct mid-flight pattern via `triggerFn`. Same shape here:

```ts
mockAppendMessage.mockImplementationOnce(async () => {
  registry.createStream("chat-race"); // turn 2 replaces the buffer mid-await
  return { ok: true };
});
```

apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
```diff
@@ -897,6 +906,12 @@ export function createMessageHandler(
       { chatId, userId, streamId, datetime, foregroundWorkspaceIds },
       streamId,
       (chunk: unknown) => {
+        // Non-web adapters (Slack, Telegram, Teams, etc.) never create a
+        // registry buffer — `appendEvent` would always return false and
+        // emit a `stream_event_dropped` warn for every chunk, drowning
+        // out the real drops the line is meant to flag. Skip the call
+        // entirely for those adapters.
+        if (!ownBuffer) return;
```
ljagiello, line 917: **Nit — short-circuit and warn-removal both look right.**

Verified: pre-PR the entire callback body was `if (isClientSafeEvent(chunk)) { appendEvent(...); if (!appended) logger.warn(...) }`. Adding `if (!ownBuffer) return;` at the top drops only the `appendEvent` + warn for adapters that never created a buffer. No upstream side-effects are lost. Strict signal-to-noise improvement on the `stream_event_dropped` warn — it now fires only when a turn that *did* create a buffer drops a chunk (i.e. real identity-mismatch case).

apps/atlasd/routes/workspace-events.ts
```diff
@@ -79,16 +89,109 @@ export const workspaceEventsRoutes = daemonFactory
...
+    // Core NATS subscription, not a JetStream consumer — the daemon
+    // re-subscribes automatically on reconnect, but messages published
+    // while the daemon's NATS connection was disconnected are LOST.
+    // Acceptable here because the playground is a dev tool and the
+    // replay endpoint covers reload-after-disconnect; the operator-
+    // action POSTs invalidate the query and refetch on every flip.
+    // For a production-grade ops UI an ordered ephemeral push consumer
+    // (DeliverPolicy=New) would survive disconnects with replay, at
+    // the cost of one consumer per HTTP client.
+    const sub = nc.subscribe(`events.>`);
```
ljagiello, line 137: **Note — small subscription-leak window before `start(controller)`.**

`nc.subscribe(...)` runs *before* the `ReadableStream` is constructed; the abort listener registers inside `start(controller)`. If the client aborts during `await nc.flush()` or between subscribe and `start` firing, there's no cleanup hook — the subscription stays open.

In practice Hono invokes `start` once the Response body is consumed, so the window is microseconds for subscribe + one round-trip for flush, on a local dev daemon. Defer-OK for this PR (dev tool, tiny window). Worth folding into the deferred slow-consumer + integration-tests follow-up — moving `subscribe` inside `start` makes the lifecycle bound by the stream's lifecycle.

Same applies to `apps/atlasd/routes/instance-events.ts:62-70`.

ljagiello (commented):
## Review — Needs Work

Race fix mechanism is **correct** — walked through every interleaving (in-turn live append, post-`finishStreamIfCurrent` drop, post-`createStream` swap, subscriber stomp via `closeSubscribers`) and the `appendEvent` identity check holds. Nice.

Blockers before merge are bounded — three Important items with cheap fixes:

1. **`as StreamBuffer` cast** on `message.raw.turnBuffer` (chat-sdk-instance.ts:896) — same shape-shifting hazard global CLAUDE.md forbids `as any` for. One-line structural check removes it.
2. **`/schedules` SSE-vs-replay race** mirrors the one the cascade banner already fixes — replay's `queryFn` overwrites SSE-merged events on resolve. Mirror the hydrate-then-subscribe ordering.
3. **Race regression test** asserts negative-only AND models turn 2 replacement *before* turn 1's awaits begin — neither matches the actual race window or the fix mechanism. Spy on `appendEvent` or assert `turn1Buffer.events` is also empty; move `createStream` into `mockAppendMessage.mockImplementationOnce`.

Detailed findings inline. SSE-route subscription leak window (subscribe before `start(controller)`), missing `refetchInterval` backstop, banner SSE-during-handshake gap, and missing tests for new SSE branch + banner are noted but defer-OK as dev-tool scope — file a tracking issue for the deferred SSE integration tests so the deferral doesn't expire silently.

Full writeup: `docs/reviews/2026-05-05-pr-206-chat-race-sse-feeds.md`

basedfriday:
## Live-daemon QA — passed end-to-end

Rebased onto latest `main`, started daemon from this worktree + playground, drove via `agent-browser` and published synthetic events on NATS.

### Backend wiring
- `GET /api/instance/events?stream=true` → 200, `text/event-stream`, holds open ✓
- `GET /api/events?stream=true` → 200, `text/event-stream`, holds open ✓
- Replay endpoints unchanged (200 JSON) ✓

### Frontend wiring
On page load, request order is:
```
8277.349 GET /api/daemon/api/instance/events?type=cascade.&limit=50  (Fetch)        — banner hydrate
8277.352 GET /api/daemon/api/events?limit=50                          (Fetch)        — schedules replay
8277.353 GET /api/daemon/api/events?stream=true                       (EventSource)  — schedules SSE
8277.363 GET /api/daemon/api/instance/events?stream=true              (EventSource)  — banner SSE
```
The banner's hydrate (entry 1) precedes its EventSource open (entry 4) by ~14ms — confirms the v2 await-before-SSE fix is live, not just a unit-test claim. No console errors across navigation.

### End-to-end event delivery (published via `nats pub`)
- `instance.cascade.queue_saturated` → banner appears: *"Cascade queue saturated — 20/32 in flight, 17 queued (deepest: qa-test:smoke-signal)"*
- `instance.cascade.queue_drained` → banner clears
- `instance.cascade.queue_timeout` (`queuedMs: 312000`) → red toast: *"Cascade queue timeout — qa-test / slow-signal sat 5m before pickup"* (5m formatting correct)
- `instance.cascade.replaced` → toast: *"Cascade replaced — qa-test / replaced-signal aborted in favor of newer envelope"*
- `events.qa-test.schedule.missed` (manual) → `/schedules` updates live: "Missed 1" badge appears, auto-switches to Missed tab via the existing `pendingEvents > 0` effect, new row rendered with status=pending and Trigger/Dismiss buttons. No duplicate row from the replay-vs-SSE dedup path.

Test plan checkbox in the PR description is now satisfied.

basedfriday:
Thanks for the thorough pass — three Important items addressed in `a5a9aad`. Walking through each:

### 1. `as StreamBuffer` cast → structural type guard ✓ Agreed

You're right that the `typeof === "object" && !== null` narrowing followed by `as StreamBuffer` is the same shape-shifting hazard `as any` is — the compiler trusts the cast even when the source is `unknown`. The silent failure mode (identity check in `appendEvent` always mismatches → events drop with no signal) is exactly what would have made this hard to debug.

Added `isStreamBuffer` exported from `stream-registry.ts` checking `chatId` / `active` / `events` / `subscribers` / `openParts` shape. Replaced the cast with the guard at the boundary in `chat-sdk-instance.ts`. Considered the WeakMap path you suggested but the structural-check approach is more surgical and matches the existing `abortSignal` narrowing pattern in the same file.

### 2. /schedules SSE-vs-replay race ✓ Agreed

Yep, exactly the same race the cascade banner already solved — the replay's `queryFn` resolving overwrites SSE-merged events that landed first. Added `if (!eventsQuery.isSuccess) return;` to the `$effect`, so the EventSource doesn't open until the replay has seeded the cache. Now mirrors the hydrate-then-subscribe ordering from `CascadeStatusBanner`.

### 3. Race regression test rewrite ✓ Agreed (combined both inline comments)

Both your points were on target:
- The test created `turn2Buffer` *before* `await handler(...)` — so a hypothetical implementation that captured `getStream(chatId)` *before* the awaits (still racy in production but earlier than the actual bug) would have passed. Moved the buffer swap into `mockAppendMessage.mockImplementationOnce` so it lands during turn 1's await window — the actual production race window the fix targets.
- Added a positive `appendEventSpy` assertion that the handler called `appendEvent` with `turn1Buffer` as `expectedBuffer` — that directly tests the fix mechanism. The previous negative-only assertion would have stayed green if a regression disabled the tap entirely. Also added `turn1Buffer.events).toEqual([])` for the chunk-landed-nowhere check.

### 4. `if (!ownBuffer) return` short-circuit (nit) — Acknowledged, no change

Thanks for verifying the signal-to-noise improvement on `stream_event_dropped`.

### 5. Subscription-leak window before `start(controller)` (note) — Folded into tracking issue

Agreed it's defer-OK for this PR. Folded into [#212](https://github.com/friday-platform/friday-studio/issues/212) alongside the slow-consumer mitigation and SSE integration tests, so the deferral doesn't expire silently.

Ready for another look.

apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
basedfriday: Agreed — cast bypassed exactly the validation it was supposed to provide. Fixed in a5a9aad: added `isStreamBuffer` exported from `stream-registry.ts` (checks `chatId` / `active` / `events` / `subscribers` / `openParts` shape), and replaced the `as StreamBuffer` cast with the guard at the boundary in `chat-sdk-instance.ts`. Considered the WeakMap path — surgical structural-check is more local and matches the existing `abortSignal` narrowing in the same file.

tools/agent-playground/src/routes/schedules/+page.svelte
basedfriday: Agreed — same race the cascade banner already solves. Fixed in a5a9aad: added `if (!eventsQuery.isSuccess) return;` to the `$effect` so the EventSource doesn't open until the replay's `queryFn` has resolved and seeded the cache. Mirrors the hydrate-then-subscribe ordering from `CascadeStatusBanner`.

apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
basedfriday: Agreed — negative-only assertion would have stayed green if a regression disabled the tap entirely. Fixed in a5a9aad: added `appendEventSpy = vi.spyOn(registry, "appendEvent")` and asserted it was called with `turn1Buffer` as the `expectedBuffer` (the third arg). That directly tests the fix mechanism — pre-fix the third arg was `getStream(chatId)` which would resolve to `turn2Buffer` by that point. Also kept `turn1Buffer.events).toEqual([])` and `turn2Buffer?.events).toEqual([])` to confirm the chunk landed nowhere.

apps/atlasd/src/chat-sdk/chat-sdk-instance.test.ts
basedfriday: Agreed — turn-2 swap before `await handler(...)` doesn't model the production race window. Fixed in a5a9aad using exactly the pattern you suggested:

```ts
mockAppendMessage.mockImplementationOnce(() => {
  turn2Buffer = registry.createStream("chat-race");
  return Promise.resolve({ ok: true });
});
```

Now the buffer swap lands during turn 1's `await ChatStorage.appendMessage()` — the actual race window the fix targets. (The arrow's not `async` because deno-lint flagged async-without-await; `Promise.resolve` keeps the same semantics.)

apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
basedfriday: Thanks for verifying — appreciated.

apps/atlasd/routes/workspace-events.ts
basedfriday: Agreed it's defer-OK for this PR. Folded into [#212](https://github.com/friday-platform/friday-studio/issues/212) alongside the slow-consumer mitigation and SSE integration tests, with the `subscribe`-inside-`start` lifecycle suggestion in scope. Same applies to `instance-events.ts` — both routes covered.

ljagiello (approved):

PR #215, opened 2026-05-06
fix(workspace): merge atlas agent config prompt with action prompt

odk-:
## Summary

- Atlas-typed workspace agents had their `prompt:` config silently dropped whenever the calling FSM action also set a `prompt:` — `buildFinalAgentPrompt` used `actionPrompt || configPrompt`, so action always won. (LLM-typed workspace agents already concatenate both via `expand-agent-actions.ts`; atlas-typed agents were inconsistent.)
- Concatenate the two prompts (config first, then action) and extract the inline composition in `WorkspaceRuntime.executeAgent` into a new `composeAgentPrompt` helper so the wiring is unit-testable.
- Add a model-side eval at `tools/evals/agents/agent-config-prompt/` that sends both layers to Haiku and asserts the LLM emits a marker token defined in the config prompt.

## Motivating bug

In the `select_noodle` workspace, the `pixel-asset-generator` (atlas/`image-generation`) had a `prompt:` instructing it to "use neon green (#00FF00) background". The FSM's `generate` state set its own per-step `prompt:` ("Generate a single pixel art game asset. Description: …"). Pre-fix, only the action prompt reached the LLM and the generated images had random backgrounds.

## Test plan

- [x] `npx vitest run apps/atlasd/src/agent-helpers.test.ts` — 29 tests pass (was 23; +6 for `composeAgentPrompt`, +0 net for `buildFinalAgentPrompt` after rewriting precedence cases as composition cases)
- [x] `deno check apps/atlasd/src/agent-helpers.ts packages/workspace/src/runtime.ts tools/evals/agents/agent-config-prompt/agent-config-prompt.eval.ts` clean
- [x] `deno run -A npm:@biomejs/biome check` clean for all touched files
- [x] `deno task evals run -t tools/evals/agents/agent-config-prompt/agent-config-prompt.eval.ts` (run before merging — needs ANTHROPIC creds)

## Files

- `apps/atlasd/src/agent-helpers.ts` — `buildFinalAgentPrompt` now concatenates; new `composeAgentPrompt` helper composes extract → interpolate → concat
- `apps/atlasd/src/agent-helpers.test.ts` — rewrote precedence cases as composition cases; added `composeAgentPrompt` suite
- `packages/workspace/src/runtime.ts` — `executeAgent` delegates to `composeAgentPrompt`
- `packages/fsm-engine/{schema,types}.ts` — doc comments updated to match new behavior
- `tools/evals/agents/agent-config-prompt/` — new eval

## Scope notes

- The eval (and the `composeAgentPrompt` unit tests) cover the helper, not the runtime call site. A regression that bypasses `composeAgentPrompt` inside `runtime.executeAgent` would not be caught — but the helper is now the single delegation point, so any such change would be obvious in code review.
- Backward compatibility: workspaces that intentionally relied on the action prompt overriding the config prompt will now see both. This matches the existing `type: llm` behavior.

### DIFF
--- a/apps/atlasd/src/agent-helpers.test.ts
+++ b/apps/atlasd/src/agent-helpers.test.ts
@@ -1,14 +1,13 @@
 /**
- * Tests for agent prompt precedence + output validation.
+ * Tests for agent prompt composition + output validation.
  *
  * These tests verify:
- * 1. action.prompt takes priority over agentConfig.prompt
- * 2. If no action.prompt, falls back to agentConfig.prompt
- * 3. If neither exists, returns context only
- * 4. validateAgentOutput hallucination-detection branching
+ * 1. agentConfig.prompt and action.prompt are concatenated (config first) by buildFinalAgentPrompt
+ * 2. composeAgentPrompt wires extract + interpolate + concat in the same order as runtime.executeAgent
+ * 3. validateAgentOutput hallucination-detection branching
  */
 
-import type { AgentResult } from "@atlas/agent-sdk";
+import type { AgentResult, AtlasAgentConfig } from "@atlas/agent-sdk";
 import type { Context } from "@atlas/fsm-engine";
 import {
   ValidationFailedError,
@@ -19,12 +18,43 @@ import type { PlatformModels } from "@atlas/llm";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import {
   buildFinalAgentPrompt,
+  composeAgentPrompt,
   extractAgentConfigPrompt,
   validateAgentOutput,
 } from "./agent-helpers.ts";
 
+// `@atlas/fsm-engine`'s `mod.ts` transitively pulls in Deno-only modules, so
+// `vi.importActual` fails under vitest's Node runtime. Stub the two functions
+// the helpers call: `expandArtifactRefsInDocuments` becomes a passthrough,
+// and `interpolatePromptPlaceholders` does the same `{{path}}` substitution
+// the real implementation does (the substring-substitution contract is pinned
+// in `packages/fsm-engine/tests/prompt-interpolation.test.ts`; this stub
+// just covers enough for `composeAgentPrompt` to round-trip its inputs).
 vi.mock("@atlas/fsm-engine", () => ({
   expandArtifactRefsInDocuments: vi.fn((docs: unknown[]) => Promise.resolve(docs)),
+  interpolatePromptPlaceholders: (
+    prompt: string,
+    prepareResult: { config?: Record<string, unknown> } | undefined,
+  ): string => {
+    const config = prepareResult?.config ?? {};
+    const scopes: Record<string, unknown> = { inputs: config, config, signal: { payload: config } };
+    const re =
+      /\{\{\s*([a-zA-Z_][a-zA-Z0-9_.]*)\s*(?:\|\s*default\s*:\s*(?:'([^']*)'|"([^"]*)"))?\s*\}\}/g;
+    return prompt.replace(re, (original, path, sq, dq) => {
+      const fallback: string | undefined = sq ?? dq;
+      let cursor: unknown = scopes;
+      for (const segment of String(path).split(".")) {
+        if (cursor === null || cursor === undefined || typeof cursor !== "object") {
+          return fallback ?? original;
+        }
+        cursor = (cursor as Record<string, unknown>)[segment];
+      }
+      if (cursor === undefined || cursor === null) return fallback ?? original;
+      if (typeof cursor === "string") return cursor === "" ? (fallback ?? cursor) : cursor;
+      if (typeof cursor === "number" || typeof cursor === "boolean") return String(cursor);
+      return JSON.stringify(cursor);
+    });
+  },
 }));
 
 const mockValidate = vi.hoisted(() => vi.fn<(...args: unknown[]) => Promise<ValidationVerdict>>());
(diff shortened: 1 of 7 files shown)


apps/atlasd/src/agent-helpers.test.ts
```diff
@@ -19,12 +18,43 @@ import type { PlatformModels } from "@atlas/llm";
...
 vi.mock("@atlas/fsm-engine", () => ({
   expandArtifactRefsInDocuments: vi.fn((docs: unknown[]) => Promise.resolve(docs)),
+  interpolatePromptPlaceholders: (
+    prompt: string,
+    prepareResult: { config?: Record<string, unknown> } | undefined,
+  ): string => {
+    const config = prepareResult?.config ?? {};
+    const scopes: Record<string, unknown> = { inputs: config, config, signal: { payload: config } };
+    const re =
+      /\{\{\s*([a-zA-Z_][a-zA-Z0-9_.]*)\s*(?:\|\s*default\s*:\s*(?:'([^']*)'|"([^"]*)"))?\s*\}\}/g;
```
ljagiello: **Mock-the-implementation, not the boundary.** This regex (and the scope-walking around it on lines 39-56) is byte-equivalent to the real `interpolatePromptPlaceholders` in `packages/fsm-engine/fsm-engine.ts:116-151`. They will drift independently — the new `composeAgentPrompt` "interpolates `{{inputs.x}}`" tests at lines 327 and 339 green-light against the *clone*, not the real function, so they're really verifying the mock's regex.

The Deno-only-import constraint justifies a stub, but the stub should be minimal — e.g. `(p) => p` or a sentinel like `(p) => "INTERP[" + p + "]"` — and the assertion should be "interpolation was invoked", not "interpolation produced this exact string". The real contract is already pinned in `packages/fsm-engine/tests/prompt-interpolation.test.ts`.

tools/evals/agents/agent-config-prompt/agent-config-prompt.eval.ts
```diff
@@ -0,0 +1,190 @@
...
+      // composeAgentPrompt itself — separately from the LLM-side score.
+      // Both layers must appear in the composed prompt before we ask the
+      // model — substring-match the prefix up to the first `{{`, since
+      // composeAgentPrompt interpolates placeholders out.
+      assert: ({ composedPrompt }) => {
+        const requirePrefix = (label: string, raw: string) => {
+          const prefix = raw.split("{{")[0] ?? "";
+          if (prefix && !composedPrompt.includes(prefix)) {
+            throw new Error(
+              `Composed prompt missing ${label} prefix.\nGot: ${JSON.stringify(composedPrompt)}`,
```
ljagiello: **Assertion is more permissive than the comment promises.** `prefix = raw.split("{{")[0]` means for the `interpolated-action-with-config-rule` case, action prompt `"Describe a {{inputs.subject}} sprite in two sentences."` is checked only as `"Describe a "`. A regression that truncated the composed prompt after the placeholder would still pass this assert. The structural guarantee ("both layers must appear") isn't what's actually checked.

Suggest asserting the post-interpolation result is present (e.g. `Describe a robot chef sprite`) — you already have the `prepareConfig` to reconstruct it.

packages/workspace/src/runtime.ts
```diff
@@ -1608,26 +1604,7 @@ export class WorkspaceRuntime {
...
-    const interpolatedConfigPrompt = interpolatePromptPlaceholders(
-      agentConfigPrompt,
-      prepareResult,
-    );
-    const prompt = buildFinalAgentPrompt(
-      interpolatedActionPrompt,
-      interpolatedConfigPrompt,
-      context,
-    );
+    const prompt = composeAgentPrompt(action, agentConfig, fsmContext.input, context);
```
ljagiello, line 1607: **Runtime call site has no direct test.** PR description acknowledges this. The original bug shape was *inline composition at the call site dropping the config prompt* — exactly what could regress again if a future refactor inlines fields back here.

Existing tests (`runtime-bootstrap-injection.test.ts`, `runtime-session-summary.test.ts`) already mock `AgentOrchestrator.executeAgent` and capture the prompt arg. One assertion ("composed prompt contains config-prompt substring") on a workspace with both layers set would close the gap at low cost. Not a blocker for this PR, but worth a follow-up issue.

ljagiello (commented):
Review (verdict: **Needs Work** — small fixes, not blocking the architecture).

The fix is correct and aligns the atlas-typed path with the existing LLM-typed concatenation in `expand-agent-actions.ts:49`. Three things flagged inline (mock fidelity, eval assertion strength, runtime call-site coverage). One follow-up worth tracking that I can't anchor inline:

**Shipped templates now emit duplicated instructions.** `packages/system/skills/workspace-api/assets/example-jobs-pipeline.yml` (`bookmark-enricher`, lines ~60-67 and ~115-118) and `packages/system/skills/workspace-api/references/agent-types.md` (`news-scout`) both ship a `type: atlas` agent with overlapping config-level `prompt:` and FSM action `prompt:`. Pre-fix the action overrode; post-fix the model receives both. This is the *intended* fix, but these are copy-paste templates — they'll start producing redundant instructions in derived workspaces. Worth a follow-up doc edit so each template demonstrates the new mental model (config = agent-wide guidance, action = per-step task).

**Minor — security widening.** Interpolation now runs against the agent-config prompt too. Signal-payload-controlled `{{inputs.x}}` placeholders authored into system-style config prompts can now be filled from external triggers. Not new in concept (already true for action prompts), but widens the surface from the previously-static config layer. Worth an authoring guideline; not a fix request.

Full review notes in `docs/reviews/2026-05-06-fix-atlas-agent-config-prompt.md` on the review branch.

PR #225, opened 2026-05-08
feat: add durable HITL, artifact refs, validation contracts, and Python input handoff

basedfriday:
## Summary

Adds durable human-in-the-loop workflows, artifact-ref handoffs, validation contracts, and Python user-agent input plumbing for Friday workspaces:

- **Refs-over-data**: agent/FSM outputs persist as artifact refs instead of bulky inline payloads.
- **Fan-out without fan-in**: delegated work stays compact and parent supervisors receive summaries/artifact refs.
- **Safe validation/output contracts**: structured outputs must complete through expected contracts; validation can be skipped/self/external by policy.
- **Real HITL blocking**: tool-access and human-input requests create durable elicitations, block execution, and resume from user answers.
- **Python user-agent handoff**: `inputFrom`/structured action input reaches Python agents through `ctx.input`.
- **Activity/UI coverage**: pending elicitations surface in Activity/sidebar/chat, including nested and repeated-choice prompts.
- **No-auth QA loop**: fake-inbox/no-Gmail eval harness covers these workflow contracts without user-data auth.

This branch also merges current `origin/main` and keeps PR #215's prompt-composition fix while preserving this branch's external-validation architecture.

### Workspace/job config

- `permissions.dangerouslySkipAllowlist`
  - precedence: job > workspace > daemon env `FRIDAY_DANGEROUSLY_SKIP_PERMISSIONS`
- `delegation`
  - `max_depth`
  - `max_steps_per_call`
  - `max_output_tokens`
  - `max_input_tokens`
  - `max_wall_time_ms`
  - `max_cost_usd` reserved
- `validation`
  - `default: auto | skip | self | external`
  - `skill`
- `artifacts.default_grace`
  - default: `24h`
- memory TTL / short-term store support for ephemeral working state

### MCP config

- per-server validation override:
  - `validation: "read-only"`
  - `validation: "mutating"`

### FSM/action config

- `validate`
- `summary`
- `skills`
- `inputFrom`
- `run_code.readOnly`

### New daemon routes

Mounted under `/api/elicitations`:

- `GET /api/elicitations`
  - list elicitations
  - filters: `workspaceId`, `sessionId`, `status`, `kind`
- `GET /api/elicitations/stream`
  - SSE over NATS subjects `elicitations.>` or workspace-filtered subjects
- `GET /api/elicitations/:id`
- `POST /api/elicitations/:id/answer`
- `POST /api/elicitations/:id/decline`

### Changed daemon routes

- Workspace signal endpoint supports internal-only `bypassConcurrency`.
  - Requires `x-friday-internal-signal-bypass` and `FRIDAY_INTERNAL_SIGNAL_BYPASS_TOKEN`.
  - Public callers attempting bypass receive 403.
- Signal payload supports `parentSessionId`.
- Agent registration/process execution now passes active `FRIDAY_NATS_URL` through as `NATS_URL`.
- Chat debug route includes active stream/buffer details useful for blocked HITL debugging.

### Removed routes/surfaces

- Removed scratchpad route and legacy scratchpad platform surface.
- Use memory stores (`memory_save`, `memory_read`, `memory_remove`) for short-term state.

### Streams

- `ELICITATIONS`
  - subjects: `elicitations.<workspaceId>.<sessionId>.<elicitationId>`
  - file-backed, Limits retention
  - per-message `Nats-TTL`
  - `allow_msg_ttl: true`
- `SIGNALS`
  - subjects: `workspaces.*.signals.*`
  - work queue for durable signal processing
- `CASCADES`
  - subjects: `cascades.*.*`
  - dispatch buffer for signal cascades/concurrency policies
- `SESSION_EVENTS`
  - durable session event log
- Existing/updated session ephemeral chunks continue on core NATS subjects `sessions.<sessionId>.ephemeral`.

### KV buckets

- `ELICITATION_STATUS`
  - key: `<elicitationId>`
  - history: 5
- `ARTIFACTS`
  - metadata keys:
    - `<id>/<revision>`
    - `<id>/_latest`
    - `<id>/_deleted`
- Existing/updated session buckets:
  - `SESSION_METADATA`
  - `SESSION_INFLIGHT`
- Migration drops orphaned `KV_SCRATCHPAD`.

### Object Stores

- `OBJ_artifacts`
  - SHA-256 content-addressed artifact blobs
- `OBJ_SKILL_ARCHIVES`
  - uploaded skill archives

### Artifacts / refs-over-data

- Persist large outputs/tool results as artifact refs.
- Prefer compact summaries + artifact ids in parent/supervisor contexts.
- Add durable vs ephemeral artifact lifecycle.
- Add grace-window cleanup and promotion when references are found in memory/session summaries.
- Scrub lifted artifact markers at persistence boundaries.

### Validation and output contracts

- Add action-level validation via `validate`.
- Add workspace/job validation defaults.
- Add self/external validation flow and validation events.
- Add `record_validation` platform tool.
- Add judge-agent external validation path.
- Ensure `outputTo` + `outputType` actions fail loudly when no valid document is produced.
- Preserve main's config-prompt/action-prompt composition fix via `composeAgentPrompt()`.

### HITL / elicitations

- Add durable elicitation storage/model/adapter.
- Add `request_tool_access` and `request_human_input` tools.
- Block execution while waiting for answer/decline/expiry.
- Reuse identical pending human-input requests instead of creating duplicates.
- Durably expire stale pending elicitations.
- Add Activity/sidebar/chat rendering for blocked sessions and nested/repeated choices.

### Signal/runtime behavior

- Split signal ingestion from cascade dispatch.
- Harden internal signal concurrency bypass.
- Preserve blocked HITL runtime/session state from idle cleanup.
- Abort stale daemon streams on UI navigation.

### Python user agents

- Runtime serializes structured action input into user-agent payloads.
- Python SDK `ctx.input` supports structured `inputFrom` hydration.
- Added fake Python user-agent fixture for no-auth eval coverage.

## Agent SDK / dependency notes

- Bumped bundled Python SDK from `friday-agent-sdk==0.1.5` to `0.1.7`.
- Kept SDK pins synchronized across:
  - `Dockerfile`
  - `tools/friday-launcher/paths.go`
  - `apps/studio-installer/src-tauri/src/commands/prewarm_agent_sdk.rs`
- Added `scripts/check-sdk-pin-sync.ts` validation.
- `apps/atlasd` now depends on `@atlas/system` instead of directly depending on inline hallucination validation.
- External validation now routes through system/judge-agent instead of old inline `@atlas/hallucination` usage.
- Core artifact work uses blob/type helpers including `file-type` and `@std/encoding` lockfile updates.

## Migrations

Added:

- `m_20260505_120000_elicitations_bootstrap.ts`
  - provisions `ELICITATIONS` stream and `ELICITATION_STATUS` KV
- `m_20260507_120000_drop_scratchpad_kv.ts`
  - deletes old/orphaned scratchpad KV stream

## Developer notes

- `jobs.<job>.execution.agents[].tools` allow/deny config still parses only; it is not carried through `compileExecutionToFsm()` and has no runtime effect yet. By contrast, MCP server `tools.allow`/`tools.deny`, FSM `type: llm` action `tools`, and `agents.<id>.config.tools` for LLM agents expanded into FSM actions are enforced.
- Top-level FSM `functions` and `tools` blocks have no runtime consumers (tracked in #226). They are still schema-accepted (`packages/fsm-engine/schema.ts` / `types.ts`); `validator.ts` only checks non-empty `code`, and MCP create/validate helpers only report counts. Runtime execution is limited to `llm`, `agent`, `emit`, and `notification` actions. Legacy `type: code` actions are not in `ActionSchema` and are not restored here; live tool access comes from MCP/platform tools narrowed by each `LLMAction.tools` allowlist.
- Scratchpad adapter/tools/routes were removed; the scratchpad migration only cleans old installs.
- Keep agent SDK pins synchronized with `scripts/check-sdk-pin-sync.ts` when bumping.
- No-auth eval loop means no Gmail/user-data auth; real `ANTHROPIC_API_KEY` is still expected for LLM-backed evals.
- `request_human_input.options` is intentionally flat; grouped repeated choices are encoded in labels/values and grouped in the UI.
- If an `outputTo` action must call a tool before `complete`, include that tool in the action `tools:` allowlist.

## QA / validation

Passed after merging `origin/main`:

```bash
deno task test apps/atlasd/src/agent-helpers.test.ts \
  packages/workspace/src/runtime-agent-prompt-composition.test.ts \
  packages/workspace/src/agent-executor-utils.test.ts \
  packages/mcp-server/src/tools/elicitations/wait.test.ts \
  packages/core/src/session/jetstream-session-history-adapter.test.ts
# 5 files passed, 42 tests passed

deno check apps/atlasd/src/agent-helpers.ts \
  packages/workspace/src/runtime.ts \
  packages/fsm-engine/schema.ts \
  packages/fsm-engine/types.ts
# passed

deno run --allow-read scripts/check-sdk-pin-sync.ts
# ✓ All sites pinned to 0.1.7

deno task typecheck
# passed; existing Svelte warnings only
```

Optional follow-up manual verification:

- full live-daemon eval for artifact-ref/HITL/validation workflows
- Python user-agent → Python user-agent `outputTo` / `inputFrom` pipeline
- blocked-HITL navigation after daemon restart


ljagiello (commented):
## Critical findings from review

### C2. Elicitation SSE `/stream` leaks pendingTool args and answers globally when no `workspaceId` filter is supplied

**File:** `apps/atlasd/routes/elicitations/index.ts:111-191` (GET `/stream`)

`workspaceId` is an optional query param. Without it, the handler subscribes to `elicitations.>` and forwards every envelope. Envelopes contain `pendingTool.args` (e.g. email body to send, target of `delete_*`) and answer values once entries flip. Any local listener can passively eavesdrop on the entire HITL stream.

**Recommendation:** Require `workspaceId`; reject the global subscribe (or gate behind the same loopback auth as C1). The "global Activity-page view" is fine for a list endpoint inside the trusted UI process, but a poor default for an SSE feed any local process can hit.

---

### C3. `wait.ts` returns `{status: "pending"}` on a transient KV miss and the wait short-circuits

**File:** `packages/mcp-server/src/tools/elicitations/wait.ts:53-68, 84`

`readTerminalElicitation` collapses "not found" and "pending" into the same return shape (`{status: "pending"}`). In the NATS-subscribed branch, `if (current) return current;` then exits immediately. If the KV read transiently misses (e.g. between create-and-first-read, or eventual-consistency on a follower), `request_tool_access` and `request_human_input` return immediately with `pending_user_approval` / `pending_user_input` and the LLM gets a non-blocking response — defeating HITL semantics.

**Recommendation:** Distinguish "not found" (return error or fall through to the subscribe loop) from genuine `pending` (continue waiting). Only return early for genuinely terminal states.

---

### C4. \`expiresAt\` is derived in \`get()\` but checked-against in \`answer()\` — user clicking Allow at the deadline sees their answer rejected as expired

**File:** `packages/core/src/elicitations/jetstream-adapter.ts:340-365` (and the `get()` derivation path)

`get()` applies `deriveExpired` (returns `status: "expired"` if `expiresAt` has passed, even if durable status is still `pending`). `answer()` calls `this.get(...)`, sees `status: expired`, and rejects with "already in terminal state." A user clicking Allow within milliseconds of the deadline gets a confusing rejection. Worse: if the elicitations sweeper is behind by even a few seconds and a long-pending request is on the cusp, the race is "always rejected post-deadline."

**Recommendation:** In `answer()` and `decline()`, read the raw KV revision without `deriveExpired` and decide based on durable status; or accept a small grace window (e.g. 5s past `expiresAt`) for the user-facing answer path.

basedfriday:
Thanks — these were good catches. I addressed all three in follow-up commits.

### C2 — global elicitation SSE leak

Fixed by making `GET /api/elicitations/stream` require `workspaceId`; the route now subscribes only to `elicitations.<workspaceId>.>` and rejects unscoped/global SSE requests.

I initially tried preserving global live updates by opening one scoped `EventSource` per workspace, but that exhausted browser per-origin request concurrency and starved normal workspace/chat requests. I fixed that too: global sidebar/Activity now use polling for the global elicitation list, while workspace-scoped Activity keeps a single scoped SSE stream.

Also filed follow-up #229 for a proper sanitized global HITL live-update surface that doesn’t expose full envelopes and doesn’t fan out browser SSE connections.

### C3 — waiter short-circuits on transient KV miss

Fixed `readTerminalElicitation()` so a missing KV row returns `null` (continue waiting), not `{ status: "pending" }`. Only terminal envelopes (`answered`, `declined`, `expired`) return a terminal result now.

Added/updated coverage so an initial KV miss keeps waiting and later observes the answer instead of returning `pending_user_*` immediately. Platform-tool tests were updated to match the blocking semantics rather than relying on the old pending short-circuit.

### C4 — derived expiry blocks deadline-edge answers

Adjusted terminal transitions to arbitrate on raw durable KV status instead of the read-time derived `expired` view. There is now a small 5s answer/decline grace window after `expiresAt`; after that, the transition is rejected as expired. This preserves hard-ish deadline semantics without making a deadline-edge user click lose solely because the read projection derived `expired` before the sweeper durably flipped KV.

Added coverage for answering during the grace window.

PR #232, opened 2026-05-08
docs(setup): hint at Deno installer's interactive shell prompt

### LINKED ISSUE #231: Clarify interactive [ ] bash selection during installation
### What would you like?

Hi! While going through the installation steps, I got confused when this appeared:

[ ] bash

There weren’t any instructions explaining how to interact with this prompt, so I got stuck for a while.

It would be really helpful to add a short explanation for users who may not be familiar with terminal-based interactive selectors. For example:

Use Space to select an option
Use arrow keys to navigate
Press Enter to confirm

A tiny note like this near the screenshot/output would make the installation process much clearer and smoother for beginners.

Thanks for the project and documentation!

## Environment

- OS: Nobara Linux
- Terminal: Konsole
- Shell: bash 
- Node.js version: v24.15.0

### Anything else?

Image

LissaGreense:
## Summary
- Print a one-line hint in `setup-dev-env.sh` right before `curl ... | sh` runs the Deno installer, explaining how to interact with the `[ ] bash` checkbox prompt (↑/↓ to navigate, Space to toggle, Enter to confirm).
- Only fires when Deno is actually missing — silent for users who already have it installed.

Closes #231

## Test plan
- [ ] Run `bash scripts/setup-dev-env.sh` on a machine without Deno → hint appears before the installer's TUI.
- [ ] Run on a machine with Deno already at >= MIN → hint does not appear.

### DIFF
--- a/scripts/setup-dev-env.sh
+++ b/scripts/setup-dev-env.sh
@@ -234,6 +234,8 @@ case "$(check_min_version deno "$DENO_MIN"; echo $?)" in
            deno upgrade
        fi ;;
     2) echo "→ deno not found — installing user-scoped from deno.land"
+       echo "  ℹ The Deno installer will ask which shells to configure (e.g. [ ] bash)."
+       echo "    Use ↑/↓ to navigate, Space to toggle, Enter to confirm."
        curl -fsSL https://deno.land/install.sh | sh
        export PATH="$HOME/.deno/bin:$PATH"
        if ! command -v deno >/dev/null 2>&1; then


LissaGreense:
Heads up — this PR is a workaround. The actual root cause is in `deno_install`: the interactive `[ ]` checkbox prompt doesn't surface its controls, so users unfamiliar with terminal multi-selects get stuck.

I opened a proper fix upstream at denoland/deno_install#349, which clarifies the controls inside the installer prompt itself.

We can still merge this PR now to unblock our users — the upstream fix is blocked on a `deno_install` release before it propagates to the `curl ... | sh` flow.

@Vpr99 thoughts?

ljagiello (approved):

PR #242, opened 2026-05-08
Fix silent empty global-skills exports after JetStream migration

LissaGreense:
## The bug

Settings → **Backup & restore → ☑ Include your skills library** silently produced empty exports. Toast said "Downloaded 3 workspace(s)"; copy to a clean machine; every skill-using run failed at `load_skill`.

## Where it lived

The May 3 migration (`m_20260503_110100_skills_to_jetstream`) moved user-published skills from SQLite (`~/.friday/local/skills.db`) to JetStream KV `SKILLS` + Object Store `SKILL_ARCHIVES`. Three places never got the memo:

- `packages/bundle/src/global-skills.ts:47-133` — read/wrote the legacy SQLite file.
- `apps/atlasd/routes/workspaces/index.ts:818-827, 917-925` — passed `skillsDbPath` into the broken primitive.
- `tools/agent-playground/.../settings/+page.svelte:780-786` — `globalSkillsLabel` ignored the failure header. `null` manifest values looked indistinguishable from "user didn't ask for skills."

End-to-end: a feature that *appeared* to work, lost data on every cross-machine migration, surfaced no error.

## Solution

- **Invert the bundle primitive.** `exportGlobalSkills`/`importGlobalSkills` now take a `SkillStorageAdapter`. The bundle package keeps a zero-nats dep graph (type-only import); the daemon passes `SkillStorage` (singleton facade — every other production caller does).
- **Keep `schemaVersion: 1`.** Discriminate via `source.filename` (`"skills.db"` legacy vs `"skills.jsonl"` new) — no bump ceremony.
- **Fail loud on pre-migration archives.** `LegacyArchiveError` → `kind: "legacy-archive-rejected"` with actionable UI copy. Closes the silent-failure loophole.
- **Idempotent re-import.** `publish()` always increments `version = max(latest) + 1`, so naive retries inflate versions. We mirror the migration's skip-check (`existing.version >= row.version && same namespace` → skip).
- **Drop assignments serialization.** Workspace IDs remap on import; half-shipping replays the silent-failure pattern. Follow-up, not this PR.

New layout inside the inner zip: `manifest.yml` + `skills.jsonl` + flat `archives/<skillId>__<version>.tar.gz`, with SHA-256 over the JSONL bytes and a per-archive SHA in each row.

## Test plan

- [x] 6 unit tests in `packages/bundle/` (round-trip, system filter, idempotency, manifest integrity, per-archive integrity, legacy rejection) — pass against an in-memory adapter fake.
- [x] Real-NATS integration test in `apps/atlasd/routes/workspaces/bundle-all.test.ts` — UUID-prefix isolation, asserts round-trip + idempotent re-import (no version inflation).
- [x] `deno check` on touched files + `deno task -f @atlas/agent-playground check` — clean.
- [ ] Manual smoke: publish skills via `atlas skill publish` → export → import on fresh atlas home.


packages/bundle/src/global-skills.ts
```diff
@@ -44,99 +70,250 @@ async function sha256Hex(bytes: Uint8Array): Promise<string> {
...
+  let skillsPublished = 0;
+  let skillsSkipped = 0;
+
+  for (const row of rows) {
+    // Idempotency: mirrors apps/atlasd/src/migrations/m_20260503_110100_skills_to_jetstream.ts:96-104
+    const existing = await adapter.getBySkillId(row.skillId);
+    if (
+      existing.ok &&
+      existing.data &&
+      existing.data.version >= row.version &&
```
basedfriday: Heads up: this skip check compares against the source `row.version`, but the publish below always creates `max(existing)+1` on the target. On a clean restore, a source skill at v3 imports as v1; a retry publishes v2, then v3 before it finally skips. That breaks the idempotency guarantee. I'd either export/import full version history oldest-to-newest, or normalize this to single-latest-row semantics and update the skip test around that.

packages/bundle/src/global-skills.ts
```diff
@@ -44,99 +70,250 @@ async function sha256Hex(bytes: Uint8Array): Promise<string> {
...
-    string,
-    unknown
-  >;
-  return stringify(safe, { lineWidth: 100 });
+  let skillsPublished = 0;
+  let skillsSkipped = 0;
+
+  for (const row of rows) {
+    // Idempotency: mirrors apps/atlasd/src/migrations/m_20260503_110100_skills_to_jetstream.ts:96-104
+    const existing = await adapter.getBySkillId(row.skillId);
```
basedfriday, line 345: Small reliability gap here: if `getBySkillId` returns `{ ok: false }`, we currently fall through and publish. For the JetStream adapter that can mean an existing skill/archive failed to materialize, so importing would mask the storage error and mutate state. I'd fail loud on `!existing.ok` before deciding whether to skip or publish.

packages/bundle/src/global-skills.ts
```diff
@@ -1,34 +1,60 @@
...
+//   archives/<skillId>__<version>.tar.gz      — skill archive bytes (when present)
+//
+// The manifest carries a sha256 over the assembled `skills.jsonl` bytes for
+// integrity. Each row that owns archive bytes carries a per-archive sha256.
+//
+// Latest-version semantics: the export pulls one row per skill — whatever
+// `adapter.list()` returns (latest-version-per-skillId today).
+
+import type { SkillStorageAdapter } from "@atlas/skills";
+import { SYSTEM_USER_ID } from "@atlas/skills";
```
basedfriday: Tiny layering mismatch: this is a runtime import from `@atlas/skills`, and `@atlas/skills/mod.ts` re-exports the JetStream adapter/NATS-facing bits. The PR text says bundle stays zero-NATS with only a type-level adapter dependency. If that boundary matters, I'd move `SYSTEM_USER_ID` to a tiny constants module or pass the filter in from `atlasd`.

packages/bundle/src/global-skills.test.ts
```diff
@@ -1,83 +1,460 @@
...
+): Promise<{ zipBytes: Uint8Array; manifest: GlobalSkillsManifest }> {
+  const result = await exportGlobalSkills({ adapter });
+  if (!result.bytes || !result.manifest) {
+    throw new Error("buildArchive: expected non-empty export");
+  }
+  return { zipBytes: result.bytes, manifest: result.manifest };
+}
+
+describe("importGlobalSkills", () => {
+  it("skips rows already present at >= version (idempotent re-import)", async () => {
```
basedfriday: This covers the already-at-version case, but it misses the restore path that matters most: empty target + source latest version > 1, then import the same archive again. That test would catch the current version inflation behavior.

basedfriday (commented):
Left a few inline notes. Main one is the import version/idempotency behavior.

packages/bundle/src/global-skills.ts
LissaGreense: fixed

packages/bundle/src/global-skills.ts
LissaGreense: fixed via replayVersion()

packages/bundle/src/global-skills.test.ts
LissaGreense: fixed

packages/bundle/src/global-skills.ts
LissaGreense: done ^^

LissaGreense:
QA'd this end-to-end.

**Flows tested (all passing):**

- Export happy path — UI Download button → zip on disk, v2 manifest with `schemaVersion: 2`, `skills-history.jsonl`, per-version `archives/<skillId>__<version>.tar.gz`, `X-Atlas-Global-Skills: included` header
- System skills filtered out — no `createdBy=system`, no `friday/*` namespace leaks
- Multi-version history — v1/v2/v3 rows in ascending order, per-version archives
- Disabled-state preservation — latest row reports `disabled: true` after PATCHing `/disable`
- Idempotent re-import — `skillsPublished: 0, skillsSkipped: N`, no version inflation
- v1 backwards-compat import — old `skills.jsonl` archive imports cleanly, no legacy/integrity error
- Legacy archive rejection — hand-crafted `skills.db` archive → `legacy-archive-rejected` (UI label: _"skills library archive predates the May 2026 migration — re-export from the source daemon"_)
- Silent-empty import (PRE-FIX fixture) — `globalSkills` field absent, no spurious "imported" claim
- Integrity tampering — byte-flipped JSONL → `integrity-failed` with expected/actual sha, no rows published (UI label: _"skills library failed integrity check"_)
- UI labels confirmed via the playground Settings page — "skills library imported", legacy, and integrity-failed labels all render exactly as `globalSkillsLabel(...)` defines

**One thing worth a second look** (not a blocker): per-row `disabled` reflects the *current* skill-level KV state, so older versions get stamped with the same value as the latest, rather than preserving each publish-time state. Worth confirming whether per-version semantics were intended.

@basedfriday — anything else you'd want covered here before this lands?

odk- (approved):

PR #243, opened 2026-05-09
feat(playground): HTTP/2 + TLS for local dev

basedfriday:
## Summary

- Two open playground tabs hit Chrome's 6-socket-per-origin HTTP/1.1 cap (3 long-lived SSE feeds × N tabs + Vite HMR socket), wedging every fetch including the page document. HTTP/2 multiplexes everything onto one connection — fix is local TLS via mkcert so the runtimes can negotiate h2.
- New `scripts/setup-tls.sh` installs mkcert, generates `localhost.{crt,key}` per FRIDAY_HOME, writes the env vars below to `.env` (chmod 600). Idempotent, opt-in via `bash scripts/setup-tls.sh`, skippable in `setup-dev-env.sh` with `SKIP_TLS=1`.
- `scripts/run-playground-vite.sh` exports `NODE_EXTRA_CA_CERTS` *before* `exec node` (Node 25 reads it once, at TLS context init). Bypasses `npx` so `deno task` shell doesn't route the bin script through the partial `node:http2` polyfill.
- `vite.config.ts` and `static-server.ts` resolve the cert pair, enable TLS on `server.https` / `Deno.serve`, strip HTTP/1.1 hop-by-hop headers (`transfer-encoding` et al — h2 forbids them) when proxying, and inject the daemon URL into route SSR via `define` (route-side `process.env` is wiped).
- `AtlasDaemonOptions` accepts `tlsCert`/`tlsKey`; `atlas-cli` reads paths from env. `getAtlasDaemonUrl()` upgrades an explicit `FRIDAYD_URL=http://…` to `https://` when TLS env is set, so a stale URL stops misrouting.
- `deno task atlas`/`atlas:dev` add `--env-file=$HOME/.atlas/.env` so `DENO_CERT` lands before Deno's RootCertStore initializes.

Without certs everything stays on plain HTTP — backwards-compatible.

## Env vars

| Var | Type | Read by | Set by |
|---|---|---|---|
| `FRIDAY_TLS_CERT` | path | `vite.config.ts`, `static-server.ts`, `atlas-cli daemon start`, `getAtlasDaemonUrl()` (scheme switch) | `setup-tls.sh` writes to `.env` |
| `FRIDAY_TLS_KEY`  | path | same | same |
| `DENO_CERT`       | path to mkcert root CA | Deno runtime at startup — `atlasd`, `atlas-cli`, anything `deno run` | `setup-tls.sh` writes to `.env`; loaded into Deno via `--env-file` on the `atlas` task |
| `NODE_EXTRA_CA_CERTS` | path to mkcert root CA | Node runtime at startup — vite SSR proxy, any Node tool fetching the daemon | `run-playground-vite.sh` exports from `mkcert -CAROOT` before `exec node`. Also written to `.env` for tests / other Node entrypoints |
| `FRIDAYD_URL` | URL | `vite.config.ts`, `static-server.ts`, `getAtlasDaemonUrl()` | optional override; launcher / wizard / `FAST_DEVELOPMENT` mode sets it. Scheme is auto-upgraded to `https://` when `FRIDAY_TLS_*` are set |
| `EXTERNAL_DAEMON_URL` | URL | `hooks.server.ts` → `window.__FRIDAY_CONFIG__` → browser code (oauth-popup, signal-row, etc.) | defaults to resolved daemon URL; explicit env wins |
| `EXTERNAL_TUNNEL_URL` | URL | same | same |
| `SKIP_TLS` | flag | `setup-dev-env.sh`, `setup-tls.sh` | user-set for sandboxed CI / no-admin envs |
| `FRIDAY_HOME` | path | `setup-tls.sh`, `atlas-cli` (via `getFridayHome()`) | user/launcher; default `~/.friday/local`, repo dev typically `~/.atlas` |

## .env sourcing — per-app

The PR settles on **`~/.atlas/.env` as the canonical dev `.env`**, but the two app sides reach trust material through different paths because they have different startup constraints:

**Daemon (`atlas`, `atlas:dev`):** sources `.env` twice.
1. `deno task atlas` boots Deno with `--env-file=$HOME/.atlas/.env` — vars are in `process.env` *before* Deno's RootCertStore initializes (critical for `DENO_CERT`; OTEL fires its first TLS context before `atlas-cli`'s own dotenv load could).
2. `apps/atlas-cli/src/commands/daemon/start.tsx` later calls `@std/dotenv load({ envPath: getFridayHome()+"/.env" })` so `~/.friday/local/.env` (or wherever `FRIDAY_HOME` points) is overlaid for runtime-only vars.

**Playground (vite dev, `static-server.ts`):** does **not** load `.env` directly. Trust comes from:
1. `run-playground-vite.sh` exports `NODE_EXTRA_CA_CERTS` from `mkcert -CAROOT` before `exec node` — has to happen before Node starts, can't wait for a dotenv load.
2. `vite.config.ts` and `static-server.ts` probe the filesystem at `~/.atlas/tls/localhost.{crt,key}` and `~/.friday/local/tls/localhost.{crt,key}` for the cert pair — env-var override (`FRIDAY_TLS_CERT/_KEY`) wins if set in the shell.
3. `process.env.FRIDAYD_URL` is read directly (vite's process; route-SSR reads come through a `define`-injected constant since vite's `process.env: "{}"` wipe blocks runtime env reads in transformed modules).

So daemon-side and playground-side end up with the *same* trust material (mkcert CA + per-host cert) but via different mechanisms — convergence happens at setup time (`setup-tls.sh` populates both `.env` writes and the on-disk cert files), not at runtime.

## Should everything source the same `.env`?

Trade-off: yes for ergonomics, no for startup-ordering correctness.

The dev `.env` (`~/.atlas/.env`) is already the source of truth for env-driven vars on the daemon side — `setup-tls.sh` always seeds it (regardless of which `FRIDAY_HOME` was detected) so the hardcoded `--env-file=$HOME/.atlas/.env` on the `atlas` task always resolves. The playground side could in principle load it too, but:

- `NODE_EXTRA_CA_CERTS` has to land before Node initializes its TLS root store. Node reads it once, lazily on first secure-context creation, but in practice that fires before user code can call `dotenv.load()` — so we set it from a shell wrapper before `exec node` instead of trying to plumb through `.env`.
- Vite traditionally reads `.env` from cwd (`tools/agent-playground/.env`), not from `$HOME`. Wiring the wrapper to also load a non-cwd `.env` is doable (`set -a; . "$HOME/.atlas/.env"; set +a`) but adds another layer that's already covered by filesystem probing for the cert files themselves.
- The compiled playground binary is launched by the Studio installer/launcher, which manages its own env and intentionally doesn't share the dev `.env`.

Net: `~/.atlas/.env` is canonical for the daemon; the playground reaches the same trust material through filesystem probing + a shell-wrapper export. Either side can be moved to read the same `.env` later — the cost is the startup-ordering plumbing, not the file format.

## Test plan

- [ ] `bash scripts/setup-tls.sh` (first run prompts for sudo to install mkcert root CA; subsequent runs no-op)
- [ ] `deno task dev:playground:stable` boots; daemon and vite both serve `https://`
- [ ] Browser loads `https://localhost:5200/platform/<workspace>/chat`, sends a message, agent responds end-to-end
- [ ] Pre-existing `http://localhost` setups (no `setup-tls.sh` run) still boot on plain HTTP
- [ ] OAuth providers with `http://localhost:8080/api/link/v1/callback/<provider>` registered will need `https://…` variants added in their dashboards before reconnecting under TLS

### DIFF
--- a/apps/atlas-cli/src/commands/daemon/start.tsx
+++ b/apps/atlas-cli/src/commands/daemon/start.tsx
@@ -621,12 +621,39 @@ async function startForeground(argv: StartArgs): Promise<void> {
   // the daemon's route types via @atlas/client don't drag the daemon's
   // module graph (NATS, JetStream, migrations) into their process at load.
   const { AtlasDaemon } = await import("@atlas/atlasd/daemon");
+
+  // TLS for HTTP/2 — gated on both cert and key being present and
+  // readable. Missing files fall back to plain HTTP so the daemon still
+  // boots in environments that haven't run `bash scripts/setup-tls.sh`
+  // (CI, fresh checkouts). Reading here (vs. inside the daemon) keeps
+  // the daemon module pure-data and lets us surface fs errors cleanly.
+  const tlsCertPath = process.env.FRIDAY_TLS_CERT;
+  const tlsKeyPath = process.env.FRIDAY_TLS_KEY;
+  let tlsCert: string | undefined;
+  let tlsKey: string | undefined;
+  if (tlsCertPath && tlsKeyPath) {
+    try {
+      tlsCert = await readFile(tlsCertPath, "utf8");
+      tlsKey = await readFile(tlsKeyPath, "utf8");
+    } catch (err) {
+      errorOutput(
+        `FRIDAY_TLS_CERT/_KEY set but unreadable — falling back to HTTP. ${
+          err instanceof Error ? err.message : String(err)
+        }`,
+      );
+      tlsCert = undefined;
+      tlsKey = undefined;
+    }
+  }
+  const corsOrigin = tlsCert ? "https://127.0.0.1:1420" : "http://127.0.0.1:1420";
   const daemon = new AtlasDaemon({
     port: argv.port,
     hostname: argv.hostname,
     maxConcurrentWorkspaces: argv.maxWorkspaces,
     idleTimeoutMs: (argv.idleTimeout || 300) * 1000,
-    cors: ["http://127.0.0.1:1420"],
+    cors: [corsOrigin],
+    tlsCert,
+    tlsKey,
   });
 
   // Catch unhandled promise rejections so a single async error
@@ -651,8 +678,9 @@ async function startForeground(argv: StartArgs): Promise<void> {
     Deno.addSignalListener("SIGTERM", shutdown);
   }
 
+  const scheme = tlsCert ? "https" : "http";
   infoOutput(
-    `Starting Atlas daemon on http://${argv.hostname || "127.0.0.1"}:${argv.port || 8080}...`,
+    `Starting Atlas daemon on ${scheme}://${argv.hostname || "127.0.0.1"}:${argv.port || 8080}...`,
   );
   successOutput("Atlas daemon is running. Press Ctrl+C to stop.");
 
--- a/apps/atlasd/routes/agents/register.test.ts
+++ b/apps/atlasd/routes/agents/register.test.ts
@@ -89,7 +89,7 @@ vi.mock("@atlas/logger", () => ({
 const { registerAgentRoute } = await import("./register.ts");
 const { daemonFactory } = await import("../../src/factory.ts");
 
-function makeApp(subPayload: unknown = null) {
+function makeApp(subPayload: unknown = null, natsUrl = "nats://127.0.0.1:14222") {
   const app = daemonFactory.createApp();
   app.use("*", async (c, next) => {
     const sub = subPayload !== null ? makeMockSub(subPayload) : null;
@@ -99,6 +99,7 @@ function makeApp(subPayload: unknown = null) {
           subscribe: mockSubscribe.mockReturnValue(sub),
           flush: vi.fn().mockResolvedValue(undefined),
         }),
+        getNatsUrl: () => natsUrl,
       },
       getAgentRegistry: () => ({ reload: vi.fn() }),
     } as never);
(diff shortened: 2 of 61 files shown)


basedfriday:
Passing this off to @ljagiello to finish

basedfriday:
# PR comment to leave for @ljagiello

Three follow-ups I want eyes on rather than handle in this PR — each is in a
codepath you own / wrote / touch most often. Happy to file as separate issues
if you'd rather, but flagging inline first.

---

### 1. Tauri webview → atlasd direct hits may regress under s2s TLS

Context: this PR splits the cert chain. The playground origin keeps a
mkcert-signed cert (system trust); atlasd and webhook-tunnel move to a
private-CA s2s cert that is **not** installed in any system trust store.
Browser traffic to atlasd/tunnel goes through the playground proxy at
`/api/{daemon,tunnel}/*` so the browser never sees the s2s cert.

The Tauri installer webview, however, uses the OS WebView's trust store
(WKWebView on macOS, WebView2 on Windows), which has no entry for the
private CA. If the installer UI calls `https://localhost:8080/...` directly
(rather than via the playground origin), it will fail with cert errors once
s2s TLS is on.

`apps/atlas-cli/src/commands/daemon/start.tsx:649` flips CORS origin to
`https://127.0.0.1:1420` when TLS is on, which assumes the installer **can**
reach atlasd over TLS — so I think the intent is for direct hits to work.

**Asks:**
- Does the installer webview call atlasd directly today? If yes, this is a
  regression for the installer flow and we need either (a) plumb the CA into
  the WebKit/WebView2 config, (b) terminate atlasd's TLS for the installer
  loopback specifically, or (c) route installer→daemon through a proxy.
- If the installer doesn't hit atlasd directly (everything is via the
  playground), then the `https://127.0.0.1:1420` CORS origin can probably
  be dropped or pinned to the playground origin instead.

---

### 2. Dead `EXTERNAL_DAEMON_URL` / `EXTERNAL_TUNNEL_URL` writers

These env keys used to flow through `hooks.server.ts` → `window.__FRIDAY_CONFIG__`
→ `externalDaemonUrl()` / `externalTunnelUrl()` in browser code. This PR
deletes that whole pipeline (browser hits same-origin proxy paths now), but
the producers were left in place since they're outside the playground:

- `apps/studio-installer/src-tauri/src/commands/env_file.rs` — `platform_vars`
  writes `EXTERNAL_DAEMON_URL=http://localhost:18080` and
  `EXTERNAL_TUNNEL_URL=http://localhost:19090`. No code reads them. Comment
  block at the top still describes `window.__FRIDAY_CONFIG__` which no
  longer exists.
- `tools/friday-launcher/project.go` — long doc comment listing those keys
  + `project_env_test.go` asserting on them.
- `tools/qa/live-daemon/scenarios/activity-sidebar-ui.ts` — exports them
  for completeness.

Cheap to delete in a follow-up; I left them alone here to keep this PR
scoped to the cert/proxy split. If you'd prefer I roll them in, let me know.

---

### 3. Why the legacy `localhost.{crt,key}` fallback got removed

Earlier setup-tls.sh wrote a single cert pair at `tls/localhost.{crt,key}`.
This PR splits to `tls/browser.{crt,key}` + `tls/s2s.{crt,key}`. The first
revision of this PR kept a fallback in `vite.config.ts` /
`static-server.ts` that would pick up `localhost.{crt,key}` if `browser.*`
wasn't present, framed as a "migration grace period."

I removed it. Reasoning:

- Setup-tls.sh under this PR only writes the new filenames. The fallback
  could only fire when a user had run an older revision of setup-tls.sh
  **and** had not re-run setup-dev-env.sh on this branch.
- That window is days (the time between branch land and the user's next
  setup re-run). Anyone on this branch will re-run setup at least once.
- Migration is one command: `bash scripts/setup-tls.sh`.
- The fallback is unreachable from a fresh setup, which makes it dead
  code on day 1 of merge.

Flagging in case there's a deployment path I'm not seeing where the old
filenames are written by something other than setup-tls.sh (e.g. an old
installer revision). If there is, we should put the fallback back; if not,
the simpler code wins.

ljagiello (approved):

PR #265, opened 2026-05-11
Workspace isolation + welcome flow + chat-render performance

basedfriday:
Big-ish branch covering four orthogonal-but-tangled bits of work.

## Workspaces are now properly tenant-scoped

Before: every workspace-scoped route trusted whatever `workspaceId` the request carried. Single-user local is fine, but the moment more than one user touches the daemon, you can read each other's chats / jobs / memory / artifacts / sessions / skills / configs.

Now: every workspace-scoped route checks workspace membership before doing anything. Four roles (`owner`, `admin`, `member`, `agent`); admin-only routes (config edits, member management) gate further. Sessions are opaque tokens, not the old JWT-as-identity shape. Cross-user denial tests run against a real KV bucket so the negative case stays proven.

The fail-closed default matters: an `atlasd` started without `FRIDAY_ENV` set used to silently mint sessions for everyone. Now it 401s, and the CLI / Studio installer set `FRIDAY_ENV=dev` explicitly so the local-first UX is unchanged.

The turn-cancel + stream registries are keyed by `(workspaceId, chatId)` after a second round of review caught that chat-id collisions across workspaces let a member of A abort B's in-flight turn. Admin-only surfaces (`/api/chat-storage/*`, `/api/config/env`, `/api/link/*`, `bundle-all?include=global-skills`) are now `FRIDAY_ENV=dev`-gated; outside dev they 403.

## One firehose per browser instead of N per tab

Before: each tab opened its own `EventSource` for elicitations, instance events, workspace events, plus the per-feed routes that already existed. Five tabs × three feeds was burning through the HTTP/1.1 per-origin connection budget; the daemon was duplicating the same wire to the same user N times.

Now: `GET /api/me/stream` is the single SSE the daemon emits. One SharedWorker per browser holds that connection and fans out frames over MessagePorts to whichever tab subscribed. Each subscription is a typed `AsyncIterable<T>` per channel — same consumer shape regardless of how many subscribers exist.

Live membership watch on the user's accessible-workspaces set runs alongside the SSE, so a workspace you're newly granted access to starts surfacing events on the existing connection — no reconnect.

## Welcome wizard + identity that doesn't hallucinate

Before: the assistant's \`<user_profile>\` block would happily tell you your name was \`zgjbPgBKLmeW\` because that's the userId nanoid the local adapter was substituting for an unset \`full_name\`. The auto-sync then wrote that placeholder back to USERS as \"the user's real name\", and the welcome gate considered the user done.

Now: a real \`/welcome\` page collects name + email + timezone + locale. Melt-ui combobox for IANA TZ and BCP-47 locale (typeahead + one-keystroke clear; no fiddling with character-at-a-time delete to fix a wrong autodetect). Onboarding gate via \`GET /api/me/onboarding\` — cloud requires email, local Studio can skip.

\`PATCH /api/me\` extended to take \`email\` / \`timezone\` / \`locale\` / \`profile_photo\`. \`/api/me\` schema now nullable for unset fields, so nothing downstream substitutes a placeholder. Migration \`m_20260511_120000_cleanup_userid_as_name\` reverses the persisted damage on installs that already hit the bug.

## Chat render hot path

Two passes on the chat surface. The first round was triage on heavy delegation lockups; the second round started from a Chrome trace.

### Round 1 — heavy-delegation triage

- Artifact cards each fired their own \`fetch\` for the same artifact, hitting \`ERR_INSUFFICIENT_RESOURCES\` and flashing visibly. TanStack Query factory dedupes — same id across N cards is one fetch.
- Lifted-marker regex tightened to UUID-only, so source code returned by tool calls (literal \`\${r.artifactId}\` inside a file body) stops conjuring fake artifact cards that then fail to load and retry forever.
- Message list virtualized with \`@tanstack/svelte-virtual\`, with a multi-pass scroll-to-bottom + \`ResizeObserver\` re-anchor to handle the async measure-settle so the initial scroll lands on the actual last message.
- Markdown copy-button injector coalesces its \`MutationObserver\` callbacks onto one animation frame instead of re-walking the full subtree on every streaming token.
- Auto-scroll trigger tracks the in-flight message's tool-call count only — earlier messages don't change mid-stream, so walking every segment of every message on every tick was wasted work.

### Round 2 — profiler-driven

Captured a 140s trace during streaming on a chat with ~10 visible turns and analyzed it. Top findings: \`DOMParser.parseFromString\` (DOMPurify backend) at 15,358 self-samples (~50% of non-idle CPU); TanStack \`Virtualizer.getMaxScrollOffset\` at 2,122 self-samples; 44,674 forced reflows in the span (~319/sec) because the inner scroll-anchor \`ResizeObserver\` synchronously read \`scrollHeight\` on every measured-row change; one 363ms blocking task in \`update-status.svelte.ts\` from a 1Hz \`now = Date.now()\` setInterval that drove a reactive cascade through every banner/check getter. \`/api/daemon/health\` polled every 5s and \`/api/daemon/api/workspaces/user/chat?limit=20\` every 15s, both fired during streaming.

Six changes:

- **Throttled markdown body.** New \`markdown-body.svelte\` component runs \`markdownToHTMLSafe\` synchronously on first mount, then caps subsequent renders to ~80ms while content keeps changing. Trailing flush captures the final value when streaming stops. Killed the dominant non-idle CPU.
- **Per-message display memoization.** \`displayedMessages\` in \`user-chat.svelte\` now caches the converted \`ChatMessage\` per \`AtlasUIMessage\` in a \`WeakMap\`, keyed on a coarse parts/last-part fingerprint. Stable history skips \`buildSegments\` / \`extractToolCalls\` / \`extractImages\` entirely. The tail (currently-streaming) message bypasses the cache — a follow-up fix after the original signature missed cases where the tail text segment wasn't the last \`msg.parts\` entry, which left blank bubbles during streaming.
- **rAF-coalesced scroll anchor.** Inner \`ResizeObserver\` in \`chat-message-list.svelte\` was reading \`scrollHeight\` on every observed resize. Coalesced to one anchor per animation frame.
- **Killed the 1Hz update-status setInterval.** The 10s manual-check window already nulls \`checkWindowEndsAt\` from a one-shot setTimeout, so \`checkInFlight\` / \`justChecked\` can read presence directly. The 24h dismiss window is compared against \`Date.now()\` at read time.
- **Chat-list poll cadence 15s → 60s.** The active chat updates itself through its own SSE stream; the sidebar poll only catches *other* chats arriving. Visibilitychange still refetches immediately on tab return.
- **Adaptive daemon health cadence.** 5s while unhealthy or in the disconnect-grace window, 30s once steady-connected. Hidden-tab ticks skipped.

Profiled before/after with the same 5-iteration streaming workload:

| | Before | After |
|---|---|---|
| Non-idle CPU samples | 93,883 | 65,468 (−30%) |
| \`parseFromString\` (DOMPurify) | 15,358 | 631 (−96%) |
| \`Blink.Style.UpdateTime\` | 604ms / 6,769 | 383ms / 5,045 (−37%) |
| \`Blink.Layout.UpdateTime\` | 135ms / 388 | 80ms / 210 (−41%) |
| \`Blink.Paint.UpdateTime\` | 909ms / 6,053 | 618ms / 4,572 (−32%) |
| \`Blink.ForcedStyleAndLayout\` total | 805ms | 532ms (−34%) |
| Long tasks (>50ms) | 1 (365ms) | 0 |
| \`/api/daemon/health\` polls | 21 | 3 (−86%) |
| Chat-list polls | 7 | 1 (−86%) |

## Test plan

- [x] \`deno task typecheck\` clean
- [x] \`deno task test\` green (6,311 / 6,330)
- [x] \`npx knip\` clean
- [x] Before/after Chrome trace captured, hot-path metrics in the table above
- [ ] Open \`/welcome\` cold — wizard renders, dropdowns typeahead, Save lands a USERS row + \`onboarding=complete\`
- [x] Heavy delegation chat — no \`ERR_INSUFFICIENT_RESOURCES\`, no flashing cards, scroll stays anchored
- [x] Long history chat scrolls smoothly; last message visible on open
- [x] Multiple playground tabs in one browser → one \`/api/me/stream\` SSE in the network panel
- [ ] Workspace-scoped route called by a user not in \`WORKSPACE_MEMBERS\` for that wsId → 403
- [ ] \`FRIDAY_ENV\` unset → daemon 401s; \`FRIDAY_ENV=dev\` → auto-mint local session on first request


apps/atlasd/src/workspace-authz.ts
```diff
@@ -0,0 +1,192 @@
...
+ * Returns the resolved Membership so handlers can branch on role
+ * without an extra lookup.
+ */
+export async function requireWorkspaceMember(
+  c: Context,
+  workspaceId: string,
+): Promise<WorkspaceMembership> {
+  const membership = await loadMembership(c, workspaceId);
+  if (!membership || !MEMBER_ROLES.has(membership.role)) {
+    throw new HTTPException(403, { message: "Forbidden" });
```
basedfriday, line 78: This throws the right thing locally, but in the full daemon `AtlasDaemon` has a global `app.onError()` that currently turns every thrown error into a 500. So these 401/403s will look like internal errors outside the route-level tests. Worth preserving `HTTPException` in the global handler (or returning responses here) before relying on this for authz.

apps/atlasd/src/factory.ts
```diff
@@ -92,5 +93,9 @@ export const createApp = (context: AppContext, options: CreateAppOptions = {}) =
   // Configure CORS - Hono natively handles string, string[], or "*"
   app.use("*", cors({ origin: options.corsOrigins ?? "*", exposeHeaders: ["X-Turn-Started-At"] }));
 
+  // Stamp `ctx.userId` from the opaque session cookie (or Bearer header).
+  // Local mode auto-mints; non-local 401s on missing/invalid token.
+  app.use("*", createSessionMiddleware());
```
basedfriday: Global session middleware is a pretty big behavior change: it also wraps `/health`, `/signals/*` provider webhooks, `/mcp`, and `/agents`. In non-dev those now need a Friday session cookie/bearer, which likely breaks health checks and signed platform webhooks (dev auto-minting hides it). I’d split public/signed/internal surfaces from browser-session surfaces rather than mounting this on `*`.

apps/atlasd/routes/artifacts.ts
```diff
@@ -712,6 +742,7 @@ const artifactsApp = daemonFactory
     }
```
basedfriday, line 745: Good catch for explicit `workspaceId`, but the other list branches below still bypass tenant filtering: `?chatId=` lists by chat id with no chat/workspace auth, and no filter falls through to `listAll()` with `includeData` defaulting true. For the no-filter case we need to filter by `getAccessibleWorkspaceIds(userId)` before returning anything.

apps/atlasd/routes/artifacts.ts
```diff
@@ -899,6 +939,9 @@ const artifactsApp = daemonFactory
     if (workspaceId && isInvalidChatId(workspaceId)) {
```
basedfriday, line 961: This gates uploads that declare a `workspaceId`, but the `artifactId` replace path below still calls `replaceArtifactFromFile()` without loading the existing artifact and checking its `workspaceId`. A caller can omit `workspaceId` and replace any artifact id they know. Need to authorize against the existing artifact before replacing.

apps/atlasd/routes/workspaces/chat.ts
```diff
@@ -40,24 +39,18 @@ const getChatQuerySchema = z.object({
...
 const workspaceChatRoutes = daemonFactory
   .createApp()
   .use("*", async (c, next) => {
     const workspaceId = c.req.param("workspaceId");
     if (!workspaceId) {
       return c.json({ error: "Missing workspaceId" }, 400);
     }
+    // Gate every chat route on workspace membership before the
+    // existence check — fail-closed for non-members even if the
+    // workspace is real. requireWorkspaceMember throws HTTPException;
```
basedfriday, line 51: This protects the path workspace, but a couple of later uses still need their own workspace checks: `foreground_workspace_ids` only checks existence (not membership), and stream resume/stop uses `streamRegistry` by `chatId` only. A member of workspace A can otherwise smuggle workspace B as foreground context or read/stop B’s active stream if they know the chat id.

apps/atlasd/routes/workspaces/chat-debug.ts
```diff
@@ -71,6 +72,7 @@ const workspaceChatDebugRoutes: Hono<AppVariables> = new Hono<AppVariables>().ge
     if (!chatId || !workspaceId) {
```
basedfriday, line 73: Same mismatch here: membership in the path workspace doesn’t prove this `chatId` belongs to that workspace. The debug payload later includes `streamRegistry.getStream(chatId)` events keyed only by chat id, so a user in workspace A can inspect a foreign active stream by guessing/leaking the chat id.

apps/atlasd/routes/me/stream.ts
```diff
@@ -0,0 +1,222 @@
...
+  "elicitations.>",
+  // `events.<wsId>.<...>` — workspace-scoped lifecycle, schedule.missed,
+  // health.degraded, etc.
+  "events.>",
+  // `instance.<type>.<...>` — daemon-wide cascade / queue-timeout /
+  // replace events. Not workspace-scoped; every authenticated user
+  // sees the same wire.
+  "instance.>",
+] as const;
+
```
basedfriday, line 56: `instance.>` is treated as unscoped, but the cascade payloads include workspace-ish data (`workspaceId`, `signalId`, session ids, `deepestSignal`). Since `classify()` sets `workspaceId: undefined` for these, the authz filter below never runs and every authenticated user sees all cascade metadata. Either split workspace-bearing cascade events onto workspace subjects or extract/filter/redact here.

apps/atlasd/routes/instance-events.ts
```diff
@@ -31,94 +26,17 @@ const QuerySchema = z.object({
...
-          }
-        })();
-      },
-    });
-
-    return c.body(body, 200, {
-      "Content-Type": "text/event-stream",
-      "Cache-Control": "no-cache",
-      Connection: "keep-alive",
+    const events = await listInstanceEvents(nc, {
```
basedfriday, line 41: Replay has the same leak as the live firehose: `listInstanceEvents()` returns global cascade events, including workspace ids / signal ids / session ids, with no membership filter. If these stay instance-wide, the response needs redaction or filtering before returning to regular users.

apps/atlasd/routes/skills.ts
```diff
@@ -182,7 +183,7 @@ export const skillsRoutes = daemonFactory
       if (!remoteInstallEnabled()) {
```
basedfriday, line 183: Authn alone isn’t enough for this route because the body can include `workspaceId` and later auto-assigns the installed skill to it. Need `requireWorkspaceAdmin(c, workspaceId)` before publishing/assigning, otherwise any logged-in user can inject a skill into a workspace they don’t admin.

apps/atlasd/routes/skills.ts
```diff
@@ -369,7 +370,7 @@ export const skillsRoutes = daemonFactory
       }),
```
basedfriday, line 379: Same issue on fork: if `workspaceId` is passed, this route unassigns the source skill and assigns the fork with no workspace admin check. Please gate `workspaceId` reassignment with `requireWorkspaceAdmin` before doing the publish/reassignment work.

basedfriday:
Left inline comments on the places I could anchor cleanly.

One more blocker that didn’t attach well to a changed line: `/api/workspaces/bundle-all` still calls `manager.list({ includeSystem: false })` and bundles every on-disk workspace. With the new session middleware, any authenticated user can hit it and download other tenants’ workspace configs/files. This should either filter to `getAccessibleWorkspaceIds(userId)`, require an instance-admin concept, or move off the user-facing HTTP surface.

Given the PR’s stated goal is tenant isolation, I’d also add cross-user negative tests for the inline cases: artifact list/replace, workspace chat stream/debug by mismatched `workspaceId/chatId`, skill install/fork assignment, foreground workspace IDs, instance-event replay/firehose filtering, and the real daemon `HTTPException` → status handling.

apps/atlasd/routes/workspaces/chat.ts
```diff
@@ -120,6 +118,16 @@ const workspaceChatRoutes = daemonFactory
         ? body.id
         : undefined;
     if (chatId) {
+      // If the chat already exists in storage it has to belong to the
+      // path workspace — chatTurnRegistry is keyed by chatId globally
+      // and `.replace` would otherwise let a member of workspace A
+      // abort an in-flight turn on a chatId they happen to know from
+      // workspace B. A first-time chatId (not yet persisted) is fine;
+      // it'll be created as a new chat under this workspace.
+      const existing = await ChatStorage.getChat(chatId);
```
basedfriday: This still doesn't prove the chat belongs to this workspace. `ChatStorage.getChat(chatId)` with no `workspaceId` resolves to the legacy `_global` bucket (`resolveWorkspaceId(undefined)`), not a global search across workspace chats. So a chat that exists in workspace B comes back `null` here, then `chatTurnRegistry.replace(chatId)` still runs and can abort B's turn because the registry key is still just `chatId`.

For this POST path, `getChat(chatId, workspaceId)` alone isn't enough either — "not found in A" is indistinguishable from "exists in B". The safe fix is to key the turn/stream registries by `(workspaceId, chatId)` or do a real ownership-aware/global uniqueness lookup before replacing. The current mocks hide this because they make `getChat(chatId)` behave like a global lookup.

basedfriday (commented):
Non-inline because these files/lines aren't in the diff:

- `apps/atlasd/routes/chat-storage.ts` is still mounted at `/api/chat-storage` and any authenticated user can list/read/replace conversations by raw `streamId`. If this endpoint is not strictly internal, it bypasses the workspace-member model completely; it needs internal auth or streamId → workspace authorization.
- Need an explicit call on global admin surfaces. `/api/config/env` reads/writes daemon `.env`, `/api/link/*` proxies to Link with the daemon `FRIDAY_KEY`, and `bundle-all?include=global-skills` can still export global skill state. If multi-user/cloud daemon is in scope, these need an instance-admin gate or need to be local/dev-only.
- Pagination nit: `GET /api/artifacts` does `listAll(limit)` and then filters inaccessible rows, so a user can get short/empty pages even when accessible artifacts exist later. Not a tenant leak, but UI pagination will be weird.

apps/atlasd/routes/workspaces/chat.ts
```diff
@@ -138,14 +146,26 @@ const workspaceChatRoutes = daemonFactory
...
     const chatId = c.req.param("chatId");
+    const workspaceId = c.req.param("workspaceId");
+    if (!workspaceId) {
+      return c.json({ error: "Missing workspaceId" }, 400);
+    }
+
+    // Stream + turn registries are keyed by chatId globally — without
+    // verifying the chat belongs to this workspace a member of A could
+    // stop B's in-flight turn by guessing/leaking the chatId.
+    const chat = await ChatStorage.getChat(chatId);
```
basedfriday: Same lookup bug in the stream routes: no-arg `getChat` checks `_global`, so normal workspace chats are not found and `DELETE /stream` returns 404 for legit cancellations. Use `getChat(chatId, workspaceId)` here; `GET /:chatId/stream` below has the same issue.

apps/atlasd/routes/workspaces/chat-debug.ts
```diff
@@ -71,6 +73,18 @@ const workspaceChatDebugRoutes: Hono<AppVariables> = new Hono<AppVariables>().ge
...
     }
+    await requireWorkspaceMember(c, workspaceId);
+    // Membership on the path workspace isn't enough — the active
+    // stream registry is keyed by `chatId` only, so without this
+    // chat-belongs-to-workspace check a member of A could read B's
+    // in-flight tool events by guessing the chat id. The JetStream
+    // stream name + KV key below are workspace-scoped and naturally
+    // won't materialise for the wrong workspace, but `streamRegistry`
+    // would.
+    const chatCheck = await ChatStorage.getChat(chatId);
```
basedfriday: Same issue as the workspace stream routes: no-arg `getChat` only checks `_global`, so this debug endpoint 404s for normal workspace chats. Since the path already has `workspaceId`, this should be `getChat(chatId, workspaceId)` before consulting the chatId-only `streamRegistry`.

apps/atlasd/routes/artifacts.ts
```diff
@@ -726,6 +760,14 @@ const artifactsApp = daemonFactory
     }
 
     if (query.chatId) {
+      // Resolve the chat's owning workspace and gate on membership —
+      // a `chatId` is just an opaque string the client guesses at,
+      // not a proof of access. ChatStorage indexes by chatId so the
+      // workspaceId argument is optional here.
+      const chat = await ChatStorage.getChat(query.chatId);
```
basedfriday: This comment is backwards. `ChatStorage` does not index workspace chats by `chatId` globally; omitting `workspaceId` reads the `_global` namespace. So `GET /api/artifacts?chatId=<workspace chat>` now returns 404 even for the owning member. Need either require/derive a workspaceId and call `getChat(chatId, workspaceId)`, add a real workspace-scoped chat lookup, or list artifacts by chat and authorize/filter each returned artifact's workspace.

apps/atlasd/src/workspace-authz.ts
basedfriday: Addressed. Global handler at `apps/atlasd/src/atlas-daemon.ts:1456-1462` checks `instanceof HTTPException` and returns `err.getResponse()`, so the 401/403 status flows through to the client instead of collapsing to a 500. Generic errors still log + return 500 as before.

apps/atlasd/src/factory.ts
basedfriday: Addressed. Session middleware is scoped to `/api/*` (`apps/atlasd/src/factory.ts:105`), not `*`. `/health`, `/signals/*`, `/mcp`, `/agents` stay public/signed/internal — the block at 96-104 names each surface and why it stays out.

apps/atlasd/routes/artifacts.ts
basedfriday: Addressed (eec246c). The `?chatId=` branch now lists by chat id and post-filters against `getAccessibleWorkspaceIds(userId)` rather than doing a chat-by-id lookup (which would have fallen into the `_global` namespace). The no-filter `listAll()` branch is gated by `requireDevEnv(c)` — pagination over cross-tenant rows can't be made correct from `listAll(limit)` without a per-user index, so we shut the operator door instead.

apps/atlasd/routes/artifacts.ts
basedfriday: Addressed (eec246c). The replace branch at `apps/atlasd/routes/artifacts.ts:1025-1042` loads the existing artifact via `ArtifactStorage.get({ id: artifactId })` and then calls `requireWorkspaceMember(c, existing.data.workspaceId)` before any `replaceArtifactFromFile`. The `workspaceId` the caller posts is no longer the source of truth on the replace path; the existing artifact's workspace is.

apps/atlasd/routes/workspaces/chat.ts
basedfriday: Addressed (eec246c). Two fixes here: (1) `foreground_workspace_ids` are each `requireWorkspaceMember(c, fgId)`-checked in `apps/atlasd/routes/workspaces/chat.ts:99-112`; (2) the turn + stream registries are re-keyed by `(workspaceId, chatId)` (see `chat-turn-registry.ts` + `stream-registry.ts`), so a foreign workspace's controller can't be aborted from a colliding chat id even if the chat lookup were ambiguous.

apps/atlasd/routes/workspaces/chat-debug.ts
basedfriday: Addressed (eec246c). `StreamRegistry` is now keyed by `(workspaceId, chatId)`; the debug route resolves the existence check via `getChat(chatId, workspaceId)` and pulls the stream from `streamRegistry.getStream(workspaceId, chatId)`, so cross-workspace inspection by guessed chat id can't happen.

apps/atlasd/routes/me/stream.ts
basedfriday: Addressed (eec246c). `apps/atlasd/routes/me/stream.ts:188-192` runs every `instance` cascade frame through `filterCascadeForUser(payload, accessible)` — the same filter the replay endpoint uses — before writing it to the wire. Workspace-bearing cascade payloads are dropped or redacted based on membership, not exposed wholesale.

apps/atlasd/routes/instance-events.ts
basedfriday: Addressed (eec246c). `apps/atlasd/routes/instance-events.ts:41-54` resolves the caller's `accessible` workspaces and runs every replayed event through `filterCascadeForUser` before returning. Live firehose and replay are now in lockstep — the same rule applies to both.

apps/atlasd/routes/skills.ts
basedfriday: Addressed (eec246c). `apps/atlasd/routes/skills.ts:196-197` — if `workspaceId` is in the body, `requireWorkspaceAdmin(c, workspaceId)` runs before any publish/assign work.

apps/atlasd/routes/skills.ts
basedfriday: Addressed (eec246c). `apps/atlasd/routes/skills.ts:391-392` — same `requireWorkspaceAdmin(c, workspaceId)` gate before the unassign/reassign sequence on fork.

apps/atlasd/routes/workspaces/chat.ts
basedfriday: Addressed (eec246c). The registries are re-keyed by `(workspaceId, chatId)` (`apps/atlasd/src/chat-turn-registry.ts` + `stream-registry.ts`); POST `/` passes the path `workspaceId` into `chatTurnRegistry.replace(workspaceId, chatId)`, so a workspace B chat with a colliding id can't be aborted from a workspace A request. The `getChat(chatId)` → `_global` resolution bug is also fixed — all chat lookups in the chat / chat-debug routes now pass `workspaceId`.

apps/atlasd/routes/workspaces/chat.ts
basedfriday: Addressed (eec246c). DELETE / GET `:chatId/stream` now call `getChat(chatId, workspaceId)` with the path workspaceId, and the registry access uses `(workspaceId, chatId)` keys — no more `_global` 404s, no more cross-workspace stop/resume.

apps/atlasd/routes/workspaces/chat-debug.ts
basedfriday: Addressed (eec246c). The chat-debug endpoint uses `getChat(chatId, workspaceId)` for the existence check and `streamRegistry.getStream(workspaceId, chatId)` for the live stream payload — both correctly scope to the path workspace.

apps/atlasd/routes/artifacts.ts
basedfriday: Addressed (eec246c). The `?chatId=` branch now lists artifacts by chat id and post-filters by `getAccessibleWorkspaceIds(userId)` — no chat-by-id global lookup, no `_global` fallback. The earlier comment that confused two paths is gone.

basedfriday:
Picking up the two non-inline threads:

**`/api/workspaces/bundle-all`** — addressed (eec246c). The route now filters every workspace through `getAccessibleWorkspaceIds(userId)` before bundling, and `?include=global-skills` runs `requireDevEnv(c)` first — outside dev it 403s. Cross-tenant negative coverage lives in `apps/atlasd/routes/workspaces/index.test.ts`.

**Round-2 non-inline list** — all gated:

- `apps/atlasd/routes/chat-storage.ts` — `.use("*", devOnlyMiddleware())` on the routes chain; outside `FRIDAY_ENV=dev` every method 403s.
- `apps/atlasd/routes/config.ts` — `configRoutes.use("/env", devOnlyMiddleware())`.
- `apps/atlasd/routes/link.ts` — `linkRoutes.use("/*", devOnlyMiddleware())`.
- `/workspaces/bundle-all?include=global-skills` — `requireDevEnv(c)` before the global-skill branch (see above).
- Artifacts no-filter pagination — the `listAll()` branch is now `requireDevEnv`-gated since pagination over cross-tenant rows can't be made correct from `listAll(limit)` without a per-user index. The two filtered branches (`?chatId=` and explicit `workspaceId`) are membership-checked.

Cross-tenant negative coverage for the surfaces you called out:

- `routes/workspaces/chat.test.ts` — POST `/`, DELETE / GET `:chatId/stream` against a chat in workspace B with the caller in workspace A.
- `routes/artifacts.test.ts:1713+` — "workspace membership gates on the upload path" suite, including the replace-by-id 403.
- `routes/skills.test.ts` — `requireWorkspaceAdmin` gates on install + fork when `workspaceId` is in the body.
- `routes/me/stream.test.ts` — "drops events from a workspace the user is not a member of" + the instance-cascade filter case.

The chat-debug route doesn't have a dedicated test file yet — it's a thin wrapper over `getChat(chatId, workspaceId)` + `streamRegistry.getStream(workspaceId, chatId)`, both of which are tested at their own seams. Happy to add a dedicated test if you'd like a belt-and-braces 403 case there.

Inline replies posted on the rest.

PR #267, opened 2026-05-12
Surface transient OAuth refresh errors to chat without killing refresh_token

ljagiello:
## Summary

Three goals, minimum surface area:

- **Don't remove `refresh_token` on transient errors.** A 60s threshold + classifier in Link distinguishes provably-dead tokens (`invalid_grant`) from transient failures (5xx / 429 / network / timeout). Storage is never written on transient.
- **Surface transient errors to chat.** Link's `refresh_unavailable` status maps to `LinkCredentialUnavailableError`, which `createMCPTools` converts to a `credential_temporarily_unavailable` disconnect entry; chat renders the right copy on the assistant message.
- **Evals + Chrome QA cover it.** New promptfoo scenario (`tools/qa/live-daemon/scenarios/oauth-refresh-transient.ts`, 18 checks) pins classifier outcomes, disconnect-shape mapping, and the chip-render contract. Verified end-to-end via Chrome: real Google credential survives a network blackhole transient with SHA-identical file content before/after.

## Test plan

- [ ] `deno run -A tools/qa/live-daemon/scenarios/oauth-refresh-transient.ts` — 18/18 pass
- [ ] `cd tools/qa/live-daemon/promptfoo && npx promptfoo eval -c oauth-refresh-transient-config.yaml` — 18/18 pass
- [ ] `apps/link/tests/oauth.test.ts` "Delegated OAuth refresh classifier integration" — covers refresh_unavailable / silent_fallback / token_dead matrix end-to-end through Link's HTTP surface


apps/link/src/routes/credentials.ts
```diff
@@ -433,6 +438,66 @@ async function resolveCredentialWithRefresh(
...
 
+    // Delegated mode: use the classifier so we can distinguish a provably
+    // revoked refresh_token (token_dead → refresh_failed) from a transient
+    // upstream issue (transient → refresh_unavailable / silent fallback).
+    if (
+      provider?.type === "oauth" &&
+      provider.oauthConfig.mode === "delegated" &&
+      oauthCredResult.data.secret.refresh_token
+    ) {
+      const outcome = await refreshDelegatedTokenClassified(
```
basedfriday: This new delegated path bypasses `OAuthService.refreshCredential()`, so it also bypasses the existing per-credential `refreshLocks` mutex. Concurrent internal reads for the same stale delegated credential can now fan out multiple refresh calls and race to return inconsistent `refreshed`/`refresh_unavailable` statuses. Can we move the classified delegated refresh behind the OAuthService lock (or add an equivalent per-credential lock around this branch)?

apps/link/src/oauth/delegated.ts
```diff
@@ -115,40 +118,249 @@ const DelegatedRefreshResponseSchema = z.object({
...
+  if (response.ok) {
+    const rawBody = await response.text();
+    let json: unknown;
+    try {
+      json = JSON.parse(rawBody);
+    } catch {
+      logger.warn("oauth_refresh_platform_bug", {
+        reason: "non_json_success_body",
+        status: response.status,
+        body: rawBody.slice(0, 500),
```
basedfriday: This logs raw token-endpoint response bytes, and the nearby `detail` strings return raw body slices too. OAuth refresh endpoints are a secret boundary; if the delegated function ever returns/echoes an `access_token` in a malformed body, this leaks into logs or internal API responses. Please redact token-like fields or avoid logging/returning body content here and in the similar 4xx platform-bug branches below.

apps/link/src/oauth/delegated.ts
```diff
@@ -115,40 +118,249 @@ const DelegatedRefreshResponseSchema = z.object({
...
+    const detail = err instanceof Error ? err.message : String(err);
+    if (name === "AbortError" || name === "TimeoutError") {
+      return { kind: "transient", reason: "timeout", detail };
+    }
+    return { kind: "transient", reason: "network", detail };
+  }
 
-  if (!response.ok) {
+  if (response.ok) {
+    const rawBody = await response.text();
```
basedfriday: If `fetch` resolves headers but the body stream errors/truncates, this `response.text()` throw escapes the classifier. That means callers see a generic 500 instead of the intended `refresh_unavailable`/silent-fallback transient handling. Please catch body-read failures on the success path and classify them as transient, with a small regression test.

basedfriday (commented):
Posted the main findings from my review. The direction looks sound, but I think the delegated refresh lock bypass should be addressed before merge.

apps/link/src/routes/credentials.ts
ljagiello: Addressed in 17e441332. Added a module-scoped `delegatedRefreshInFlight` Map<credentialId, Promise<CredentialResponse>> mirroring `OAuthService.refreshLocks` — same in-flight Promise dedup pattern, scoped to the delegated branch since it returns the richer `RefreshOutcome` shape that the OAuthService path doesn't surface. Concurrent reads now collapse to a single upstream fetch and a single storage write per credential id.

apps/link/src/oauth/delegated.ts
ljagiello: Addressed in 17e441332, then hardened further in 77676fd75. Initial fix added a `redactTokens` regex helper, but that's defense-in-depth via pattern-matching — easy to miss the next branch / next token shape (alternate key names, XML wrapping, etc.). Refactored to remove raw body bytes from logs and `detail` strings entirely:

- `RefreshOutcome.detail` (exposed via `/internal/v1/credentials/:id`) now emits only structural strings — status code, statusText, byte length. Never any body content. Safe by construction.
- `logger.warn("oauth_refresh_platform_bug", …)` emits `body_length` + `content_type` instead of `body: rawBody.slice(…)`. Same diagnostic value for operators ("100 KB HTML error page" vs "empty body" vs "JSON shape regression") without a leak channel.
- The `redactTokens` helper was deleted.

Tests assert the *absence* of body content in `detail`, keyed on multi-format leak strings (JSON-quoted, bare `=`, HTML-wrapped, prose) — they don't rely on a regex pattern, so a future refactor that lets raw bytes back in will fail them. 28 tests passing.

apps/link/src/oauth/delegated.ts
ljagiello: Addressed in 17e441332. Wrapped the success-path `response.text()` in a try/catch and classified body-read failures as `transient/network` (consistent with "network failed mid-transfer"), so callers reach the silent-fallback (≥60s remaining) or `refresh_unavailable` (<60s) branch instead of an unhandled rejection bubbling up as a 500.

Regression test added: `kind=transient network when 2xx body read throws (truncated stream)` — uses a `ReadableStream` that calls `controller.error()` after headers, asserts `kind: "transient", reason: "network"`.

PR #268, opened 2026-05-12
Chat tools: one-call artifact saves + lighter claude-code reach + MCP name discipline

basedfriday:
## What users will notice

Three rough edges go away in the workspace-chat assistant:

- **"Read this PDF" no longer spawns the heavyweight code agent.** The chat would previously reach for `agent_claude-code` for any task that touched a file — reading a PDF, fetching a URL, writing marketing prose. It now routes those to the right direct tool (`parse_artifact`, `web_fetch`, `run_code`) and reserves the code agent for things it actually exists for (multi-file refactors, codebase debugging).

- **"Save this and show it to me" is one tool call instead of two.** A new `save_artifact` tool takes inline content and registers it as an artifact in one step. The chat no longer has to do `write_file` → `create_artifact(path)` — which was slow and added a visible seam between "I have the content" and "I have an artifact ID".

- **No more invented MCP tool names.** When a workspace wires up an MCP server (Gmail, Slack, etc.), the chat used to guess at tool names that looked plausible and call them blind. Those calls were rejected at the SDK boundary and surfaced as confusing failures. The chat now discovers tool names via `list_mcp_tools` / `describe_mcp_tool` before calling anything.

## Design notes

`save_artifact` rejects binary-MIME filenames at the schema boundary (`.png`, `.pdf`, `.zip`, etc.) — the model can't accidentally stuff base64 into a string field and produce a corrupted artifact. For binary outputs the path stays: `run_code` writes to scratch, then `create_artifact({path})` registers it.

Both artifact tools now return `{id, type, summary}` instead of `{artifactId}`. That shape matches the agent-SDK's artifact-ref harvester, so refs from these direct tools flow into the same pipeline as the MCP `create_artifact` — previously the workspace-chat direct tools were invisible to harvesting.

The `agent_claude-code` carve-out lives in the workspace-chat prompt only. The bundled agent's own description is kept terse so the steering doesn't bleed into surfaces (like an agent catalog UI) that have no business seeing workspace-chat-specific routing rules.

## Eval coverage

New eval suite under `tools/qa/prompt-tuning/` — 15 scenarios driven by Anthropic against the real prompt + real bundled-agent description + a mocked tool catalog. Runs via promptfoo so each scenario's full system prompt, user message, and captured tool sequence are visible in the UI. Provider warns when serving a cached report older than 10 minutes.

Scenarios cover claude-code overreach negatives (PDF, prose, URL, hash, DOCX), the positive multi-file-refactor case, save_artifact preference across markdown / JSON / CSV, the binary counter-case, MCP discipline across one and two servers, a path-traversal security boundary, and a no-library-hint variant of the binary case.

Unit tests added for `save_artifact` cover MIME inference, the new return shape, binary-filename rejection, and the path-traversal guard.

## Tradeoffs

- `save_artifact` doesn't mirror its content to the scratch dir. If a follow-up `run_code` needs to read what was just saved, pass the content through or write to scratch separately. The mirror was speculative and added a failure mode (disk-full / permission errors blocking the upload).
- Mock tool descriptions in the eval are copied from source. If production descriptions drift, the eval still passes against the old text. Acceptable for now; worth a shared-constants refactor if these evals become CI-gating.

## Test plan

- [x] `deno task typecheck` — clean
- [x] `deno task test packages/system/agents/workspace-chat/tools/artifact-tools.test.ts` — 11/11 pass (5 existing `create_artifact` cases + 6 new mixed across both tools)
- [x] `deno task test packages/agent-sdk/src/vercel-helpers` — 39/39 pass
- [x] `npx knip` — clean
- [x] Eval suite: 15/15 pass via `deno run --allow-all tools/qa/prompt-tuning/scenarios/tool-choice.ts` and via `npx promptfoo eval -c tools/qa/prompt-tuning/promptfoo/tool-choice-config.yaml`
- [ ] Smoke-test in a live workspace-chat session against the new tool to confirm the chat actually picks `save_artifact` for "save this for me" prompts


ljagiello:
## Code review — 7 issues worth addressing before merge

Verdict: **Needs Work** (no critical findings; all fixes are surgical). Full review at `docs/reviews/2026-05-11-pr-268-chat-tools.md`.

---

### 1. `save_artifact.content` has no size cap

`packages/system/agents/workspace-chat/tools/artifact-tools.ts:200` declares `content: z.string().min(1)` with no `.max(...)`. Sibling `write_file` caps at 2 MB (`MAX_WRITE_BYTES`); the receiving artifact route enforces size only on the multipart `/upload` path, not the JSON `POST /`. An LLM emitting a runaway transcript streams straight into the JSON body and persists.

**Fix:** one-line schema change — `z.string().min(1).max(2_000_000)` (or whatever parity is right).

### 2. Binary-MIME guard / docstring mismatch — `.zip` etc. are not actually rejected

`isTextSafeFilename` (`artifact-tools.ts:159-167`) returns `true` (allow) when `inferMimeFromFilename` returns `undefined`. `EXTENSION_TO_MIME` has no entry for `.zip`, `.exe`, `.bin`, `.tar`, `.gz`, `.wasm`, `.parquet`, `.sqlite`, `.dmg`. The tool description and SKILL.md both call out `.zip` as explicitly rejected. The test suite doesn't catch this.

Threat is bounded (LLM mistake corrupts an artifact; not a security issue), but the prompt and description lie.

**Fix:** pick one — drop the `.zip` claim from the description and pin current behavior with a test, **or** invert the default to "unknown extension ⇒ reject unless inferred MIME is text-safe".

### 3. `extractClaudeCodeDescription` parser is brittle and fails silent

`tools/qa/prompt-tuning/scenarios/tool-choice.ts:82-89` parses TS source with a hardcoded literal marker (`'description:\n      \"'`). On miss, it `console.warn`s and returns a 6-word stub — the eval still reports "pass", but `claude-code-overreach-*` scenarios are now measuring whether the model rejects a stub, not the production description. Whole point of the eval is fidelity to the live prompt+description.

**Fix:** import the TS module directly and read `.description`, **or** `throw` on miss. Two-line change.

### 4. Harvester fixture doesn't cover the new `save_artifact` branch

`packages/agent-sdk/src/vercel-helpers/tool-usage.ts` widened `result.toolName === \"create_artifact\"` to `ARTIFACT_PRODUCING_TOOLS.has(result.toolName)`. The test file `tool-usage.test.ts` has 6 tests on `extractArtifactRefsFromToolResults`, all hardcoded `toolName: \"create_artifact\"`. No fixture exercises `save_artifact`.

**Fix:** one additional test row asserting a `toolName: \"save_artifact\"` result is harvested. The behavior change is one Set entry; no test pins it.

### 5. Eval is non-deterministic — set `temperature: 0`

`tool-choice.ts:768` calls `streamText` with no `temperature`. Anthropic's default is non-zero, so each scenario is a single non-deterministic sample. A green run can flake red on the next pass — meaningful when this eval gates prompt edits.

**Fix:** set `temperature: 0`. One line.

### 6. `prompt.txt` still references the old return field name

`packages/system/agents/workspace-chat/prompt.txt:247` says "immediately call \`display_artifact\` with the returned \`artifactId\`". The tool now returns \`id\`, not \`artifactId\`. The paragraph just below correctly says \`id\`.

**Fix:** one-word edit.

### 7. MCP \`no-fabricated-tool-names\` eval bakes in a debatable policy

\`tool-choice.ts:633-672\`. The user explicitly names \`gmail_get_thread\`; the eval insists on \`list_mcp_tools\` / \`describe_mcp_tool\` before any \`delegate\`. The prompt rule it enforces ("never invoke an MCP tool by a name you haven't seen returned from discovery") is reasonable as schema hygiene, but is debatable when the user provides the exact name. The eval and the prompt rule will now drift together — fine if intentional.

**Fix (decide, then sync):** if the policy is correct, keep both. If it's overreach, reword the prompt rule and the scenario in lockstep so they don't silently disagree.

basedfriday:
Thanks @ljagiello — six of seven landed in `9c2a242`, one stays as-is with reasoning.

1. **`save_artifact.content` size cap** ✅ — `.max(2 * 1024 * 1024)` to mirror `write_file`'s ceiling, plus prose update on the description.
2. **Binary-MIME guard / docstring mismatch** ✅ — flipped the default: unknown extensions are rejected now, not allowed. Description tightened to say "must resolve to a known text MIME". New test loops `.zip` / `.exe` / `.weirdext` and pins the rejection.
3. **`extractClaudeCodeDescription` brittle + silent** ✅ — regex is now whitespace-tolerant and `throw`s on miss. The stub fallback is gone; running the eval against a placeholder description was worse than failing loudly.
4. **Harvester fixture for `save_artifact`** ✅ — new test row in `tool-usage.test.ts` exercises the `toolName: "save_artifact"` branch end-to-end.
5. **`temperature: 0`** ✅ (with caveat) — set conditionally. `claude-opus-4-7` is reasoning-class and the SDK warns + ignores the field, so we skip it on `opus-4-X` patterns to avoid log noise. For Sonnet / Haiku (and any future temp-supporting model someone swaps in via `EVAL_CHAT_MODEL`) outcomes pin deterministically across runs.
6. **`artifactId` → `id` in `prompt.txt`** ✅ — one-word fix on the artifacts paragraph. (Line 119 talks about job-tool returns which still carry `artifactIds[]` — that's a different envelope, left untouched.)
7. **MCP `no-fabricated-tool-names` policy** — keeping the discovery-before-invoke rule. The reviewer's framing was "user provided the exact name", but the original failure mode is *parameter* hallucination, not just name hallucination. Even with a correct tool name the model doesn't know the inputSchema — that's the dominant break per the comments on `describe_mcp_tool` itself. Discovery before invoke is the rule the prompt enforces and the eval tests; both stay synchronized.

Local: 52/52 unit tests, 15/15 eval scenarios, typecheck + biome clean. Watching CI.

ljagiello (approved):

PR #271, opened 2026-05-12
Delegate: faster responses when sub-agent answers are document-shaped

basedfriday:
## Summary

Trims wasted decode time on chats where a delegate task produces substantial prose — long emails to draft, multi-row tables, document summaries, batched fetch results. The user-visible stream is identical; what changes is the supervisor's behavior after the sub-agent finishes — it stops re-rendering content the sub-agent has already streamed.

Small short delegations are unaffected and the supervisor still wraps them with its own context. The savings only kick in when the answer is big enough to be worth shipping as an artifact.

## Design — four moving parts

**Pre-return answer lift.** Large successful answers route through the shared artifact-lift walker before they leave the delegate tool's `execute`. The supervisor's tool-result view sees an artifact marker; the next step references it via `display_artifact` instead of regenerating prose. Judge-style direct-execute callers opt out via a `liftAnswer: false` flag since they parse the answer text themselves.

**Text-on-success child contract.** The sub-agent's system prompt now directs it to emit final answers as plain text content. The `finish` tool stays for failures only (`{ ok: false, reason }`). Removes the wasted double-decode where the model wrote the same long answer once as text and once as a `finish` argument.

**Supervisor stop predicate.** A new `delegateAnsweredUser()` stop condition fires when `delegate` returns the lifted-marker shape. The chat agent's `stopWhen` array picks it up and the supervisor loop terminates cleanly — no synthetic re-wrap step. Short answers (under the lift threshold) leave the predicate dormant so the supervisor keeps its wrap-up capability.

**Per-step model swap.** Sub-agent's `streamText` now has a `prepareStep` callback. Step 0 (initial plan + first tool calls) stays on the conversational model. Subsequent tool-processing steps drop to the classifier-slot model (Haiku by default) for mechanical work. Falls through if the classifier slot isn't configured on a given deployment.

## Scope

Items 1, 2, 4 are in the shared `createDelegateTool` so workspace-chat, FSM, judge, and workspace-runtime all inherit them automatically. Item 4's `stopWhen` is workspace-chat-only — FSM uses `complete` tools with typed schemas (no prose regeneration to skip) and judge calls `execute` directly (no supervisor loop).

The session-supervisor agent is unrelated; it doesn't use delegate.

## Test plan

- [x] All 7 promptfoo suites pass with no regressions (62/62 cases).
- [x] New unit coverage for lift wiring, judge opt-out, failure-path skip, and prepareStep model swap.
- [x] `stop-conditions.test.ts` covers the new predicate including a regression guard against non-delegate tool results triggering it.
- [x] `deno task typecheck` clean.
- [x] knip clean.

## Follow-up

A second PR moves delegate's stream plumbing to the SDK's preliminary-tool-results AsyncIterable shape, replacing the bespoke `data-delegate-chunk` envelope plus tee/forward proxy. That refactor lets us adopt `toModelOutput` and `Output.text()` natively once `execute` becomes a generator — the canonical SDK 6 sub-agent shape.


packages/system/agents/workspace-chat/workspace-chat.agent.ts
```diff
@@ -1298,6 +1302,7 @@ export const workspaceChatAgent = createAgent<string, WorkspaceChatResult>({
                 stepCountIs(40),
                 connectServiceSucceeded(),
                 connectCommunicatorSucceeded(),
+                delegateAnsweredUser(),
```
basedfriday: I think this stop is too broad. In the AI SDK loop, if this `stopWhen` fires after the delegate tool result, the supervisor never gets another generation to decide whether the delegate output was final vs just intermediate context.

Example failure shape: user asks “collect these 50 emails and draft a reply.” The delegate may return a large lifted collection, this predicate fires, and the parent never does the drafting/wrapping/display step. Also, the streamed delegate prose renders as the delegate card’s nested `response`, not as normal top-level assistant text.

Can we either let the supervisor continue with a compact model projection, or require an explicit “delegate answered the user” signal before stopping?

packages/core/src/delegate/index.ts
```diff
@@ -583,9 +614,29 @@ export function createDelegateTool(deps: DelegateDeps, toolSetThunk: () => Atlas
...
+      // Successful answers above the shared text threshold are routed
+      // through the artifact-lift walker before they leave `execute`.
+      // The supervisor LLM's tool-result view sees an artifact marker
+      // rather than the inline prose — prevents re-decoding content the
+      // child has already streamed to the user via `data-delegate-chunk`
+      // text-deltas. Direct-execute callers (judge agent) that parse the
+      // raw answer themselves pass `liftAnswer: false`.
+      const projectAnswer = async (raw: string): Promise<string> => {
+        if (!liftAnswer) return raw;
+        const lifted = await liftValueForModel(raw, {
```
basedfriday: This looks like the wrong layer for the projection. AI SDK 6 already has `tool({ toModelOutput })`, which maps the tool output only for the next model call. Here we mutate the actual `DelegateResult.answer`, so every non-model consumer gets the artifact marker unless it remembers to opt out (`judge` does, future direct callers may not).

Could keep `execute` returning the raw answer and put the lifted marker in `toModelOutput` instead. That matches the SDK subagent pattern and avoids the opt-out footgun.

packages/core/src/delegate/index.ts
```diff
@@ -424,6 +449,12 @@ export function createDelegateTool(deps: DelegateDeps, toolSetThunk: () => Atlas
           maxOutputTokens,
           abortSignal: childAbortSignal,
           providerOptions: getDefaultProviderOpts("anthropic"),
+          prepareStep: ({ stepNumber }) => {
+            if (stepNumber > 0 && fastModel) {
+              return { model: fastModel };
```
basedfriday: Careful: step 1 is often the final synthesis step, not just mechanical tool processing. In the common loop, step 0 calls tools, then step 1 reads the tool results and emits the final text — and with this PR that text is directly user-visible.

If we want the speedup, I’d gate the classifier swap to tool-selection-only steps, or keep the final free-text synthesis on the conversational model. Otherwise long emails/summaries/tables get silently downgraded right on the path this PR optimizes.

packages/system/agents/workspace-chat/stop-conditions.ts
```diff
@@ -1,6 +1,41 @@
...
+ * delegate's stream be the assistant's answer.
+ *
+ * Short delegate answers (under the lift threshold) do NOT trigger this
+ * — they leave room for the supervisor to wrap them in context, ask a
+ * follow-up, or add a short prose preamble.
+ */
+const DelegateLiftedAnswerSchema = z.object({
+  output: z.object({
+    ok: z.literal(true),
+    answer: z.string().startsWith("[attachment lifted to artifact "),
```
basedfriday: This prefix is a pretty brittle control signal. The scrubber intentionally returns strings that already start with `[attachment lifted to artifact `, so a delegate can output marker-shaped text and trip the stop condition without any lift/upload having happened.

Can we make this structured instead — e.g. `answerLifted: true` / `answerArtifactId` stamped by the lift path — and have the stop predicate check that field rather than human-readable marker text?

basedfriday:
Non-inline note: the targeted suite is still red for me:

```bash
deno task test packages/system/agents/workspace-chat/stop-conditions.test.ts packages/core/src/delegate/index.test.ts packages/system/agents/workspace-chat/workspace-chat.handler.test.ts
```

Failure is the existing workspace-chat handler assertion expecting 3 stop conditions, now getting 4:

```
packages/system/agents/workspace-chat/workspace-chat.handler.test.ts:908
expected ... to have a length of 3 but got 4
```

Small fix, but worth updating before merge since it also gives us a good place to add an integration assertion for the new stop path / rendered answer shape.

packages/system/agents/workspace-chat/workspace-chat.agent.ts
basedfriday: Agreed — dropped the predicate (commit f4fb7e8). The "collect 50 emails and draft a reply" failure shape was the right framing. The supervisor now always gets its next step after delegate returns, so it can wrap, follow up, draft, or whatever else the user asked for. The marker projection on the tool result still prevents prose regeneration on the simple "fetch and show" case.

packages/core/src/delegate/index.ts
basedfriday: Tried this — landed on a compromise (commit f4fb7e8).

The clean SDK-native shape (execute returns raw, toModelOutput projects marker) has a persist-layer footgun in this codebase: `scrubAssistantMessage` walks the persisted tool-result's `output.answer` field; with raw text sitting there, it would lift again and create a second artifact record pointing at the same blob (the object store dedupes by sha256, but the artifact metadata records duplicate). Avoiding that needs either a smarter scrubber that recognizes a sibling `answerArtifactId` field, or keeping the marker substitution in `execute` so the persist walker's `REF_MARKER_PREFIX` short-circuit handles it for free.

Went with the latter for this PR — single-lift end-to-end, no scrubber knowing about delegate-specific shapes. The opt-out footgun is real but the flag is now explicit + documented; default is the safer "lift" path; only direct-execute callers that parse the answer themselves (the judge) need to set it. Also added `answerArtifactId` as a structured signal alongside the marker so future consumers branch on the field instead of the marker text.

Open to revisiting on the follow-up PR that does the AsyncIterable refactor — that one will rework the persist path anyway and `toModelOutput` lands more naturally there.

packages/core/src/delegate/index.ts
basedfriday: Agreed — reverted the prepareStep entirely (commit f4fb7e8). The synthesis-step concern is real and we don't have an eval that would have caught a quality drop on long-prose answers. Keeping the child on the conversational model throughout until there's a gated swap with eval support behind it.

packages/system/agents/workspace-chat/stop-conditions.ts
basedfriday: Good catch — moot now that the predicate is dropped (commit f4fb7e8), but the structural fix is in either way: `DelegateResult.answerArtifactId` is stamped by `execute` when the lift fires, so future consumers can branch on the structured field rather than the marker text.

PR #288, opened 2026-05-12
Workspace bundle: include user agents referenced by workspace.yml

odk-:
## Summary
- `exportBundle` only walked `<workspaceDir>/agents/`, so workspaces that reference globally-installed user agents (e.g. `type: user` entries that resolve to `<atlasHome>/agents/<id>@<version>/`) bundled the reference but **no definition** — the lockfile recorded `primitives.agents: {}` and the bundle was unusable on import for those agents.
- Add an optional `externalAgents: { name, sourceDir }[]` to `ExportOptions` so callers can pin agents from outside the workspace dir.
- `buildWorkspaceBundleBytes` now resolves every `type: user` entry via `UserAdapter` against `<atlasHome>/agents/`. Workspace-local agents still win on name collision; unresolvable refs log a warning and the export proceeds without that agent.

Repro: download a workspace whose `workspace.yml` references a user agent that lives under `~/.atlas/agents/<id>@<version>/`. Before the fix, the zip contained `workspace.yml` + `workspace.lock` only. After the fix, it contains `agents/<id>/...` and the lockfile pins the agent hash.

Note: this is the export side only. On import the agent files land in `<workspaceDir>/agents/<id>/` (workspace-local). Re-installing them into `<atlasHome>/agents/<id>@<version>/` so the global resolver finds them is a follow-up.

## Test plan
- [x] `deno task test packages/bundle/` — 40 tests pass (incl. new round-trip + collision tests for `externalAgents`)
- [x] `deno task test apps/atlasd/routes/workspaces/` — 275 tests pass (incl. new e2e test seeding `<homeDir>/agents/spritesheet-normalizer@1.0.0/` and asserting the zip contains `agents/spritesheet-normalizer/...`)
- [x] `deno check` clean on changed files

### DIFF
--- a/apps/atlasd/routes/workspaces/bundle-all.test.ts
+++ b/apps/atlasd/routes/workspaces/bundle-all.test.ts
@@ -189,12 +189,17 @@ describe("bundle-all endpoints (end-to-end)", () => {
     await seedWorkspaceDir(wsA, "alpha");
     await seedWorkspaceDir(wsB, "beta");
     mockFetchLinkCredential.mockReset();
+    // `/bundle-all?include=global-skills` is gated behind `requireDevEnv`;
+    // the runner sets DEV_MODE/LINK_DEV_MODE but not FRIDAY_ENV, so opt in
+    // here for the duration of the test.
+    vi.stubEnv("FRIDAY_ENV", "dev");
   });
 
   afterEach(async () => {
     await rm(wsA, { recursive: true, force: true });
     await rm(wsB, { recursive: true, force: true });
     await rm(homeDir, { recursive: true, force: true });
+    vi.unstubAllEnvs();
   });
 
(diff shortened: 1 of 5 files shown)


LissaGreense:
Could we throw on unresolvable `type: user` agents in `resolveExternalUserAgents` instead of just warning? Right now the bundle ships 200 OK but `workspace.yml` references an agent that isn't in the zip — failure surfaces much later, on the target machine.

LissaGreense (approved):

odk-:
Switched to fail-fast — `resolveExternalUserAgents` now collects all missing IDs and throws a single error listing them, so the operator sees the full set of unresolvable refs in one go instead of one-at-a-time. Added a route test that asserts a 500 + error message when a referenced user agent isn't installed.

PR #296, opened 2026-05-13
Home-dir isolation: dev and prod each get their own NATS broker

basedfriday:
## Summary

- Running `deno task atlas` (dev) alongside Friday Studio.app (prod) on
  the same machine no longer cross-contaminates state. Each install
  picks its own NATS port from a Friday-reserved range (14222–14231)
  and writes the chosen URL to `<home>/nats/url` so siblings can find
  it. Daemon, CLI, and launcher all discover the broker through that
  file instead of hard-coding `:4222`.
- Workspaces registered under a different Friday home (the residue of
  historical cross-contamination) are now filtered out at runtime
  rather than mounted, watched, or operated on. Lookups by id / name /
  path mask cross-home entries; `deleteWorkspace`, `renameWorkspace`,
  `updateWorkspacePersistence`, and the signal-restart path all
  refuse to touch them. Entries stay in KV; the `generateUniqueId`
  walk still sees them so we never re-mint a colliding id.
- The default home for a bare `friday` binary is now `~/.friday/local`
  (matches the installer/launcher convention). `deno task atlas` pins
  `FRIDAY_HOME=\$HOME/.atlas` in `deno.json` so dev behavior is
  unchanged. The cwd-`.atlas` heuristic at the bottom of
  `getFridayHome()` is gone — it never fired in current layouts.
- `ENVIRONMENTS.md` lands as the operator-facing config reference.

## Env-var changes

| Var | Status | Notes |
|---|---|---|
| `FRIDAY_HOME` | semantics changed (var unchanged) | Default when unset flips from `~/.atlas` to `~/.friday/local`. The dev `atlas` / `atlas:dev` / `start` deno tasks now explicitly pin `FRIDAY_HOME=\$HOME/.atlas` so dev keeps writing to `~/.atlas`. Bare `friday` invocations now match the installer/launcher convention. |
| `FRIDAY_NATS_URL` | role tightened (var unchanged) | Still wins as the cloud-shared-broker override. Now short-circuits the URL-file rendezvous — the daemon connects directly and never writes the URL file. Operator-set in shell / k8s / launchd, never in per-home `.env`. |
| `FRIDAY_NATS_PORT` | new (reserved for future use) | Documented in `ENVIRONMENTS.md` as the cloud-with-per-tenant-broker hook for orchestrators (k8s, multi-tenant launcher) that want to pin a specific port. NOT consumed by code in this PR — auto-allocation via `pickPort()` covers today's needs. Reserved so adding the env later doesn't surprise anyone. |
| `FRIDAY_LOCAL_ONLY`, `FRIDAY_DUCKDB_PATH`, `FRIDAY_SQLITE3_PATH`, `AGENT_SOURCE_DIR` | confirmed dead (no code change) | Audited and confirmed to have zero readers in `apps/` / `tools/` / `packages/`. Documented for future cleanup; not removed in this PR. |

The new `pickPort()` reserved range (`14222..14231`) and `<home>/nats/url` discovery file are internal contracts, not env vars. Operators see them through logs (port number, URL file path) but never set them directly.

## Test plan

- [ ] Clean install: install Studio, confirm boot — broker lands on
      14222, daemon connects, workspaces auto-import.
- [ ] Dev alongside prod: with Studio running, \`deno task atlas
      daemon start\` in a checkout. Verify the dev daemon boots on a
      different port (likely 14223) and does not see prod's
      workspaces.
- [ ] Cross-home residue: on an existing install with cross-home
      registry entries, verify the daemon logs \`level=warn\` for each
      skipped entry and that the entries don't appear in the
      workspace list.
- [ ] \`atlas migrate\` while the daemon is up: discovers the broker
      via \`<home>/nats/url\` and runs as a connected client (no
      second broker spawn).

## Migration

None — data layout is unchanged. JetStream stores stay where they
are; the new daemon picks a different port but reads the same
on-disk store. Rollback to a pre-PR binary is safe: it re-binds
\`:4222\` against the same store dir and resumes.

## Notes

- Evals weren't run for this PR — the change is infrastructure
  (NATS port + URL-file rendezvous + workspace-registry filter),
  not agent behavior. Unit + integration coverage is extensive
  (cross-home filter, user-ID stability under N=20 races, two-
  daemon port-iteration in Go, daemon-startup with isolated home
  dirs, connectOrSpawn URL-file + spawnFallback contract).

### DIFF
--- a/apps/atlasd/src/nats-manager.ts
+++ b/apps/atlasd/src/nats-manager.ts
@@ -43,7 +43,11 @@ import type { NatsConnection } from "nats";
 export class NatsManager {
   private spawned: SpawnedNats | null = null;
   private nc: NatsConnection | null = null;
-  /** Set when we own the URL file (i.e. we spawned the broker). */
+  /**
+   * Set after a successful start. The daemon is the canonical writer
+   * of `<home>/nats/url` — `stop()` deletes the file regardless of
+   * which path put us on the broker (env URL, cached URL, own spawn).
+   */
   private ownsUrlFile = false;
 
   async start(): Promise<NatsConnection> {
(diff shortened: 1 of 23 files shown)


tools/friday-launcher/main.go
```diff
@@ -371,6 +371,18 @@ func onReady() {
 		"source", jetstreamStoreSource,
 	)
 
+	// Pick the NATS port once at launcher startup. Iterates Friday's
+	// reserved range (14222..14231) and falls back to OS-assigned if
+	// all 10 slots are taken. The chosen port lands in the
+	// `natsServerPort` package-level so both `natsServerArgs` and
+	// `commonServiceEnv` see the same value when they build the
+	// nats-server spec and the child env respectively.
+	natsServerPort = pickNATSPort()
```
ljagiello, line 380: **Critical: launcher never writes `<home>/nats/url`, so the rendezvous design is dead for Studio users.**

The launcher picks a port here and sets `FRIDAY_NATS_URL` for its supervised children, but nobody writes the URL file. `apps/atlasd/src/nats-manager.ts:55-61` returns early when `FRIDAY_NATS_URL` is set, skipping the URL-file write too. `grep -rn writeBrokerUrl tools/friday-launcher/` returns zero hits.

Failure mode: a user runs `friday migrate` (or any CLI) from a separate terminal with no inherited env. It falls through to `pickPort()` and tries to spawn a second nats-server against the same `store_dir`, which file-lock-conflicts. The error message `"Failed to spawn ephemeral nats-server"` gives no hint a daemon is already running. The `connectOrSpawn` doc-comment in `packages/jetstream/src/connect.ts` claims "The launcher / daemon writes this file when it spawns the broker" — only the daemon does, and only when *it* (not the launcher) spawned the broker.

**Recommendation:** simplest fix is to have the daemon write the URL file even on the external-URL-from-env path (it knows the URL it just connected to). Single writer responsibility, smaller diff than teaching the launcher about the file.

ljagiello (commented):
Posting the two **Critical** findings from review as inline comments. Full review (verdict: Needs Work, 2 critical + 8 important + 5 test gaps) at `docs/reviews/2026-05-12-home-dir-isolation.md`. Architectural direction is sound — these are localised fixes, not a redesign.

apps/atlasd/src/nats-manager.ts
```diff
@@ -51,52 +60,51 @@ export class NatsManager {
...
+    // but throwaway for the daemon).
+    this.spawned = await spawnNatsServer({ port, workDir: join(home, "nats"), storeDir, logger });
+
+    const url = this.spawned.url;
+    await writeBrokerUrlFile(home, url);
+    this.ownsUrlFile = true;
+    logger.info("Wrote broker URL file", { url, urlFile: brokerUrlFilePath(home) });
+
+    if (process.env.FRIDAY_NATS_MONITOR === "1") {
+      logger.info(`NATS monitoring enabled at http://localhost:${DEFAULT_NATS_MONITOR_PORT}`);
```
ljagiello: I think at this point, we know exactly protocol / host so you should not hardcode here http://localhost

PR #301, opened 2026-05-13
Improve chat rendering performance during large tool renders

Vpr99:
## Summary

Three small fixes to cut streaming jank during large tool-result renders. The
email-list reproducer (50-row markdown table streamed alongside 15 tool cards)
showed ~9% active CPU, a 39ms blocking task, and Shiki tokenisation dominating
the profile.

- **Lazy-render tool-call drawers.** Svelte `<details>` doesn't lazy-mount;
  closed drawers were still running Shiki on every chunk for content the user
  never opened. `bind:open` + `{#if open || isExport}` gates it. Error drawers
  still default to `errorOpen = true` to preserve the prior `<details open>`
  visibility.
- **Memoise `formatRawOutput`.** The `{@html formatRawOutput(data)}` callsite
  re-evaluates on every parent render; the new
  `tools/agent-playground/src/lib/components/chat/format-raw-output.ts`
  module caches by `output` reference (WeakMap) and by content (bounded
  64-entry FIFO). `tool-call-card.svelte` and `delegate-tool-card.svelte`
  both import it — the latter previously kept its own divergent inline copy.
- **Bump markdown throttle 80ms → 120ms.** Cuts `marked.parse` + DOMPurify
  rate without crossing the perceptual threshold.

Verified via Chrome perfetto traces over the same prompt:

|          | baseline | after  |
|----------|----------|--------|
| active   | 9.1%     | 2.6%   (-71%) |
| max task | 39ms     | 11.5ms (-71%, under one 60Hz frame) |
| >16ms    | 1        | 0      |
| Shiki    | top-3 hot | not in top-15 |

## Excluded from this PR

The first iteration of this branch also included a scroll/follow-bottom rework
of `chat-message-list.svelte` and a tree-stabilising refactor of
`extractToolCalls` / `buildSegments`. Both were reverted before this review
pass because (a) the ref-stability layer protected microseconds while the win
was already covered by lazy drawers + the `formatRawOutput` WeakMap, and (b)
the scroll rework introduced three real regressions flagged in the first
review pass (sentinel race, gesture-only detach, streaming-text drift). The
scroll work will land as its own PR — design plan posted as a PR comment.

## Test plan

- [x] All chat tests pass (190/190 in `tools/agent-playground/src/lib/components/chat/`)
- [x] New `format-raw-output.test.ts` (10 cases): null/undefined, WeakMap
  object hit, object cache miss on different refs, string hit, FIFO eviction
  past 64, circular-ref → escape fallback, invalid-JSON string, valid-JSON
  string
- [x] `copy-buttons-export.test.ts` tightened: replaced toothless
  `toContain("input")` with structural regexes pinning text to `<summary>`;
  added error-drawer cases for both export and live modes
- [x] svelte-check: 0 errors
- [x] Manual: drawers still expand, tool cards still update, markdown still
  streams, rendered output unchanged

### DIFF
--- a/packages/core/src/chat/export/render.ts
+++ b/packages/core/src/chat/export/render.ts
@@ -126,6 +126,99 @@ function readNestedParentId(data: unknown): string | undefined {
 }
 
 /**
+ * Optional inputs that let `extractToolCalls` preserve reference identity
+ * for tool-call entries (and their `Segment` parents) across re-runs.
+ *
+ * The live chat UI calls `extractToolCalls` on every streaming chunk for
+ * the tail assistant message. Without ref-stable output, every downstream
+ * `$derived` in `ToolBurst` / `ToolCallCard` re-runs each tick, and any
+ * cached work keyed on the call object (e.g. the Shiki JSON highlighter
+ * memo in `format-raw-output.ts`) misses immediately. See profiling notes
+ * in `/Users/ericskram/.../memory/` if reintroducing this contract.
+ *
+ * Pass the flattened map from the *previous* `extractToolCalls` result as
+ * `prevByToolCallId`. Entries whose shallow shape matches (state, input ref,
+ * output ref, error text, duration, reasoning, delegate text, progress
+ * length, child refs) are returned by reference instead of as fresh objects.
+ *
+ * Pure helper — never mutates the previous map.
+ */
+export interface ExtractToolCallsOptions {
+  prevByToolCallId?: ReadonlyMap<string, ToolCallDisplay>;
+}
+
+/**
+ * Cheap structural equality for `ToolCallDisplay`. The deep fields
+ * (`input`, `output`, `children`) are compared by reference because the
+ * upstream `msg.parts[]` reuses object references for the same payload
+ * across re-runs (AI SDK v6 contract — see `lifted-markers.ts` for the
+ * same reliance). `progress` is array-by-reference; we compare lengths
+ * as a fast falsifier in case the array was rebuilt in place.
+ */
+function toolCallShallowEqual(a: ToolCallDisplay, b: ToolCallDisplay): boolean {
+  if (a === b) return true;
+  if (
+    a.toolCallId !== b.toolCallId ||
+    a.toolName !== b.toolName ||
+    a.state !== b.state ||
+    a.input !== b.input ||
+    a.output !== b.output ||
+    a.errorText !== b.errorText ||
+    a.durationMs !== b.durationMs ||
+    a.reasoning !== b.reasoning ||
+    a.delegateText !== b.delegateText ||
+    a.workspaceId !== b.workspaceId ||
+    a.sessionId !== b.sessionId ||
+    a.actionId !== b.actionId ||
+    a.jobName !== b.jobName
+  ) {
+    return false;
+  }
+  if (a.progress !== b.progress) {
+    const al = a.progress?.length ?? 0;
+    const bl = b.progress?.length ?? 0;
+    if (al !== bl) return false;
+  }
+  // Children are stabilised bottom-up before the parent equality check, so
+  // a reference match here implies the subtree is unchanged.
+  return a.children === b.children;
+}
+
+/**
+ * Walk `entries` bottom-up. For each entry, recursively stabilise children,
+ * then compare against `prev`. If everything matches, reuse the previous
+ * reference; otherwise return a fresh node carrying the stabilised children.
+ */
+function stabilizeTree(
+  entries: ToolCallDisplay[],
+  prevByToolCallId: ReadonlyMap<string, ToolCallDisplay>,
+): ToolCallDisplay[] {
+  let mutated = false;
+  const out: ToolCallDisplay[] = new Array(entries.length);
+  for (let i = 0; i < entries.length; i++) {
+    const entry = entries[i]!;
+    let nextChildren = entry.children;
+    if (nextChildren && nextChildren.length > 0) {
+      const stabilisedChildren = stabilizeTree(nextChildren, prevByToolCallId);
+      if (stabilisedChildren !== nextChildren) {
+        nextChildren = stabilisedChildren;
+      }
+    }
+    const candidate: ToolCallDisplay =
+      nextChildren === entry.children ? entry : { ...entry, children: nextChildren };
+    const prev = prevByToolCallId.get(entry.toolCallId);
+    if (prev && toolCallShallowEqual(prev, candidate)) {
+      out[i] = prev;
+      if (prev !== entry) mutated = true;
+    } else {
+      out[i] = candidate;
+      if (candidate !== entry) mutated = true;
+    }
+  }
+  return mutated ? out : entries;
+}
+
+/**
  * Extract tool-call parts from an {@link AtlasUIMessage} in stream order,
  * reconstruct any nested delegate children, and reconcile their final
  * states using the `delegate-end` blanket rule.
@@ -145,8 +238,15 @@ function readNestedParentId(data: unknown): string | undefined {
  * for a future reflection layer, not the UI tree. `durationMap` is scoped
  * by `delegateToolCallId` so multiple delegates with identically-named
  * children do not collide.
+ *
+ * When `options.prevByToolCallId` is supplied, returned nodes (and their
+ * subtrees) preserve reference identity wherever the structural shape is
+ * unchanged. See {@link ExtractToolCallsOptions}.
  */
-export function extractToolCalls(msg: AtlasUIMessage): ToolCallDisplay[] {
+export function extractToolCalls(
+  msg: AtlasUIMessage,
+  options?: ExtractToolCallsOptions,
+): ToolCallDisplay[] {
   if (!Array.isArray(msg.parts)) return [];
 
   // Pass 1: flat top-level tool calls.
@@ -425,6 +525,9 @@ export function extractToolCalls(msg: AtlasUIMessage): ToolCallDisplay[] {
     }
   }
 
+  if (options?.prevByToolCallId && options.prevByToolCallId.size > 0) {
+    return stabilizeTree(trees, options.prevByToolCallId);
+  }
   return trees;
 }
 
@@ -451,16 +554,42 @@ export function flattenToolCalls(calls: ToolCallDisplay[]): Map<string, ToolCall
 }
 
 /**
+ * Inputs that let `buildSegments` preserve reference identity for the
+ * returned segments (and any nested `ToolCallDisplay` entries) across
+ * re-runs of the same message. See {@link ExtractToolCallsOptions}.
+ *
+ * `prevSegments` lets `tool-burst` segments reuse their previous reference
+ * when the contained calls are identical by reference (post tool-call
+ * stabilisation). Text segments reuse their previous reference when the
+ * coalesced content matches the prior content character-for-character.
+ */
+export interface BuildSegmentsOptions extends ExtractToolCallsOptions {
+  prevSegments?: readonly Segment[];
+}
+
+function callsArrayEqual(a: ToolCallDisplay[], b: ToolCallDisplay[]): boolean {
+  if (a === b) return true;
+  if (a.length !== b.length) return false;
+  for (let i = 0; i < a.length; i++) {
+    if (a[i] !== b[i]) return false;
+  }
+  return true;
+}
+
+/**
  * Build chronological {@link Segment}s from an {@link AtlasUIMessage}'s
  * `parts[]` array. Consecutive text parts coalesce into a single `text`
  * segment; consecutive tool-call parts (and any reasoning that arrived
  * between them) group into a `tool-burst` segment. This preserves the
  * true stream order so the UI can render prose and tool activity exactly
  * where they happened.
+ *
+ * Pass `options` to opt into reference-stable output. See
+ * {@link BuildSegmentsOptions} for the contract.
  */
-export function buildSegments(msg: AtlasUIMessage): Segment[] {
+export function buildSegments(msg: AtlasUIMessage, options?: BuildSegmentsOptions): Segment[] {
   if (!Array.isArray(msg.parts)) return [];
-  const allToolCalls = extractToolCalls(msg);
+  const allToolCalls = extractToolCalls(msg, options);
   const toolMap = flattenToolCalls(allToolCalls);
   // Walk parts as opaque records — the AI SDK discriminated union narrows
   // each adjacent type check into dead code once we touch part.type.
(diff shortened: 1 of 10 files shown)


basedfriday (commented):
## Code review — Verdict: **Rethink**

The four genuine performance changes (memoised `formatRawOutput`, lazy-rendered tool-call drawers, deferred `extractLiftedArtifactIds`, throttle bump) are sound and low-risk. But the largest part of the diff is a scroll/follow-bottom rework of `chat-message-list.svelte` that is unrelated to rendering performance, unmentioned in the description, and introduces two regressions in core chat behaviour. Inline comments below.

### PR-level concerns

**Description / diff are desynchronized.** The description claims a "reference-stable extraction" fix wired through `user-chat.svelte` plus "6 new ref-stability tests" in `render.test.ts` — none of which appear in the diff (the only `render.ts` change is one blank line). Conversely, the scroll/follow-bottom rework — the largest change — is not mentioned at all. Please realign the description with what actually ships.

**The scroll rework belongs in its own PR.** It is a behaviour change, not a perf change; bundling it under a "rendering performance" title hides it from reviewers and couples two unrelated risk profiles. It also carries the PR's two critical regressions (see inline) and has zero tests. The four real perf changes are mergeable on their own once `format-raw-output.ts` has coverage.

**Critical: streaming text no longer re-anchors to the bottom.** The deleted ResizeObserver `$effect` (old lines 335-367) was the only thing re-anchoring on `.virtual-inner` growth. The `scrollToBottom` effect only depends on `messages.length` and `tailToolCallCount` — neither changes during pure text-token streaming inside an existing assistant message (that path mutates `MarkdownBody`'s `html`). Removing the observer **and** adding `overflow-anchor: none` (which disables the browser's native compensation) means a streaming assistant message visibly drifts up off the bottom of the viewport — the exact bug the observer's docstring says it existed to fix. Keep the observer (or a `followBottom`-gated equivalent); if the goal was cutting its reflow cost, tune it rather than delete it.

packages/core/src/chat/export/render.ts
```diff
@@ -541,6 +541,7 @@ export function buildSegments(msg: AtlasUIMessage): Segment[] {
 
   flushText();
   flushBurst();
+
```
basedfriday: This blank line is the only change to this file — pure diff noise, and it drags a `packages/core` file into a playground-scoped PR. Please revert so the PR stays within `tools/agent-playground`.

tools/agent-playground/src/lib/components/chat/chat-message-list.svelte
```diff
@@ -273,23 +273,110 @@
...
+  // `scrollHeight - clientHeight`. Storing the pre-clamp value as the
+  // sentinel makes the comparison miss by `clientHeight` (hundreds of
+  // pixels), so every programmatic snap fell through to re-attach
+  // `followBottom = true` and stomped every gesture detach.
+  let ignoreNextScrollTo: number | null = null;
+  function snapToBottom() {
+    if (!containerEl) return;
+    containerEl.scrollTop = containerEl.scrollHeight;
+    ignoreNextScrollTo = containerEl.scrollTop;
+  }
```
basedfriday: **`ignoreNextScrollTo` can get stuck and silently swallow a real scroll.** `snapToBottom` records the post-write `scrollTop` as the sentinel — but if the write didn't change `scrollTop` (already at the bottom, the common steady state while following), no `scroll` event fires, the sentinel is never cleared, and it persists until the *next* scroll event. That next event may be a genuine user scroll, now ignored if it lands within 2px. `scrollToBottom` calls `snapToBottom` 3× across rAF boundaries, widening the window. Clear `ignoreNextScrollTo` unconditionally after the last rAF in `scrollToBottom`, or expire it by frame-count/timestamp rather than relying on the next event arriving.

tools/agent-playground/src/lib/components/chat/chat-message-list.svelte
```diff
@@ -273,23 +273,110 @@
...
+      return;
+    }
+    ignoreNextScrollTo = null;
     const distanceFromBottom =
       containerEl.scrollHeight - containerEl.scrollTop - containerEl.clientHeight;
-    followBottom = distanceFromBottom < STICK_THRESHOLD_PX;
+    // Re-attach only. Detachment is owned by the gesture listeners below
+    // so a programmatic scroll can never accidentally clear follow.
+    if (distanceFromBottom < ATTACH_THRESHOLD_PX) followBottom = true;
   }
```
basedfriday, line 288: **Critical: detach is now gesture-only, so scrollbar-drag re-snaps the user to the bottom.** `handleScroll` only ever sets `followBottom = true` now; detachment relies on `wheel`/`touchmove`/`keydown` listeners. Dragging the scrollbar thumb upward fires *only* a `scroll` event — none of the listened gestures — so `followBottom` is never cleared and the user is yanked back down on the next streaming tick. Same gap for find-in-page scroll-into-view and middle-click autoscroll. The old `distanceFromBottom < STICK_THRESHOLD_PX` logic covered all of these uniformly because it derived detach from *position*, not input device. Recommend reinstating position-based detach here (set `followBottom = false` past a larger threshold) — the `ignoreNextScrollTo` sentinel already suppresses programmatic-scroll false positives, so the two coexist, and the gesture `$effect` below then becomes largely redundant.

tools/agent-playground/src/lib/components/chat/chat-message-list.svelte
```diff
@@ -970,6 +1022,10 @@
     flex: 1;
     flex-direction: column;
     gap: var(--size-4);
+    /* Disable the browser's built-in scroll anchoring — our follow
+       logic owns scroll position, and the UA's anchor can fight us
+       when a tool card mounts above the viewport mid-stream. */
+    overflow-anchor: none;
```
basedfriday: `overflow-anchor: none` combined with deleting the ResizeObserver re-anchor effect (old lines 335-367) is what produces the critical streaming-drift regression noted in the PR-level comment. The UA's scroll anchoring was the last fallback keeping a growing message pinned; with both gone, nothing re-anchors during pure text streaming. Don't pair the observer removal with this.

tools/agent-playground/src/lib/components/chat/format-raw-output.ts
```diff
@@ -0,0 +1,78 @@
...
+
+function escapeHTML(value: string): string {
+  return value.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
+}
+
+function highlight(jsonStr: string): string {
+  return jsonHighlighter.codeToHtml(jsonStr, { lang: "json", theme: "atlas-json" });
+}
+
+export function formatRawOutput(output: unknown): string {
```
basedfriday, line 42: **Critical: this new module ships with zero tests, and a divergent copy still exists.** The branching here is exactly the kind that ships bugs — `WeakMap` object cache, bounded string cache with eviction, `JSON.parse` branch, `escapeHTML` fallback on `JSON.stringify` throw, null/undefined handling. Please add a table test (object cache hit, string cache hit, eviction past 64, unstringifiable object → escaped fallback, invalid-JSON string → escaped, null/undefined). Note that module-level `objectCache`/`stringCache` will leak across cases in the same test file — use fresh object references per case.

Separately: `delegate-tool-card.svelte` keeps its **own inline `formatRawOutput`** that this PR doesn't touch, so the caching win is partial (delegate cards stay un-cached, and they're a streaming hot path too) and two implementations will now drift. Wire it to this shared module or document why it's excluded.

tools/agent-playground/src/lib/components/chat/format-raw-output.ts
```diff
@@ -0,0 +1,78 @@
...
+    try {
+      const parsed: unknown = JSON.parse(output);
+      result = highlight(JSON.stringify(parsed, null, 2));
+    } catch {
+      result = escapeHTML(output);
+    }
+    if (stringCache.size >= STRING_CACHE_MAX) {
+      const firstKey = stringCache.keys().next().value;
+      if (firstKey !== undefined) stringCache.delete(firstKey);
+    }
```
basedfriday, line 76: This eviction is FIFO, not LRU — it drops the oldest *inserted* key, not the least-recently-*used*. The docstring above and the PR description both call it an "LRU"; a hot, frequently-reused string can still be evicted while one-off entries survive. For 64 tool-output entries it rarely matters in practice, but the claim should match the code: either re-insert on hit for true LRU, or rename to "bounded FIFO cache."

tools/agent-playground/src/lib/components/chat/copy-buttons-export.test.ts
```diff
@@ -84,13 +84,24 @@ describe("tool-call-card copy buttons — export mode", () => {
...
     const { body } = render(ToolCallCardExportParent, {
       props: { call: callWithOutput },
     });
 
-    expect(body).toContain("json-copy-wrapper");
-    expect(body).toContain("json-copy-btn");
-    expect(body).toContain("Forecast");
+    // Drawer chrome (summary labels) renders eagerly...
+    expect(body).toContain("input");
+    expect(body).toContain("output");
```
basedfriday: These positive assertions are too loose to protect what the comment claims. `toContain("input")` / `toContain("output")` also match class names, aria attributes, and the tool name itself — if the drawer chrome disappeared entirely, these could still pass. Only the `not.toContain` lines below have teeth. Tighten to structural assertions, e.g. `toContain(">input</summary>")` or count `<details>` occurrences.

Also worth noting: this is an SSR-string-only check. `inputOpen` is `$state(false)` and never toggled under `svelte/server`, so a bug where a drawer never renders content *even when the user opens it* would not be caught. If the lazy-open behaviour matters, it needs a client-render test that toggles `<details>` open and asserts the payload then appears.

tools/agent-playground/src/lib/components/chat/tool-call-card.svelte
```diff
@@ -307,42 +314,51 @@
...
+                  class="json-copy-btn"
+                  aria-label="Copy error"
+                  onclick={(e: MouseEvent) => copyToClipboard(c.errorText, e.currentTarget as HTMLButtonElement)}
+                >
+                  Copy
+                </button>
+                <pre class="json-render error-text">{c.errorText}</pre>
+              </div>
+            {/if}
           {/if}
```
basedfriday, line 362: The error drawer moved from `<details open>` to `bind:open={errorOpen}` + a nested `{#if errorOpen || isExport}` gate — a behaviour-sensitive restructure with no test covering it before or after. Add a `callWithError` case asserting the error text and copy affordance render eagerly (error defaults open).

Vpr99:
## Scroll / follow-bottom rework — follow-up PR plan

Posting this so the scroll-fix design isn't lost; the local copy lives at `tools/agent-playground/docs/scroll-follow-bottom-rework-plan.md` on the branch (gitignored per repo convention for the playground `docs/` folder).

### Original problem

The auto-scroll fought the user mid-stream. Position-based detach lost to programmatic snaps: each `containerEl.scrollTop = scrollHeight` fired a synthetic `scroll` event that re-evaluated distance from bottom and reset `followBottom = true`, stomping any detach the user had just made.

### What the reverted patch (53676d3) got right

- **Sentinel-with-read-back filter.** Recording the post-clamp `scrollTop` (not the pre-clamp value) is the correct fix for the "ignore my own synthetic scroll event" problem. Setting `scrollTop = scrollHeight` is silently clamped to `scrollHeight - clientHeight`; storing the pre-clamp value misses by `clientHeight` and the filter becomes a no-op.
- **`scrollToBottom` re-checks `followBottom` between rAF passes** so a mid-pass detach short-circuits the remaining passes.
- **Text-selection drag suppression** so scrolling past the bottom while extending a selection doesn't strand the user.

Keep all three in the follow-up.

### What the reverted patch got wrong

#### 1. `ignoreNextScrollTo` sentinel can stick

If `snapToBottom()` is called when already at the bottom, the browser fires no `scroll` event, the sentinel never clears, and it persists until the next event — which may be a genuine user scroll within 2px, now silently ignored. `scrollToBottom` calls `snapToBottom` 3× across rAF boundaries, widening the window.

**Fix.** Convert to a generation counter that `handleScroll` compares against and clears on first mismatch:

```ts
let snapGen = 0;
let lastSnap: { gen: number; top: number } | null = null;

function snapToBottom() {
  if (!containerEl) return;
  containerEl.scrollTop = containerEl.scrollHeight;
  lastSnap = { gen: ++snapGen, top: containerEl.scrollTop };
}

function handleScroll() {
  if (!containerEl) return;
  if (
    lastSnap &&
    lastSnap.gen === snapGen &&
    Math.abs(containerEl.scrollTop - lastSnap.top) <= 2
  ) {
    lastSnap = null;
    return;
  }
  lastSnap = null; // any non-matching event clears the sentinel
  // position-based logic below
}
```

#### 2. Detach must be position-based, not gesture-only

The reverted patch made `handleScroll` re-attach-only and put all detach on `wheel`/`touchmove`/`keydown` listeners. That misses every other path that scrolls the container: scrollbar-thumb drag (fires only `scroll`), find-in-page scroll-into-view, middle-click autoscroll.

**Fix.** Restore position-based detach in `handleScroll` *after* the sentinel filter. Use a wider band than re-attach so we don't oscillate:

```ts
const ATTACH_THRESHOLD_PX = 4;
const DETACH_THRESHOLD_PX = 80;

function handleScroll() {
  if (!containerEl) return;
  // sentinel filter (§1) here
  const distanceFromBottom =
    containerEl.scrollHeight - containerEl.scrollTop - containerEl.clientHeight;
  if (distanceFromBottom < ATTACH_THRESHOLD_PX) followBottom = true;
  else if (distanceFromBottom > DETACH_THRESHOLD_PX) followBottom = false;
}
```

The gesture listeners then become largely redundant — keep the text-selection drag suppression but drop the rest in favour of the position check.

#### 3. Streaming text drifts off the bottom

Deleting the ResizeObserver `\$effect` and adding `overflow-anchor: none` together meant nothing re-anchored during pure text-token streaming inside an existing assistant message — `messages.length` and `tailToolCallCount` don't change when MarkdownBody mutates its `html` state.

**Fix (recommended).** Restore the ResizeObserver gated on `followBottom`. Cheap, already-debounced via rAF, smallest delta from current behaviour.

Alternatives: drop `overflow-anchor: none` back to `auto` (verify tool-card mount above the viewport doesn't visibly shift), or add a text-streaming generation counter that `MarkdownBody` bumps on each throttled flush (more intrusive).

### Test plan

- **Manual.** Repro the original "scroll-up gets yanked back" bug; confirm scrollbar-drag, find-in-page, and middle-click autoscroll all detach correctly; confirm streaming text stays pinned when `followBottom === true`.
- **Unit-ish.** The scroll logic is hard to test without a real layout engine, but at minimum a unit test that calls the detach/attach logic with synthetic `scrollTop`/`scrollHeight`/`clientHeight` and asserts the `followBottom` transitions in both directions, plus a programmatic-snap-doesn't-flip case.
- **Regression.** Existing chat test suite — these internals aren't covered today, so the bar is "no new failures".

### Out of scope for follow-up

- Virtualizer behaviour during massive backscroll (the existing `virtual-inner` setup handles this).
- Touch momentum scrolling on iOS — different gesture model, separate follow-up.

PR #302, opened 2026-05-13
FSM: union allow_always grants into action toolset (fixes #280)

### LINKED ISSUE #280: request_tool_access doesn't actually grant the requested tool mid-action
## What I expected

When an FSM action calls `request_tool_access(toolName, reason)` and the user clicks Allow, the LLM should be able to call `toolName` for the rest of that action.

## What actually happens

The block is real, but the granted boolean has no path to actually unlock the tool. The LLM gets `{ ok: true, granted: true }`, then tries to call the underlying tool — the Anthropic API rejects it because the tool isn't in the action's registered `tools` array. The grant record exists in storage, but nothing reads it except `request_tool_access` itself.

## Trace

1. `buildTools` runs once at action start (`packages/fsm-engine/fsm-engine.ts:3200-3356`). It assembles the LLM's tool array from `action.tools`, workspace MCP servers, and the `atlas-platform` ambient set, then hands that fixed array to `streamText`. The toolset doesn't change for the rest of the action.

2. `request_tool_access` blocks the run via `waitForTerminalElicitation` (`packages/mcp-server/src/tools/permissions/request-tool-access.ts`). That pause is real. The pause is the only thing about this tool that genuinely affects the action.

3. On `allow_always`, `ToolAccessGrants.grantAlways` persists `(workspaceId, toolName)` to storage. But `ToolAccessGrants.hasGrant` is only consulted from inside `request_tool_access` itself (to short-circuit duplicate elicitations). I grepped `fsm-engine`, `agent-server`, `agent-sdk`, `core/delegate`, `core/agent-conversion` — no other consumer. So the grant is a cache for future elicitations, not a runtime permission gate.

4. The only mechanism that actually expands the LLM's callable surface is `permissions.dangerouslySkipAllowlist` at the job/workspace/daemon-env level (`fsm-engine.ts:3322-3329`). When that's on, `buildTools` skips per-agent narrowing and the LLM gets every platform-allowlisted tool — and `request_tool_access` returns `bypass` as a no-op formality.

## So what does request_tool_access actually do?

- Blocks the action until the user answers.
- Returns a `granted` boolean the LLM can use as a *signal* — to decide whether to route around (delegate to a sub-agent that does have the tool, fail-step gracefully, etc.). The LLM cannot use the granted response to invoke the originally-requested tool in the same action.
- For `allow_always`, primes the grant cache so the next `request_tool_access` for the same tool short-circuits. The LLM still needs the tool in the action's `tools` array to actually use it.

## Why this matters

The tool reads like a permission gate ("request access"), and the workspace-jobs SKILL.md describes the allow path as "the same action continues" — easy to read as "the LLM can now call the tool." It cannot. Authors who declare `tools: ['fs_read_file', 'request_tool_access']` expecting the LLM to gain `fs_write_file` after an allow will be confused when the next tool call fails.

## Possible directions

Pick one or some combination:

1. **Union declared tools with prior `allow_always` grants when `buildTools` runs.** Cross-action only — once the user has allow-always-ed a tool, every future action in that workspace sees it. Cheapest to ship. Doesn't help `allow_once`.

2. **Mid-action toolset rebuild on grant.** Interrupt `streamText`, rebuild the toolset including the newly-granted tool, resume. Closest match to the tool's name, but mechanically fiddly with current ai-sdk semantics.

3. **Dynamic dispatch tool.** Add a `call_tool(name, args)` meta-tool the LLM can use to invoke any granted-this-session tool. Sidesteps the static-toolset constraint, but introduces an indirection layer.

4. **Tighten the framing.** Rename or re-document the tool as an "approval signal" rather than an access grant. Update the skill text to say the LLM has to route around, not retry. Cheapest but doesn't fix the underlying gap.

## Files touched in this trace

- `packages/fsm-engine/fsm-engine.ts` (`buildTools`)
- `packages/mcp-server/src/tools/permissions/request-tool-access.ts`
- `packages/core/src/elicitations/tool-access-grants.ts`
- `packages/core/src/agent-conversion/agent-tool-filters.ts` (`wrapPlatformToolsWithScope` — no permission gate)
- `packages/system/skills/writing-workspace-jobs/SKILL.md` (the misleading "action continues" wording)

odk-:
## Summary

`request_tool_access` advertised itself as a permission gate but never affected what tools the LLM could actually call. `buildTools` (`packages/fsm-engine/fsm-engine.ts`) snapshots the toolset once per action and ignored `ToolAccessGrants`; the grant store was only consulted by `request_tool_access` itself to short-circuit duplicate elicitations. A user who `allow_always`'d `workspace-mcp/create_task` would see `granted: true`, then watch the next action's tool call fail because nothing in production code read that grant.

This PR closes the gap end-to-end across four commits.

### 1. `ToolAccessGrants.listForWorkspace` + storage server-awareness

- New `listForWorkspace` method on the storage adapter + facade. KV prefix scan over `keySegment(workspaceId).*`, drains keys before fetching values (matching the `jetstream-adapter` pattern to dodge the nats.js v2.29 iterator-termination bug).
- `ToolAccessGrantSchema` gains an optional `serverId` field. Additive; legacy grants keep deserializing.
- `grantAlways` derives `serverId` from a qualified `toolName` when the caller doesn't pass one explicitly. KV key derivation is unchanged (still based on the full `toolName`), so `hasGrant` lookups by the original LLM-facing string keep resolving both legacy and new grants.
- `listForWorkspace` returns `ListedGrant[]` with `{ toolName, bareToolName, serverId? }`. `bareToolName` is what `mcpResult.tools` is keyed by; `serverId` is the source MCP server when known.

### 2. `buildTools` union + eager-load

- Hoist the grant fetch above `effectiveConfigs` choice so the engine can react to grants before MCP servers are loaded.
- Any grant-referenced workspace MCP server the workspace still declares is **eagerly added** to `effectiveConfigs` before `createMCPTools`. Closes the "strictly-qualified action + grant for a different server's tool" failure mode — a grant for `gmail/send_email` now works even when the action declares only `google-calendar/list_events`.
- Stale grants (server dropped from workspace config since the grant was made) are silently skipped — no MCP load attempt, no confusing error.
- After the per-server narrowing step, the union widens `scoped` with any granted tool the engine actually loaded. **serverId-aware attribution:** a grant with `serverId: "gmail"` only widens when `mcpResult.toolsByServer.gmail` actually contains the bare tool name. Defense in depth against cross-server bare-name collisions. Legacy bare-name grants fall through to the existing bareToolName-in-`filtered` check.
- Bypass mode (`dangerouslySkipAllowlist`) skips the query — it already widens past per-agent narrowing, so the union would be redundant. Failure to read grants → warn-log + continue with the un-widened toolset; no crash, no escalation.

### 3. `request_tool_access` distinguishes `allow_once` vs `allow_always` for the LLM

- `persistent: boolean` is now first-class on every answered/bypass/persistent_allow response shape — previously present only when `true`. The LLM branches on a single field instead of parsing `answer`.
- `grantType` lands in the operator info-log (`allow_once` / `allow_always` / `bypass` / `persistent_allow` / `deny`) so audits don't need to join against the elicitation store.
- Mirrored on the chat-side parallel (`packages/system/agents/workspace-chat/tools/request-tool-access.ts`).
- Tool descriptions and SKILL.md teach the LLM to read `persistent` and behave accordingly.

### 4. SKILL.md rewrite

- Drops the "same action continues" framing that the code path never supported.
- Re-documents `request_tool_access` as an approval signal: `allow_once` requires routing around in-action; `allow_always` makes the tool available from the next action onward.
- Adds a "Reading the response" subsection covering `persistent` + the operator `grantType` log field.
- Worked example switched from `fs_write_file` (ambient platform tool, no narrowing applies) to a workspace MCP tool, which is where the narrowing actually applies.

### 5. `vitest.setup.ts`

Bootstraps + binds the tool-access-grant adapter so every suite that exercises `buildTools` end-to-end has the storage facade available.

## Test plan

- [x] `deno task typecheck` — green
- [x] `deno task test` — 6495 / 6514 pass, no regressions; the 19 skipped are pre-existing
- [x] `deno run -A npm:@biomejs/biome check` — 0 errors
- [x] Storage tests cover empty bucket, workspace scoping, reserved-character round-trip, serverId derivation from qualified `toolName`, and explicit-`serverId` override.
- [x] `grant-union.test.ts` covers (10 cases):
  - happy path: granted workspace-MCP tool unions in
  - no grant → declared-only narrowing holds
  - grant for an unloaded tool doesn't materialise (defense in depth)
  - bypass mode skips the grant query entirely (spy assertion)
  - grant-read failure → warn-and-continue with un-widened toolset
  - action without a `tools:` array (`hasNameAllowlist === false`) → no-op union
  - serverId match widens; serverId mismatch does not
  - eager-load happy path: grant-referenced server joins the MCP load
  - eager-load skipped when the grant-referenced server isn't in workspace config

## Compat

- Schema change is additive. Existing grants deserialize cleanly; read-path infers `serverId` from a qualified `toolName` when feasible.
- `hasGrant` key derivation unchanged — back-compat with both legacy and new grants.
- `request_tool_access` response now always carries `persistent: boolean`. Previously present only when `true`. Existing exact-match tests updated; no behavioural regression.

## Scope notes

- **`allow_once` still can't widen the current action.** Mid-stream toolset rebuild against the AI SDK is meaningfully larger work; out of scope. SKILL.md tells authors to delegate / failStep on `allow_once`.
- **Adjacent paths don't need this.** Audited `apps/atlasd/routes/agents/run.ts` and `packages/workspace/src/runtime.ts` — neither has per-action `tools:` narrowing, so grants have nothing to widen there. The fsm-engine `buildTools` path is the sole narrowing surface.

Closes #280.


basedfriday:
@odk- could the mechanism actually be for `request_tool_access` be to reflect the user decision for always allow back into the `workspace.yml`? For the `allow_once` use case I get that we'll need an ephemeral grant, but seems like config is the right place to track permanent grants.

odk-:
> @odk- could the mechanism actually be for `request_tool_access` be to reflect the user decision for always allow back into the `workspace.yml`? For the `allow_once` use case I get that we'll need an ephemeral grant, but seems like config is the right place to track permanent grants.

I don't think so. If you allow it, then export the workspace and I import it does it mean I get it auto allowed? If it should ask on import then we can also have it without storing other ppl decisions in the workspace.yml.

PR #303, opened 2026-05-13
fix(fsm): make complete() work on non-Anthropic models

LissaGreense:
## Summary

The auto-injected `complete()` tool in FSM `outputTo:` actions only worked on Anthropic. Other providers either skipped the call or invoked `complete({})`, breaking every non-Claude FSM agent.

## Root cause

Two layers, both in the path that ships `complete()`'s schema to the provider:

1. **Schema too loose.** The default fell back to `z.record(z.unknown()).refine(non-empty, ...)`. That compiles to `{type: "object", additionalProperties: true}` with no `required` keys — the runtime `.refine()` doesn't translate to JSON Schema. Anthropic infers good keys from the description prose; everything else reads "object with any keys" as "empty is fine."
2. **Strict mode never set.** Even with a tighter schema, the AI SDK doesn't send `strict: true` on the function definition. OpenAI-compatible providers (Groq, OpenAI, OpenRouter, LiteLLM) treat tool schemas as advisory without it.

## Changes

**`packages/fsm-engine/fsm-engine.ts`** — replace the open `z.record` fallback with `z.object({ result: z.string().min(1) })`. Use `tool({...})` + explicit `aiJsonSchema(...)` so the AI SDK serializes the schema verbatim (no Zod→JSON quirks). Preserves the existing `compiledSchema` branch for actions with a declared `outputType:`.

**`packages/fsm-engine/llm-provider-adapter.ts`** — pass `providerOptions: { groq: { strictJsonSchema: true } }` (or `openai: { strictJsonSchema: true, structuredOutputs: true }`) into `streamText()` for every non-Anthropic call. Strict mode constrains generation to the schema server-side; providers reject `{}` instead of accepting it.

## Why no model whitelist

An earlier revision gated strict mode by a model-name regex (`llama-4-scout`, `gpt-4o`, etc.) to avoid HTTP 400 from models that don't support `response_format: json_schema`. That was wrong:

- Those same models are *already broken* under this code path — they silently call `complete({})` and the FSM throws "did not call complete."
- A 400 with a clear provider error message ("model X doesn't support json_schema") is a strictly better failure mode than silent empty output.
- New models (Groq adds them weekly, OpenAI adds them periodically) silently regress with a whitelist; they work by default without one.

So strict mode is enabled for **every** Groq + OpenAI call. Models that can't honor it fail visibly, with an actionable error. Models that can — including everything modern + everything routed through OpenRouter/LiteLLM — get schema enforcement automatically.

## What still needs work (separate PRs)

- **Workspace-chat path** (`packages/core/src/agent-conversion/from-llm.ts:154`) has the same `complete()` injection and the same brittleness. Not touched here.
- **The synthetic `complete()` tool itself** — when an FSM action declares no `outputType:`, we still inject a `{result: string}` schema and force the model through it. Better default would be to skip injection entirely and use the streamed text response. Behavior change for ~28% of existing untyped `outputTo:` actions, needs its own review.
- **`OPENAI_BASE_URL` passthrough** for OpenAI-compatible endpoints (Ollama, LM Studio) — was originally bundled into this PR, split out to keep scope tight.

Tracked in `docs/plans/free-models-bugs.md`.

## Test plan

- [x] `@atlas/fsm-engine` and `@atlas/llm` test suites pass locally (with `friday.yml` moved aside)
- [x] End-to-end on `groq:meta-llama/llama-4-scout-17b-16e-instruct`: tool-calling FSM actions succeed across multiple runs, including memory writes that round-trip
- [x] Anthropic path unchanged — `providerOptions.groq/openai` is ignored by the Anthropic provider
- [x] Models that don't support `response_format: json_schema` now return a clear HTTP 400 instead of silently emitting empty output

---

## Follow-up: two more strict-mode gaps + eval matrix

Caught running the full `tools/qa` suite against three providers.

**2a. Per-action override bypassed strict mode.** `providerName` came from `defaultModel.provider`, so workspaces overriding provider per-action never engaged strict mode. Fix: scope the override to the strict-mode block only (keeps `isDefaultOptsProvider` semantics intact — broadening it would silently flip Anthropic `cacheControl` on every user message).

**2b. Typed `outputType:` lost `additionalProperties: false` on the wire.** The compiled-schema path round-tripped through Zod, which serialized `catchall: unknown` instead. OpenAI/Groq strict mode silently downgrades without `additionalProperties: false`. New `withStrictObjects()` walks the JSON Schema recursively, injects strictness at every object level, hands it to `aiJsonSchema()` verbatim.

## Eval matrix (FSM-only sub-tests, 20 scenarios)

`tools/qa/run-evals.sh --profile {anthropic,groq,openai}`. FSM-only scope excludes 4 workspace-chat scenarios (called out as "not touched here") + 1 Python user-agent plumbing case — same failures across all providers.

| Provider / Model | Pass | Notes |
|---|---|---|
| anthropic / claude-sonnet-4-6 | **20/20 — 100%** | baseline |
| openai / gpt-4o-mini (via OpenRouter) | 15/20 — 75% | small-model multi-field limit |
| groq / llama-4-scout-17b-16e-instruct | 12/20 — 60% | scout struggles more with 3-field required outputs |

Wire schema verified strict on every call via instrumented log during development. Remaining failures are scenarios whose `outputType:` has 3+ required fields — model-quality limit, not strict-mode regression. Single-field schemas pass on all providers.

## Test coverage

Added: 9 unit tests for `withStrictObjects` covering nested-objects-in-arrays, deep nesting, additionalProperties preserved, untyped-object trigger, immutability, non-record passthrough.

Skipped (intentionally): wire-payload mocks (would couple to AI SDK version-pinned internals; live matrix is the integration test); targeted strict-mode-override regression (workspace.yml fixtures already parameterize per-action provider/model, every Groq/OpenRouter run exercises it); harness-fix unit tests (inverse-assertion — the fact that daemon suites run at all is the regression signal).

## Targeted `complete()` tool-call eval

Added `tools/qa/complete-eval/` — promptfoo-native, no daemon. Each case is one HTTP call per provider per schema shape. The full 5-case × 3-provider matrix runs in ~17s with ~2K tokens (vs. the daemon-based suite at 50K+ tokens and several minutes), so it can be re-run cheaply on every change.

Five cases cover the schema shapes the FSM ships: 1-field smoke, 3 flat primitives (`ReviewResult`), primitive + `string[]` (`ArrayReviewResult`), array of nested objects (`EmailBatch` — the regression surface for `additionalProperties` propagation), 4 fields incl. boolean (`AgentHydrationResult`). Each case ships in two variants: post-fix (current wire shape) and pre-fix (permissive schema, no `strict`, no forced `tool_choice`).

| | Anthropic claude-sonnet-4-6 | Groq llama-4-scout | OpenRouter openai/gpt-oss-120b |
|---|---|---|---|
| Pre-fix | 4/5 | 3/5 | 5/5 |
| Post-fix | 5/5 | 5/5 | 5/5 |

**Pre-fix 12/15 (80%) → Post-fix 15/15 (100%).** The fix prevents three measurable failure modes:

- **Groq scout type coercion** (cases 2, 5): without `additionalProperties: false` + `strict: true`, scout emits `count: "12"` (string) instead of `12` (number) and the server-side validator 400s the call. With the fix, wire-level constrained decoding forces the right type.
- **Anthropic skipping the tool call** (case 3): without forced `tool_choice: { type: tool, name: complete }`, sonnet returns natural text. With the fix, the tool is forced.
- **Strict-mode-rejected `complete({})` calls** (not surfaced in these particular prompts, but the foundational reason the FSM rejects an empty/incomplete call vs accepting it silently).

This eval is the integration test for the wire-shape contract the fix produces — fast enough to iterate on, narrow enough to interpret.


basedfriday:
@LissaGreense need to regression test full eval suite on anthropic (should be 100% pass as that's main case today) and the new providers.

basedfriday:
@LissaGreense this work is looking good. We're gonna target release of this for `0.1.9` as we're focusing next release on some critical fixes.

Issue #299, opened 2026-05-13
Export Chat broken in v0.1.7 — 'unsupported format' error on open


basedfriday:
## Bug Report

**Version:** 0.1.7 (live installer)

**Description:**
Export Chat is not working. After exporting, attempting to open the file results in an &#39;unsupported format&#39; error.

**Steps to Reproduce:**
1. Use Export Chat feature in v0.1.7
2. Attempt to open the exported file

**Expected Behavior:**
Exported file opens successfully.

**Actual Behavior:**
Error: &#39;unsupported format&#39; when trying to open the exported file.

**Environment:**
- Version: 0.1.7 (live installer)

## Error Details

**macOS Archive Utility dialog:**
> Unable to expand "friday-chat-chat_4zc (1).zip". It is in an unsupported format.

PR #310, opened 2026-05-13
Chat-UI: export preview survives chats containing artifact cards

basedfriday:
## Summary
- Exporting a chat that contains any artifact card (`display_artifact` tool calls or lifted-artifact previews) no longer crashes the preview SSR — the page renders end-to-end from the prefetched artifact map.
- Live chat UI is byte-identical: same query keys, same fetches, same caching behavior. No new DOM mutations for `chat-message-list`'s virtualizer to re-measure.

## Why it broke
ArtifactCard called `createQuery()` unconditionally and relied on a null query id (→ `skipToken`) to "disable" the query in export mode. But the export-preview layout is intentionally chromeless and never mounts a `QueryClientProvider` — so the hook threw at `useQueryClient()` even when the query itself was skipped. `skipToken` disables the *query*, not the *hook*.

## Known limitation
Chats containing a `request_human_input` tool call still crash export-preview SSR with the same root cause in `human-input-tool-card.svelte` (`createQuery` + two mutation hooks). The proper fix renders a static view from the recorded answer in `call.output` instead of from live elicitation state — shape-different enough to track separately rather than bundle here.

## Test plan
- [ ] Open `/platform/<ws>/chat/<id>/export/preview` for a chat that contains at least one displayed artifact (image / PDF / tabular / generic). Confirm the artifact title, mime badge, asset URL, and any inline preview render — no spinner, no "Loading…" placeholder, no 500.
- [ ] Open the same chat in the live UI. Artifact cards still fetch from the daemon, table previews still render, download / open buttons still work.
- [ ] `deno task typecheck` clean and the agent-playground vitest suite is green (494/494 locally; new regression test pins the bug).

### DIFF
--- a/tools/agent-playground/src/lib/components/chat/__test-stubs__/artifact-card-export-parent.svelte
+++ b/tools/agent-playground/src/lib/components/chat/__test-stubs__/artifact-card-export-parent.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-  import { QueryClient, QueryClientProvider } from "@tanstack/svelte-query";
   import { setExportContext, type ExportContext } from "../export-context.ts";
   import ArtifactCard from "../artifact-card.svelte";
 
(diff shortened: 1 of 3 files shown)


basedfriday:
Fixes #299

PR #314, opened 2026-05-14
agent runtime: lift undici timeout, forward jobTimeoutMs, --directory spawn, drop bundled knowledge agent

ljagiello:
## Summary

Four commits, +57 / −2369. Three are infrastructure for running user-space Python agents reliably; the fourth removes the bundled `knowledgeHybridAgent` now that the user-space replacement lives in a separate workspace repo.

### 1. Agent timeouts

- **`playground/proxy: disable undici 5-min headersTimeout for daemon calls`** — Node 22+'s undici defaults a 5-minute `headersTimeout` and `bodyTimeout` on the global `fetch`. Long-running daemon endpoints (chiefly `POST /api/workspaces/{id}/signals/{id}` for jobs like `reindex`, where the daemon awaits the cascade before returning headers) exceed that and the proxy fails with `UND_ERR_HEADERS_TIMEOUT`. Adds a dedicated `undici.Agent` with a 1-hour `headersTimeout` / `bodyTimeout` (vs disabled — keeps server-side back-pressure on a wedged daemon) and threads it as `dispatcher` on the proxy fetch. Connection-level timeouts (TCP, TLS handshake) keep their defaults — those still indicate real upstream sickness.
- **`runtime: wire jobTimeoutMs into ProcessAgentExecutor.execute`** — `executeCodeAgent` was constructing its `ProcessAgentExecutor` options without passing the FSM job's `timeoutMs`, so a `config.timeout: "2h"` on a job in `workspace.yml` was meaningful only for elicitation TTL — the actual subprocess kill ceiling silently fell back to the executor's 180s default. One-line conditional spread closes the gap for `type: user` agents.

### 2. Python user-agent spawn

- **`agent-spawn: use --directory for project-style Python agents`** — when an agent dir contains `pyproject.toml`, spawn via `uv run --directory <agent_dir> --with friday-agent-sdk==<ver> --python 3.12 python <agentPath>` so the project's declared dependencies (e.g. fireworks-ai, sqlite-vec, groq) are installed from its lockfile **and** the launcher's SDK pin still applies. Previously only `--with` was passed, which broke any agent that needed additional libraries.

### 3. Knowledge agent migrates out of bundled into user-space

- **`Remove bundled knowledge-hybrid agent`** — the hybrid-RAG knowledge agent now ships as a Python user agent in a separate workspace repo. Drops the bundled TS port (6 source + 6 test files, ~2.3K lines) plus the `@ai-sdk/fireworks`, `@ai-sdk/groq`, and `@db/sqlite` dependencies. Decouples knowledge-agent release cadence from the daemon.

## Test plan

- [x] `deno task dev:playground` — no undici timeout when triggering long-running signals (verified end-to-end against a real `answer-ticket` flow that now consistently completes in ~60-90s rather than failing at exactly 5min)
- [x] `agent-spawn --directory` — both single-file and project-style Python agents install + spawn correctly; `agent-spawn.test.ts` covers both branches plus the SDK-pin invariant
- [x] `jobTimeoutMs` forwarding — `ProcessAgentExecutor` receives the `timeoutMs` opt when set on the FSM job; the conditional-spread guard preserves `timeoutMs: 0` (would-be "no ceiling") rather than coercing through truthy
- [x] Bundled removal — no remaining references to `knowledgeHybridAgent` / `knowledge-hybrid` in the daemon, registry, or bundled-agents exports
- [x] `proxy.test.ts` — new assertions cover dispatcher singleton + the elapsed-ms surfaced on 502 (observability for slow upstreams)
- [x] `deno task lint:check` + `npx knip` + `deno task test --run` — clean


basedfriday:
Merged latest `main` into this branch (`0b1ca8c6`) and pushed.

One conflict, in `tools/agent-playground/src/lib/server/proxy.ts`:
- `main` extracted the shared `executeProxyFetch` helper (now used by both the SvelteKit and Hono proxy paths).
- this branch added the long-lived undici dispatcher (1-hour timeout ceiling) + elapsed-time error logging, inline in `buildProxyHandler`.

Resolved by keeping `main`'s `executeProxyFetch` structure and folding this branch's `dispatcher: longLivedDispatcher` + elapsed-time logging into `executeProxyFetch` — so both proxy runtimes get the longer timeout ceiling.

Verified: no leftover conflict markers, `vitest` proxy suites 12/12 pass (incl. `main`'s new `proxy.hono.test.ts`), `svelte-check` 0 errors.

@ljagiello heads up ^

apps/atlasd/src/agent-spawn.ts
```diff
@@ -20,16 +28,34 @@
...
   if (agentPath.endsWith(".py")) {
     const uvPath = process.env.FRIDAY_UV_PATH;
     const sdkVersion = process.env.FRIDAY_AGENT_SDK_VERSION;
     if (uvPath && sdkVersion) {
-      return [
-        uvPath,
-        ["run", "--python", "3.12", "--with", `friday-agent-sdk==${sdkVersion}`, agentPath],
-      ];
+      const agentDir = dirname(agentPath);
+      const hasPyproject = existsSync(join(agentDir, "pyproject.toml"));
```
basedfriday, line 48: **Important** — `dirname(agentPath)` on a relative path keys pyproject detection off the daemon's cwd. `RegisterRequestSchema` (`routes/agents/register.ts`) documents `entrypoint` as "the absolute path" but only validates `z.string().min(1)`. A relative path makes `dirname("agent.py")` return `"."`, so `existsSync` resolves `pyproject.toml` against `process.cwd()` and `--directory .` points uv at the wrong tree. Callers likely pass absolute paths today (the old bare-`--with` shape never cared), and the new tests all use absolute tmp paths — so a regression here would be invisible.

Suggest `resolve(agentPath)` at the top of `buildAgentSpawnArgs`, or `.refine(isAbsolute)` on the schema.

tools/agent-playground/src/lib/server/proxy.test.ts
```diff
@@ -79,13 +79,56 @@ describe("buildProxyHandler", () => {
...
+    const handler = buildProxyHandler({ upstream: "https://daemon.local:8080", label: "daemon" });
+    await handler(event({ path: "long" }));
+    const init = fetchMock.mock.calls[0]?.[1] as
+      | { dispatcher?: { constructor?: { name?: string } } }
+      | undefined;
+    expect(init?.dispatcher).toBeDefined();
+    // It's an undici Agent instance; we can't import the value here
+    // (that would couple this test to undici internals), but the
+    // constructor name is a sufficient smoke check.
+    expect(init?.dispatcher?.constructor?.name).toBe("Agent");
```
basedfriday: **Important** — this assertion passes through the exact regression it's meant to guard. `constructor.name === "Agent"` confirms the dispatcher is *an* undici `Agent`, but not that it carries the 1-hour `headersTimeout`/`bodyTimeout` — which is the whole point. `new Agent({})` with the default 5-min timeouts passes this test while reintroducing `UND_ERR_HEADERS_TIMEOUT`. The constructor-name check is also brittle to undici internals.

Suggest exporting `longLivedDispatcher` (or its timeout config) from `proxy.ts` and asserting identity (`toBe(longLivedDispatcher)`) or the timeout value. The "can't import the value" comment is unfounded — and if you do this, the singleton test below collapses into the same identity assertion.

tools/agent-playground/src/lib/server/proxy.test.ts
```diff
@@ -79,13 +79,56 @@ describe("buildProxyHandler", () => {
...
     expect(json.error).toMatch(/^tunnel proxy fetch failed:/);
     expect(json.error).toContain("connection refused");
+    // Elapsed ms is surfaced both in the JSON body and via console.warn
+    // so a slow-then-failing upstream is observable in dev logs (no
+    // observability used to exist for the proxy fail path).
+    expect(typeof json.elapsedMs).toBe("number");
+    expect(json.elapsedMs).toBeGreaterThanOrEqual(0);
+    expect(warnSpy).toHaveBeenCalledOnce();
+    expect(warnSpy.mock.calls[0]?.[0]).toMatch(/\[tunnel proxy\] fetch failed after \d+ms:/);
+    warnSpy.mockRestore();
```
basedfriday: **Important** — `warnSpy.mockRestore()` as the last line of the test body isn't failure-safe: any assertion throwing between the spy setup and here leaks the spy into later tests. The file's `afterEach` only restores `globalThis.fetch`, not mocks. Suggest `afterEach(() => vi.restoreAllMocks())` and drop the inline `mockRestore()`.

packages/workspace/src/runtime.ts
```diff
@@ -2934,6 +2934,18 @@ export class WorkspaceRuntime {
...
+        // workspace.yml. Without this, ProcessAgentExecutor falls back to
+        // its silent 180s default and a `config.timeout: "2h"` setting on
+        // the job is meaningful only for elicitation TTL, not for actually
+        // letting the agent run that long.
+        //
+        // KEEP `!== undefined` — a truthiness check (`opts.jobTimeoutMs && ...`)
+        // would silently drop `timeoutMs: 0`. The executor treats undefined as
+        // "fall back to default" but `0` could plausibly mean "no ceiling" in
+        // a future config schema; preserve the distinction at this boundary.
+        ...(opts.jobTimeoutMs !== undefined && { timeoutMs: opts.jobTimeoutMs }),
```
basedfriday: **Important** — this conditional-spread ships with zero tests. The inline comment itself flags the footgun (`!== undefined` vs truthiness, to preserve `timeoutMs: 0`) — exactly the logic that breaks silently under a careless refactor. It's testable at this layer: assert the executor options object includes/excludes `timeoutMs` for `jobTimeoutMs` of `undefined`, `0`, and a large value.

tools/agent-playground/src/lib/server/proxy.ts
```diff
@@ -1,4 +1,35 @@
 import type { RequestHandler } from "@sveltejs/kit";
+import { Agent } from "undici";
+
+// Node 22+ ships undici 6.x with a 300_000ms (5-min) default
```
basedfriday: **Minor** — stale version claim. Says "Node 22+ ships undici 6.x"; the runtime actually bundles undici 7.x. The bug premise still holds and the standalone `^7` dep is aligned with the bundled version (so no skew risk), but the version detail will mislead a future reader. Suggest "Node's bundled undici ships a 300_000ms default `headersTimeout`" without the version number.

tools/agent-playground/src/lib/server/proxy.ts
```diff
@@ -1,4 +1,35 @@
...
+//
+// Bound — don't disable. A `0` ceiling means a wedged daemon parks
+// SvelteKit request workers indefinitely with no server-side back-
+// pressure. Use a dedicated dispatcher with a 1-hour ceiling: covers
+// the worst documented job (30-min reindex) with headroom, and still
+// fails fast on a genuinely-stuck socket. Connection-level timeouts
+// (TCP, TLS handshake) keep their defaults because those still
+// indicate actual upstream sickness, not slow work.
+const ONE_HOUR_MS = 60 * 60_000;
+const longLivedDispatcher = new Agent({
```
basedfriday: **Minor** — module-scoped `Agent` has no cleanup. Under SvelteKit dev HMR the module re-evaluates and orphans the previous `Agent`'s socket pool. Harmless in production, mildly leaky in dev — fine to leave, or optionally `process.once("exit", () => longLivedDispatcher.close())`.

basedfriday (commented):
Code review (architecture/correctness + test quality lenses). Verdict: **Needs Work** — all four changes are sound in intent and the knowledge-agent removal is verified clean (no dangling code references), but two new `proxy.test.ts` tests don't pin the behavior they guard, and there's a relative-path edge case in `agent-spawn`. No critical issues. Inline comments below.

**Needs Decision — knowledge-agent removal needs a coordinated cross-repo update.** The commit footer notes workspaces referencing `agentId: knowledge-hybrid` (`type: atlas`) must switch to `type: user, agent: knowledge`. Code-side removal in this repo is complete (greps for `knowledgeHybrid` / `knowledge-hybrid` / `KnowledgeOutputSchema` / `KnowledgeResult` are empty except a historical CHANGELOG entry, no `workspace.yml` here references it). Just confirm the matching workspace-bundle update lands alongside this so existing workspaces don't break on deploy.

apps/atlasd/src/agent-spawn.ts
ljagiello: Fixed in e26fde037.

`buildAgentSpawnArgs` now does `isAbsolute(agentPath) ? agentPath : resolve(agentPath)` at the top of the uv branch, then derives `agentDir` from the resolved path and threads `absoluteAgentPath` into both arg shapes. Added a relative-path test that asserts the resulting agent-path arg is absolute and ends with the expected suffix — pins the cwd-independence contract regardless of what callers pass.

Didn't touch `RegisterRequestSchema` itself (out of scope), but the spawn layer is now self-defending.

tools/agent-playground/src/lib/server/proxy.test.ts
ljagiello: Fixed in e26fde037.

Exported `longLivedDispatcher` and `PROXY_DISPATCHER_TIMEOUT_MS` from `proxy.ts`. The dispatcher test now asserts `expect(dispatcherA).toBe(longLivedDispatcher)` (and same for B) — collapses into the singleton check as you suggested, and a bare `new Agent({})` regression now fails it. Added a sibling test that pins `PROXY_DISPATCHER_TIMEOUT_MS === 60 * 60_000` so a future shrink of the ceiling is intentional, not accidental.

The "can't import the value" comment that triggered the brittle assertion is gone.

tools/agent-playground/src/lib/server/proxy.test.ts
ljagiello: Already addressed in 0734e2d81 (the previous fixup landing the round-2 review fixes).

`afterEach` now calls `vi.restoreAllMocks()`, the inline `warnSpy.mockRestore()` is gone, and the test's spy gets cleaned up even if the assertions between setup and the would-be restore throw. Comment near the spy install now says "the afterEach `vi.restoreAllMocks()` cleans this spy up even if the assertions below throw — no manual mockRestore needed".

packages/workspace/src/runtime.ts
ljagiello: Fixed in e26fde037.

Extracted the spread shape into `buildExecutorTimeoutOption(jobTimeoutMs)` — a tiny pure helper next to `parseJobTimeoutMs` that takes `number | undefined` and returns `{}` or `{ timeoutMs: N }`. Three new cases in `runtime-job-timeout.test.ts` pin the contract directly:

- `undefined` → `{}` (executor uses its default)
- positive number → `{ timeoutMs: N }`
- `0` → `{ timeoutMs: 0 }` (foot-gun guard against a truthy refactor; `parseJobTimeoutMs` already filters `<= 0` upstream, but the spread-shape contract stays load-bearing on its own)

The call site at line 2962 now reads `...buildExecutorTimeoutOption(opts.jobTimeoutMs)` — same emitted JS, but the `!== undefined` distinction is no longer hidden inside an inline conditional spread.

tools/agent-playground/src/lib/server/proxy.ts
ljagiello: Fixed in e26fde037. Comment now reads "Node's bundled undici" without the version number, per your suggestion.

tools/agent-playground/src/lib/server/proxy.ts
ljagiello: Skipping per your "fine to leave" note — the dispatcher is module-scoped so the orphaned-pool concern only applies under SvelteKit dev HMR, and `process.once("exit", ...)` would only fire in production where the leak doesn't exist. Tracking as future polish if anyone notices socket-pool growth in long-running dev sessions.

packages/workspace/src/__tests__/runtime-no-op-signal-status.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  if (runtime) {
+    await runtime.shutdown().catch(() => {});
+    runtime = undefined;
+  }
+  if (originalAtlasHome === undefined) delete process.env.FRIDAY_HOME;
+  else process.env.FRIDAY_HOME = originalAtlasHome;
+  await rm(testDir, { recursive: true, force: true });
+});
+
+it("[FAILING] flags a signal that took no transition as SKIPPED, not COMPLETED", async () => {
```
Vpr99: **[Critical] This will actually fail vitest.**

It's a plain `it(...)`, not `it.fails()` / `.todo()` / `.skip` — the `[FAILING]` prefix is just a string in the test name, Vitest has no special handling for it. The docstring confirms the runtime bug at `runtime.ts:2083` is unfixed, so this assertion legitimately fails today and breaks CI for anyone running `vitest` on the workspace package.

Pick one:
- `it.fails("[red landmark for #322] flags a signal that took no transition as SKIPPED…", …)` — makes failure the pass state and the intent unambiguous in Vitest semantics.
- `it.todo(…)` until the #322 fix lands.
- Split the test into a follow-up PR that ships with the runtime/engine fix.

As-is, anyone seeing red doesn't know whether it's the known landmark or a real regression.

packages/workspace/src/runtime-job-timeout.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  it("rejects 0 — would otherwise become an instant-rejection in nats.js", () => {
+    // Foot-gun guard: nats.js `nc.request({ timeout: 0 })` rejects on the
+    // next tick, so a workspace.yml `config.timeout: "0s"` would silently
+    // kill every job in scope. Authors who write `0` almost certainly
+    // mean "no ceiling"; surface a warn and fall back to the executor
+    // default (caller treats undefined as "use default").
+    expect(parseJobTimeoutMs("job", "0s")).toBeUndefined();
+    expect(parseJobTimeoutMs("job", "0ms")).toBeUndefined();
```
Vpr99: **`"0ms"` exercises the malformed-input branch, not the new `<= 0` guard.**

`parseDuration` regex in `packages/config/src/base.ts:40` is `^(\d+)([smh])$` — accepts `s/m/h` only. `"0ms"` fails the regex, throws, hits the `catch` block, returns `undefined`. So only the `"0s"` assertion on line 23 actually proves the new `<= 0` guard fires.

If a future refactor drops the `<= 0` check, this test still passes on `"0ms"`.

Table-drive `0s / 0m / 0h` (all three hit the guard), or add a `vi.spyOn(logger, "warn")` assertion to lock the operator-signal.

apps/atlasd/src/agent-spawn.test.ts
```diff
@@ -42,6 +45,100 @@ describe("buildAgentSpawnArgs", () => {
...
+        expect(args.slice(0, -1)).toEqual([
+          "run",
+          "--python",
+          "3.12",
+          "--with",
+          "friday-agent-sdk==0.1.8",
+        ]);
+        const finalPath = args[args.length - 1] ?? "";
+        expect(finalPath.startsWith("/")).toBe(true);
+        expect(finalPath.endsWith("/relative/agent.py")).toBe(true);
```
Vpr99, line 138: **The relative-path test doesn't assert `--directory` is absent.**

The whole point of this branch (no `pyproject.toml`) is that `--directory` should NOT be prepended. But the test only checks the resolved path shape — a regression that pre-pended `["run", "--directory", "/some/cwd"]` while preserving the absolute-path suffix would still pass.

One-line fix:
```ts
expect(args).not.toContain("--directory");
```

Bonus: if you want the cwd-pollution guard to be a real test rather than a comment, `process.chdir` to a tmpdir without `pyproject.toml` in `beforeEach` and restore in `afterEach`.

tools/agent-playground/src/lib/server/proxy.test.ts
```diff
@@ -79,13 +89,56 @@ describe("buildProxyHandler", () => {
...
+    expect(dispatcherA).toBe(longLivedDispatcher);
+    expect(dispatcherB).toBe(longLivedDispatcher);
+  });
+
+  it("dispatcher timeout matches the exported constant (1 hour by default)", () => {
+    // Pin the timeout value so a future bump or shrink is intentional.
+    // 1 hour covers the worst documented job (30-min reindex) with
+    // headroom; shorter values would re-create the original bug for
+    // the longest-running signals.
+    expect(PROXY_DISPATCHER_TIMEOUT_MS).toBe(60 * 60_000);
```
Vpr99, line 142: **Pinning the constant doesn't lock both timeout knobs.**

This assertion proves `PROXY_DISPATCHER_TIMEOUT_MS === 60 * 60_000` but undici doesn't expose dispatcher options for direct inspection — so a partial-revert that does `new Agent({ headersTimeout: 5000, bodyTimeout: PROXY_DISPATCHER_TIMEOUT_MS })` (half-baked fix) passes both this and the dispatcher-identity check on line 116.

Cheap closeup: extract the options into a tested constant.
```ts
export const PROXY_DISPATCHER_OPTIONS = {
  headersTimeout: PROXY_DISPATCHER_TIMEOUT_MS,
  bodyTimeout: PROXY_DISPATCHER_TIMEOUT_MS,
} as const;
const longLivedDispatcher = new Agent(PROXY_DISPATCHER_OPTIONS);
```
Then assert `PROXY_DISPATCHER_OPTIONS.headersTimeout === PROXY_DISPATCHER_OPTIONS.bodyTimeout`. Or accept the gap and rely on the long-signal integration test.

Vpr99 (commented):
Drive-by review from the polish pass. Bones look clean — four findings worth addressing before merge, all low-cost. Full writeup lives in `docs/reviews/2026-05-14-pr-314-runtime-fixes.md` on this branch.

packages/workspace/src/__tests__/runtime-no-op-signal-status.test.ts
ljagiello: Good catch — addressed in `bfb846160`. Switched to `it.fails(...)` so CI treats "still failing" as the expected state; when the runtime/engine fix lands and the inner assertion starts passing, vitest will fail this test and force someone to drop the annotation. Self-cleaning bug tracker.

packages/workspace/src/runtime-job-timeout.test.ts
ljagiello: Right — `parseDuration`' regex only accepts `s|m|h`, so `"0ms"` was hitting the malformed-input branch, not the `<= 0` guard. Fixed in `23cc88e63` with `it.each(["0s", "0m", "0h"])` so each unit hits the guard via a value that genuinely parses to 0.

apps/atlasd/src/agent-spawn.test.ts
ljagiello: Added in `23cc88e63`: `expect(args).not.toContain("--directory")` after the existing path-shape assertions. Now a regression that prepends `["run", "--directory", "<cwd>", ...]` would trip this test.

tools/agent-playground/src/lib/server/proxy.test.ts
ljagiello: Took the extract-into-tested-constant path in `23cc88e63`. Added `PROXY_DISPATCHER_OPTIONS = { headersTimeout: PROXY_DISPATCHER_TIMEOUT_MS, bodyTimeout: PROXY_DISPATCHER_TIMEOUT_MS } as const`, the `Agent` is now constructed from it, and added an assertion that both knobs reference the same source-of-truth so a partial-revert that drops one would trip the test.

Vpr99 (approved):

PR #320, opened 2026-05-14
Launcher TLS readiness probe + OAuth callback shim for desktop installs

ljagiello:
## Summary

Three intertwined fixes the **desktop install** hit once the s2s mesh (#243, #300, #306) was on — none of them surface on host-side dev because the dev playground binds at `localhost:5200`. From the original installer screenshot:

> Some services are still starting / Waiting on: Authentication, Webhook tunnel, Studio UI, Friday daemon.

…through "Link service is unavailable right now" in chat, all the way down to the Gemini "Success! Credentials Ready" manual JSON-paste page on Google OAuth.

### 1. Native TLS-aware readiness probe in friday-launcher

Process-compose's `HttpProbe` has no TLS knobs (no client, no `InsecureSkipVerify`, no CA path). After the s2s flip, the probe hit our private-CA-signed leaf via the default Go transport, failed every 2s, and the supervisor bounced every service at 62s.

- `tools/friday-launcher/readiness.go` — native `readinessRunner` per spec; `http.Client` with `tls.Config{InsecureSkipVerify: true}` on loopback (we own both ends; CA plumbing buys no security on 127.0.0.1). Drives `HealthCache.SetReady` and calls `sup.RestartProcess` after `probeFailureThreshold` consecutive failures.
- process-compose's `ReadinessProbe` set to `nil`; `deriveStatus` consults `customReady` instead of `ps.Health`.

### 2. `.env` URL migration + defensive auto-upgrade

The Tauri installer pinned `FRIDAYD_URL` / `EXTERNAL_DAEMON_URL` / `EXTERNAL_TUNNEL_URL` / `LINK_SERVICE_URL` to `http://` before TLS landed. Stale `.env`s were sending cleartext into TLS listeners.

- `tools/friday-launcher/cert_env.go` — `migrateStaleURLSchemes()` rewrites those four keys to `https://` when `s2sCertsValid()`. Idempotent, no-op when TLS off, leaves explicit `https://` alone.
- Consumers also auto-upgrade defensively (same pattern already in `packages/openapi-client/src/utils.ts:63`): `apps/atlasd/routes/{link,mcp-registry}.ts`, `apps/atlasd/src/services/communicator-wiring.ts`, `tools/agent-playground/static-server.ts`.

### 3. OAuth callback shim for Gemini's Cloud Function

Gemini CLI's Cloud Function ([cloud_function/index.js#L91-L104](https://github.com/gemini-cli-extensions/workspace/blob/main/cloud_function/index.js#L91-L104)) does a literal hostname check — only `localhost` / `127.0.0.1` pass. The desktop install opens the playground at `https://local.hellofriday.ai:15200` (the only name the browser-trusted cert covers), so the X-Forwarded-Host-derived callback URL fails the check and users land on the manual JSON-paste page.

- `tools/agent-playground/static-server.ts` — second `Deno.serve` on `http://127.0.0.1:<PLAYGROUND_OAUTH_SHIM_PORT>` that 302-redirects `/api/daemon/api/link/v1/callback/*` to the TLS origin, preserving query string.
- `apps/link/src/oauth/service.ts` — `rewriteToOAuthShim(callbackUrl, FRIDAY_OAUTH_SHIM_BASE)` swaps host+port+scheme on the delegated flow's `finalRedirectUri`. JWT state's `c` field stays unchanged so the callback handler validates against the real TLS origin.
- `tools/friday-launcher/project.go` — `commonServiceEnv()` injects both env vars when `hasValidBrowserCert()`. Shim port = playground port + 1 (overridable).

## Test plan

- [x] Launcher: `go test ./tools/friday-launcher/...` — all green (incl. new readiness, probe-scheme, url-migration, oauth-shim tests).
- [x] Link: `deno task test src/oauth/oauth-shim.test.ts` — 6 cases (round-trip, fallbacks).
- [x] VM: all 5 supervised services reach healthy in ~70s, no restart loops, no TLS handshake errors.
- [x] VM: `.env` after restart shows all 4 URLs flipped to `https://`; launcher log `keys_migrated:4`.
- [x] VM: shim listener bound on `127.0.0.1:15201`; `curl http://127.0.0.1:15201/api/daemon/api/link/v1/callback/google-calendar?...` returns `302 → https://local.hellofriday.ai:15200/...`.
- [x] VM: Decoded OAuth `state.uri` Link emits is now `http://127.0.0.1:15201/...` — Cloud Function will accept.
- [ ] Owner click-through: complete a real Google OAuth in browser end-to-end (the manual JSON page should be gone).

### DIFF
--- a/apps/atlasd/routes/link.ts
+++ b/apps/atlasd/routes/link.ts
@@ -37,11 +37,23 @@ linkRoutes.use("/*", devOnlyMiddleware());
 // https. Without TLS env, fall back to http on loopback — the
 // pre-s2s-mesh behavior.
 const linkScheme = process.env.FRIDAY_TLS_CERT && process.env.FRIDAY_TLS_KEY ? "https" : "http";
-const LINK_SERVICE_URL =
-  process.env.LINK_SERVICE_URL ??
-  (process.env.FRIDAY_PORT_LINK
+// Auto-upgrade http:// → https:// when s2s is on — see the matching
+// comment in apps/atlasd/src/services/communicator-wiring.ts. The
+// installer wizard pins LINK_SERVICE_URL=http://localhost:13100 on
+// installs that predate TLS, so a literal env-var read here would
+// send cleartext into link's TLS listener and every chat tool call
+// routed through link surfaces as "Link service is unavailable".
+function upgradeLinkScheme(url: string): string {
+  if (linkScheme === "https" && url.startsWith("http://")) {
+    return "https://" + url.slice("http://".length);
+  }
+  return url;
+}
+const LINK_SERVICE_URL = process.env.LINK_SERVICE_URL
+  ? upgradeLinkScheme(process.env.LINK_SERVICE_URL)
+  : process.env.FRIDAY_PORT_LINK
     ? `${linkScheme}://localhost:${process.env.FRIDAY_PORT_LINK}`
-    : `${linkScheme}://localhost:3100`);
+    : `${linkScheme}://localhost:3100`;
 const PROXY_PREFIX = "/api/link";
 
 /**
--- a/apps/atlasd/routes/link.ts
+++ b/apps/atlasd/routes/link.ts
@@ -37,23 +37,11 @@ linkRoutes.use("/*", devOnlyMiddleware());
 // https. Without TLS env, fall back to http on loopback — the
 // pre-s2s-mesh behavior.
 const linkScheme = process.env.FRIDAY_TLS_CERT && process.env.FRIDAY_TLS_KEY ? "https" : "http";
-// Auto-upgrade http:// → https:// when s2s is on — see the matching
-// comment in apps/atlasd/src/services/communicator-wiring.ts. The
-// installer wizard pins LINK_SERVICE_URL=http://localhost:13100 on
-// installs that predate TLS, so a literal env-var read here would
-// send cleartext into link's TLS listener and every chat tool call
-// routed through link surfaces as "Link service is unavailable".
-function upgradeLinkScheme(url: string): string {
-  if (linkScheme === "https" && url.startsWith("http://")) {
-    return "https://" + url.slice("http://".length);
-  }
-  return url;
-}
-const LINK_SERVICE_URL = process.env.LINK_SERVICE_URL
-  ? upgradeLinkScheme(process.env.LINK_SERVICE_URL)
-  : process.env.FRIDAY_PORT_LINK
+const LINK_SERVICE_URL =
+  process.env.LINK_SERVICE_URL ??
+  (process.env.FRIDAY_PORT_LINK
     ? `${linkScheme}://localhost:${process.env.FRIDAY_PORT_LINK}`
-    : `${linkScheme}://localhost:3100`;
+    : `${linkScheme}://localhost:3100`);
 const PROXY_PREFIX = "/api/link";
 
 /**
--- a/apps/atlasd/routes/mcp-registry.ts
+++ b/apps/atlasd/routes/mcp-registry.ts
@@ -60,19 +60,12 @@ async function createLinkProvider(
   // the legacy :3100 default. Keeps desktop port-override installs
   // talking to the right link binary. Scheme follows the s2s mesh
   // (https when FRIDAY_TLS_CERT/_KEY set, http fallback otherwise).
-  // Stale http://localhost:13100 in installer-written .env is also
-  // auto-upgraded — same fix as routes/link.ts and the playground
-  // proxy (tools/agent-playground/static-server.ts).
   const linkScheme = process.env.FRIDAY_TLS_CERT && process.env.FRIDAY_TLS_KEY ? "https" : "http";
-  const upgradeLinkScheme = (url: string): string =>
-    linkScheme === "https" && url.startsWith("http://")
-      ? "https://" + url.slice("http://".length)
-      : url;
-  const linkServiceUrl = process.env.LINK_SERVICE_URL
-    ? upgradeLinkScheme(process.env.LINK_SERVICE_URL)
-    : process.env.FRIDAY_PORT_LINK
+  const linkServiceUrl =
+    process.env.LINK_SERVICE_URL ??
+    (process.env.FRIDAY_PORT_LINK
       ? `${linkScheme}://localhost:${process.env.FRIDAY_PORT_LINK}`
-      : `${linkScheme}://localhost:3100`;
+      : `${linkScheme}://localhost:3100`);
   const url = `${linkServiceUrl}/v1/providers`;
   const headers: Record<string, string> = { "Content-Type": "application/json" };
   const atlasKey = process.env.FRIDAY_KEY;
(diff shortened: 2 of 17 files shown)


ljagiello:
Code-review follow-ups pushed in 32e794795. All 8 findings addressed:

**Production fixes**
- **Restart-cap on the readiness runner** (Finding #1) — `restartsIssued` counter on `readinessRunner`, capped at `supervisedMaxRestarts`. After cap: keep probing (cache stays accurate if service recovers) but stop calling `sup.RestartProcess`. Process-compose's MaxRestarts only gates its in-band restart loop, confirmed against `ProjectRunner.doRestart` in the local checkout. Misleading comment replaced.
- **Bidirectional `.env` migration** (Finding #2) — `migrateStaleURLSchemes` now also handles `https→http` when `s2sCertsValid()==false`. Only flips launcher-managed loopback URLs (`localhost`/`127.0.0.1`); operator-pinned external `https://my-proxy.local:443` passes through. Triggered behavior matches `Downgrade .env to http://` option (confirmed with author).

**Test fixes**
- **#3** Tautological TLS-skip-verify test rewritten — now mutates `readinessTLSConfig.InsecureSkipVerify` (with `t.Cleanup` to restore) and routes through `newReadinessClient("https")` so a regression in the singleton wiring is actually caught.
- **#4** `Run()` + `startReadinessRunners` covered — atomic-counter handler, ctx-cancel deadline, fan-out verification.
- **#5** Production-order transition pinned — `Update` first (status=starting), `SetReady`, next `Update` → healthy + `transitionAt` reset.
- **#6** Tests assert via `Snapshot` (public API) instead of `cache.customReady` map peek. Dedicated `TestHealthCache_SetReady_NotifiesSubscribers` isolates the three branches (first-write silent, idempotent no-op, flip = exactly one notify).
- **#7** `.env` structure preservation: CRLF + LF tests assert byte-for-byte preservation of comments, blank lines, line endings, and quoted values.
- **#8** Migration fixture is now generated from `staleHTTPSchemeKeys` — adding a key automatically grows the assertion.

**Tests:** `go test ./tools/friday-launcher/...` (75+ cases), `deno task test apps/link/src/oauth/` (57 cases). All green.

**QA on the VM:** redeployed launcher + link, all 5 services healthy in ~70s, no restart loops.

basedfriday (commented):
## Code review

High-quality PR overall — excellent test coverage (failure thresholds, restart caps, CRLF preservation, TLS round-trips, goroutine-leak guards), comments explain *why*, and the shim has a tight security posture (fixed redirect target, GET-only, path-prefix gated — no open-redirect surface). Confirmed the launcher has no `DependsOn: process_healthy` wiring, so `ReadinessProbe: nil` doesn't break process-compose startup ordering.

A few things to address:

### 1. Stale/contradictory comment in `project.go` (should fix before merge)
The `probeInitialDelay` const-block comment says the timings are *"shared between process-compose's stub probe (which we keep wired to a never-fails endpoint ...) and our own native readinessRunner."* But `ReadinessProbe` is set to `nil` — there is no stub probe. The constants are consumed only by `readinessRunner`. The comment describes an approach that isn't in the code.

### 2. Shim listener has no error handling — playground crash-loop risk (`static-server.ts`)
```ts
Deno.serve({ port: SHIM_PORT, hostname: "127.0.0.1" }, (req) => { ... });
```
If `SHIM_PORT` (playground port + 1, not operator-chosen) is already bound on the host, `Deno.serve` throws unhandled and takes down the entire playground process — which the launcher then restarts, into a loop. The OAuth shim being unavailable degrades gracefully to the old manual-paste page; crashing the playground does not. Suggest try/catch + log-and-continue.

### 3. `tick` logs bad-URL error every probe period (`readiness.go`)
The comment on the `http.NewRequestWithContext` error path says *"Surface once via log"*, but since the URL is fixed for the runner's lifetime this `log.Error` fires every `probePeriodSeconds` forever. Low impact (programmer-error path) — either dedupe the log or fix the comment.

### Minor
- **`SetReady` transitionAt/status window**: on a readiness flip, `SetReady` bumps `transitionAt` immediately but leaves `status` stale until the next `Update` (~500ms). Snapshot consumers briefly see `SinceSecs=0` against the old status. Tests confirm this is intentional — just a subtle contract.
- **`startsWithLoopbackHost`** doesn't treat `::1`/`[::1]` as loopback. Almost certainly fine (installer writes `localhost`), but IPv6 loopback URLs would silently skip migration.
- **`.env` parsing** in `migrateStaleURLSchemes` requires `KEY=value` with no spaces around `=` and no `export ` prefix. Fine for installer-written files; a hand-edited `KEY = http://...` won't migrate. `0o644` matches existing `ensureCertEnvFile` behavior — no perms regression.
- Two restart mechanisms now coexist (process-compose crash-restart + `readinessRunner` probe-restart) with no coordination. The per-runner `restartMax` cap bounds runaway probe-restarts, so acceptable, but a service that both crash-loops and fails probes could see extra churn.

None blocking except your call on #2 (the one with real user impact). #1 is a quick comment fix worth landing before merge.

ljagiello:
@basedfriday Thanks for the review — all four items addressed in `10a816285`:

**1. Stale ReadinessProbe-stub comment in `project.go`** — rewrote the doc block. `ReadinessProbe` is intentionally `nil` (process-compose's HttpProbe has no TLS controls and would reject our private-CA s2s leaf), so the constants are consumed only by `readinessRunner`. Comment now reflects that.

**2. Shim listener crash-loop risk in `static-server.ts`** — wrapped `Deno.serve` in `try/catch` + `onError`. If `PLAYGROUND_OAUTH_SHIM_PORT` is already bound, we log the AddrInUse and continue without the shim — the main playground listener stays up and Google OAuth falls back to the Gemini manual JSON-paste page (the pre-shim status quo). Crash loop avoided.

**3. `tick` per-period bad-URL log spam in `readiness.go`** — `r.url` is fixed for the runner's lifetime, so the `http.NewRequestWithContext` error path was indeed firing every ~2s forever. Added a `badURLLogged` guard: log once, then go quiet. The cache `SetReady(false)` flip still fires every tick so the "starting forever" signal still surfaces in the UI even after we stop logging.

**4. IPv6 loopback in `startsWithLoopbackHost`** — added a bracketed-URL branch per RFC 3986. `https://[::1]:18080` and `https://[::1]` now flip alongside `localhost` / `127.0.0.1`. Bare unbracketed `::1` is not a valid URL host and stays unrecognized; IPv4-mapped IPv6 (`[::ffff:127.0.0.1]`) intentionally passes through.

Plus two new tests:
- `TestReadinessRunner_BadURLLoggedOnce` — pins the dedupe (flag flips on first tick, stays flipped for 20+ more, `consecutiveFail` still increments).
- `TestMigrateStaleURLSchemes_IPv6Loopback` — pins `[::1]` recognition and the non-flip of `[::ffff:127.0.0.1]`.

On the minor observations:
- **`SetReady` transitionAt/status window** — confirmed intentional. The 500ms `runHealthPoll` tick re-derives status; `SetReady` alone only bumps `transitionAt` so SSE consumers get the "something changed here" signal without waiting for the next derive.
- **`.env` parsing strictness** (no spaces around `=`, no `export `) — fine for installer-written files; left as-is.
- **Two restart mechanisms coexisting** — yes, the per-runner `restartMax` cap is what bounds it. A crash-looping service plus probe-restart can see extra churn within the launcher's lifetime budget but stays well-bounded.

`go test ./tools/friday-launcher/...` + `golangci-lint run` both clean.

PR #321, opened 2026-05-14
Workspace chat now reflects its own config changes mid-session

basedfriday:
## Summary

- A workspace job or signal you create or edit in chat now shows up on the next message instead of being stale for up to 5 minutes — the chat's workspace cache is dropped on every `upsert_*` / `publish_draft`.
- New `trigger_signal` tool: run a job you just created in the *same* chat turn. Before this, a freshly-created job had no callable tool until the next session, so the agent would tell you to send another message.
- Triggering a signal over HTTP with the wrong body shape now returns a clear 400 naming the `{"payload": {...}}` envelope, instead of a misleading "expected X, received undefined" error that pointed at the signal's own schema.
- `load_skill` results are no longer lifted out to artifacts. Skill bodies over the scrubber's size threshold were being replaced with a "[lifted to artifact]" marker, so a "loaded" skill was silently never actually read — the agent proceeded without it. Covers delegate sub-agent calls too.
- The volatile workspace inventory moved to its own prompt-cache breakpoint, so editing a workspace no longer burns a full-rate cache write on the otherwise-stable identity/skills tier. The "force fresh" cache salt stays on the earlier breakpoint so it still busts everything.
- Two debugging skills updated: a "run a control before blaming your own change" triage rule, and corrected guidance for invoking a job created in the current session.

## Test plan

- [ ] In a chat: `upsert_job`, then run it the same turn via `trigger_signal` — no "send another message"
- [ ] In a chat: `upsert_signal`, confirm the change is visible without starting a new chat
- [ ] `curl` a signal endpoint with a bare body -> clear envelope error; with `{"payload": {...}}` -> works
- [ ] `load_skill` on a large skill -> the body lands in the prompt, not a marker
- [ ] typecheck + knip + full test suite (green locally)

## Not yet run

The promptfoo eval suite hasn't been run on this branch yet — opening for review with typecheck / knip / unit tests green. Evals to follow.

### DIFF
--- a/apps/atlasd/routes/workspaces/index.test.ts
+++ b/apps/atlasd/routes/workspaces/index.test.ts
@@ -175,6 +175,24 @@ describe("POST /workspaces/:workspaceId/signals/:signalId envelope guard", () =>
     expect(triggerWorkspaceSignal).not.toHaveBeenCalled();
   });
 
+  test("rejects a bare payload on the SSE handler too (Accept: text/event-stream)", async () => {
+    // The SSE handler is a separate `.post` registration dispatched by the
+    // accept header — it reads rawBody directly, so its guard branch needs
+    // its own coverage.
+    const { app, triggerWorkspaceSignal } = createTestApp();
+    const res = await post(
+      app,
+      "/workspaces/ws-1/signals/sig-1",
+      { input: "hello" },
+      { Accept: "text/event-stream" },
+    );
+
+    expect(res.status).toBe(400);
+    const body = (await res.json()) as { error: string };
+    expect(body.error).toContain('{"payload"');
+    expect(triggerWorkspaceSignal).not.toHaveBeenCalled();
+  });
+
   test("allows a properly enveloped body through the guard", async () => {
     const previous = process.env[INTERNAL_SIGNAL_BYPASS_TOKEN_ENV];
     process.env[INTERNAL_SIGNAL_BYPASS_TOKEN_ENV] = "test-token";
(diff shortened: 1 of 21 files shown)


Vpr99:
One gap on the cache-invalidation invariant: `enable_mcp_server` and `disable_mcp_server` mutate `config.tools.mcp.servers` but don't call `invalidateBlock2`, so the `<mcp_servers>` block (rendered from `Object.keys(config?.tools?.mcp?.servers)` at `workspace-chat.agent.ts:306`) stays stale for up to 5 min after the user enables/disables a server mid-chat — same failure class this PR is fixing for `upsert_*`.

Two lines:

- `packages/system/agents/workspace-chat/tools/enable-mcp-server.ts` — on the 200 success branch after the `$put`, `invalidateBlock2(effectiveWorkspaceId)`.
- `packages/system/agents/workspace-chat/tools/disable-mcp-server.ts` — same, after the `$delete`.

`install_mcp_server` / `create_mcp_server` are fine (they hit `mcpRegistry`, not workspace config), and `connect_communicator` is just a stop-condition trigger.

(Otherwise LGTM)

basedfriday:
Good catch — fixed in 8c16186b. `enable_mcp_server` and `disable_mcp_server` now call `invalidateBlock2(effectiveWorkspaceId)` on their 200 success branch, scoped to the effective target workspace (so the cross-session `workspaceId` param is respected). Tests in both tool specs assert it fires on success and stays unfired on the failure paths (404 / 409 conflict).

Agreed on the rest — `install_mcp_server` / `create_mcp_server` hit the registry, not workspace config, and `connect_communicator` is a stop-condition trigger, so none of those need it.

PR #323, opened 2026-05-14
Workspace env supply + MCP registry doctor + catalog detail redesign

basedfriday:
## Summary

- Agents can now read and write workspace env vars from chat. `env_set` surfaces a confirmation card in the conversation — nothing is written until you approve. Use `env_get` on the next turn to verify.
- A workspace no longer bricks when an MCP server has an unresolved env var. The daemon drops just the broken server and keeps going; the missing vars show up in workspace metadata so you know what to set.
- Workspace settings page is live: edit workspace identity, set per-agent env vars, and manage MCP servers (enable, configure, connect credentials) in one place.
- MCP catalog detail page redesigned — section-routed per-server pages (Overview, Connections, Config Reference, Testing, Readme) with a raw tool invoker for ad-hoc testing.
- `~/` and `$HOME` now expand in file tool path args so agents can reference home-relative paths naturally.

## Test plan

- [ ] Set a workspace env var from chat, confirm the card, verify with `env_get` on the next turn
- [ ] Enable an MCP server with a missing required env var — workspace should come up with the server listed as degraded, not fail entirely
- [ ] Workspace settings: edit identity, set agent env vars, enable an MCP server, configure settings, connect credentials
- [ ] `/mcp` catalog: drill into a server, navigate each section, invoke a tool directly from the testing panel
- [ ] `deno task typecheck`, knip, full test suite — all green ✓

## Open issues

- [#330](https://github.com/friday-platform/friday-studio/issues/330) — env read/delete tools work in dev mode (auto-mint); a non-dev install would need facade-direct calls instead of bare HTTP.


Vpr99 (commented):
Code review pass — flagging critical + security-sensitive findings inline. Full review (verdict: **Needs Work**) at `docs/reviews/2026-05-14-workspace-env-mcp-doctor.md` on this branch; includes test-quality and polish findings I'm not posting inline.

---

### Critical — `/test-chat` still has the workspace-less auth gap that `a67515b` closed on `/invoke`

Line 1090 isn't in the PR diff (the `/test-chat` route predates this branch) so I can't anchor inline, but it deserves the same fix as `/invoke`:

`apps/atlasd/routes/mcp-registry.ts:1090–1099` —
```ts
zValidator("query", z.object({ workspaceId: z.string().optional() })),
...
if (workspaceId) {
  await requireWorkspaceMember(c, workspaceId);
}
```

A workspace-less `POST /api/mcp-registry/:id/test-chat` runs `streamText` over `server.configTemplate` with the daemon's ambient `process.env` resolving `from_environment` / `auto` vars — identical attack surface to what `/invoke` was hardened against, plus it burns platform-model tokens per call. Mirror the `/invoke` fix: require `workspaceId.min(1)` and drop the `if (workspaceId)` branch.

apps/atlasd/routes/workspaces/env.ts
```diff
@@ -0,0 +1,144 @@
...
+async function resolveWorkspaceEnvPath(
+  c: Context<AppVariables>,
+  workspaceId: string,
+): Promise<string | null> {
+  const workspace = await c.get("app").getWorkspaceManager().find({ id: workspaceId });
+  if (!workspace) return null;
+  return join(workspace.path, ".env");
+}
+
+const workspaceEnvRoutes = daemonFactory
```
Vpr99, line 57: **Critical — no `env.test.ts` sibling.**

Brand-new HTTP surface (GET list / GET key / PUT / DELETE) with no route-level coverage. The authz boundary (`requireWorkspaceMember`), the POSIX-key regex at line 36, the no-newline value regex at line 41, and 404-on-missing-workspace are all unverified at the route layer. Every other workspace sibling route has a test file.

Add `env.test.ts` mirroring `apps/atlasd/routes/workspaces/mcp.test.ts`'s structure. The unit tests in `packages/workspace/src/workspace-env.test.ts` cover the file editor, not the HTTP layer.

apps/atlasd/routes/workspaces/env.ts
```diff
@@ -0,0 +1,144 @@
...
+      const workspaceId = c.req.param("workspaceId");
+      const { key } = c.req.valid("param");
+      const { value } = c.req.valid("json");
+      if (!workspaceId) {
+        return c.json(
+          { success: false, error: "bad_request", message: "Missing workspaceId" },
+          400,
+        );
+      }
+      await requireWorkspaceMember(c, workspaceId);
```
Vpr99, line 105: **Member-level authz on mutations is inconsistent with siblings.**

`mcp.ts:183/301/399` and `config.ts:738–743` require `requireWorkspaceAdmin` for writes. The workspace `.env` now feeds MCP-server `from_environment` resolution and per-agent `env:` wiring — a workspace member who isn't an admin can rewrite `LITELLM_API_KEY` or any MCP server's plain-config values via this route.

Switch PUT/DELETE to `requireWorkspaceAdmin`. Cheapest change: lift the gate into a `.use("*", ...)` block the way `config.ts` does. Same comment applies to the DELETE handler at line 127.

apps/atlasd/routes/elicitations/index.ts
```diff
@@ -211,6 +279,23 @@ elicitationApp.post(
       // require the caller to be a member of that workspace.
       await requireWorkspaceMember(c, got.data.workspaceId);
```
Vpr99, line 287: **`scope: "global"` is reachable from any workspace's member.**

Authz here is against the elicitation's own `workspaceId`, but the commit dispatches on `pendingTool.args.scope` (`env-commit.ts:71–72` → `commitGlobalEnvWrite`). In a multi-tenant install, a workspace member who isn't a daemon admin can clobber instance-wide keys via the chat-confirm flow — agent proposes `LITELLM_API_KEY` global, user confirms, daemon's LLM access dies.

Either require a global-admin check when `parsed.data.scope === "global"`, or refuse `scope === "global"` here and force global-env writes through `routes/config.ts` (which has its own gate). Single-user installs are fine; multi-tenant is not.

packages/mcp-server/src/tools/env/set.ts
```diff
@@ -0,0 +1,143 @@
...
+        );
+      }
+
+      // Reads of secret-looking keys come back masked as `********`. An agent
+      // doing read-modify-write would otherwise propose writing the mask
+      // sentinel as the real value — refuse it rather than corrupt the `.env`.
+      const maskedKeys = keys.filter((k) => vars[k] === MASKED_VALUE);
+      if (maskedKeys.length > 0) {
+        return createErrorResponse(
+          `Refusing to write the masked sentinel (${MASKED_VALUE}) as a value for ` +
```
Vpr99, line 84: **Masked-sentinel refusal is good; missing denylist for daemon-internal keys.**

The tool accepts any key. An agent can propose `LITELLM_API_KEY` for global scope — the confirmation card looks reasonable, the user accepts, daemon LLM access breaks. Similarly, `HOT_RELOAD_DENYLIST` keys appear to apply but don't take effect at runtime, so the user thinks they edited a config setting that's actually inert.

Refuse the `HOT_RELOAD_DENYLIST` set with a clear error pointing at the daemon admin surface, and consider an explicit denylist for `LITELLM_*` / `FRIDAY_*` — they're daemon-internal, not workspace settings.

packages/mcp-server/src/tools/env/get.ts
```diff
@@ -0,0 +1,66 @@
...
+      },
+    },
+    async ({ scope, workspaceId, key }): Promise<CallToolResult> => {
+      ctx.logger.info("MCP env_get called", { scope, workspaceId, key });
+
+      const route = envRouteBase(ctx.daemonUrl, scope, workspaceId);
+      if (!route.ok) return createErrorResponse(route.error);
+
+      try {
+        const res = await fetch(`${route.base}/${encodeURIComponent(key)}`);
```
Vpr99, line 37: **Fetch with no `Authorization` header — 401s in production.**

The daemon's `/api/*` middleware requires either the `friday_session` cookie or a Bearer token and auto-mints **only when `FRIDAY_ENV === "dev"`**. In production / non-dev installs, every `env_get` / `env_list` / `env_delete` will 401. Same issue at `list.ts:33` and `delete.ts:38`.

Thread `FRIDAY_KEY` through `ToolContext` and add a Bearer header (see `getLinkAuthHeaders()`). Better: move the env tools to call into the workspace facade directly — same process, no HTTP hop, no auth dance. The HTTP-hop pattern was inherited from memory tools and is a pre-existing wart; extending it deepens the bug.

apps/atlasd/routes/mcp-registry.ts
```diff
@@ -183,6 +191,209 @@ function createUpstreamClient(): MCPUpstreamClient {
...
+    )
+    .min(1),
+});
+
+/** Body for `POST /manual-config/:id` — credentials vs. plain settings, split by the user. */
+const ManualConfigSchema = z.object({
+  credentials: z
+    .array(
+      z.object({
+        name: z.string().min(1),
```
Vpr99: **`name: z.string().min(1)` accepts non-POSIX keys.**

`"FOO BAR"` and `"1FOO"` both pass. Downstream writers (`workspaces/env.ts:36`, `workspace-env.ts` line editor) expect `/^[A-Za-z_][A-Za-z0-9_]*$/`. The manual-config path isn't currently invoking `setEnvFileVar` directly, so the immediate blast radius is bounded — but bad keys land in the registry KV and in Link's `secretSchema`, and downstream consumers may not be as tolerant.

Tighten both schemas (this line and `settings[i].name` at line 265) to the same POSIX-identifier regex used in `env.ts:36`. Cheap consistency win.

Vpr99 (approved):

PR #334, opened 2026-05-15
Simplify chat export to remove HTML page

Vpr99:
## Summary

- The chat-export Export button worked in dev but 500'd in the compiled binary. The orchestrator lived in a SvelteKit `+server.ts` and the in-zip viewer in a `+page.server.ts` preview page — `adapter-static` strips both at build time, so requests fell through to the SPA fallback and the frontend tried to parse `index.html` as a ZIP. Dev only worked because Vite ran the SvelteKit server handlers in-process.
- Moved the export to a Hono route at `/api/export/:wsId/:chatId` (lives in `lib/server/routes/export.ts`, mounted by both the dev catch-all and `static-server.ts`). The ZIP now contains `chat.json` (userId stripped, home paths scrubbed) + `assets/artifacts/{id}/{slug}` — no HTML viewer; recipients open `chat.json` in whatever they like. This drops -1923 LOC because the whole preview-rendering pipeline (and its `getExportContext` plumbing across 4 chat components + tests) goes with it.
- Limits and slug helpers are preserved: 25 MB/artifact, 250 MB aggregate, the same `artifactZipPath` slug helper. 404 still 404s, 502 on daemon failure, 499 on client abort.

## Test plan

- [x] `deno task check` — 0 errors across 8400 files
- [x] `deno task test` — 482 passed (2 pre-existing unrelated failures: workspace manager + bundled gh-agent)
- [x] Smoke test against running dev daemon: `curl /api/export/user/<chat>` → 200, valid ZIP with `chat.json` (userId stripped, 2 messages, 3 top-level keys)
- [x] 404 path: `curl /api/export/user/<missing>` → `{"error":"Chat not found"}`
- [ ] Compiled binary smoke test — needs a rebuild of the `playground` binary to verify the deploy-time path; the dev path proves the same code works


ljagiello (commented):
(disregard — line-resolution probe while figuring out the API; the real review is below)

tools/agent-playground/src/lib/server/routes/export.ts
```diff
@@ -0,0 +1,312 @@
...
+    }
+  }
+
+  if (signal.aborted) return new Response(null, { status: 499 });
+
+  const zip = new JSZip();
+  // Strip the account-ownership id. Transcript, system prompt context, and
+  // tool I/O are exported verbatim — recipients see what the sender saw,
+  // minus the username paths scrubHomePaths catches downstream.
+  const { userId: _userId, ...chatWithoutUserId } = chatParsed.data.chat;
```
ljagiello, line 344: **Critical — forward-compat PII risk.** `GetChatResponseSchema` uses `.passthrough()`, so blacklisting only `userId` means any future daemon-side column (`ownerEmail`, `accountId`, `sessionToken`, `apiKeyId`, …) silently lands in shared `chat.json` exports the moment it's added daemon-side. The current code is correct *today* — but the threat model is "future columns shouldn't leak by being added," and blacklists don't satisfy that.

Recommend inverting to a whitelist:
```ts
const { id, workspaceId, source, title, createdAt, updatedAt } = chatParsed.data.chat;
const chatForExport = { id, workspaceId, source, title, createdAt, updatedAt };
```
Safe-by-default; a new sensitive column doesn't leak just by existing. (Exact field list is a product call — confirm against whichever columns ship in `GetChatResponseSchema` today.)

tools/agent-playground/src/lib/server/routes/export.ts
```diff
@@ -0,0 +1,312 @@
...
+  const wsPath = encodeURIComponent(workspaceId);
+  const chatPath = encodeURIComponent(chatId);
+  const daemonBaseUrl = effectiveDaemonUrl().replace(/\/+$/, "");
+
+  const chatUrl = `${daemonBaseUrl}/api/workspaces/${wsPath}/chat/${chatPath}?full=true`;
+  const artifactsUrl = `${daemonBaseUrl}/api/artifacts?chatId=${chatPath}`;
+
+  let chatRes: Response;
+  try {
+    chatRes = await fetch(chatUrl, { signal });
```
ljagiello: **Important — lost 10s upstream timeout (regression from the old route).** The previous `+server.ts` raced the pipeline against `EXPORT_TIMEOUT_MS = 10_000`. The new route has no timeout on any daemon fetch — a hung daemon pins the request indefinitely. The artifact byte loop downstream multiplies the exposure.

```ts
const signal = AbortSignal.any([c.req.raw.signal, AbortSignal.timeout(10_000)]);
```
Map a timeout to 504 (distinct from 499 client abort) so it shows up clearly in logs.

tools/agent-playground/src/lib/server/routes/export.ts
```diff
@@ -0,0 +1,312 @@
...
+    2,
+  );
+  zip.file("chat.json", scrubHomePaths(chatJson));
+  byteResults.forEach((result, idx) => {
+    if (result.status === "fulfilled") {
+      zip.file(result.value.path, result.value.bytes);
+    } else {
+      const summary = fetchableArtifacts[idx];
+      const reason = result.reason instanceof Error ? result.reason.message : String(result.reason);
+      console.warn(
```
ljagiello, line 361: **Important — silent drops, no recipient signal.** With the HTML viewer gone there's no broken-link hint — the zip just contains a smaller-than-expected `assets/artifacts/` and `chat.json` references ids that don't exist on disk. Recipient doesn't know whether the export is complete.

Recommend an `export-manifest.json` (or a top-level `errors[]` in `chat.json`) listing dropped ids + reasons. Cheap, makes the export self-describing, replaces the affordance the viewer used to carry. Same `console.warn` text can feed it.

ljagiello (commented):
## Review — credit for the harden pass, two open

Nice work on `a65dad36f` — the streaming `readResponseBytesWithLimit`, declared-size pre-check, `unsafeArtifactIdPathSegment` guard, JSON parse error paths, plus the new `export.test.ts` (10 cases) and `artifact-zip-path.test.ts` cover the main correctness/test gaps from my first read. `scrubHomePaths` is exercised indirectly through the integration test's `/Users/alice` → `/Users/~` assertion, which is fine.

**Two findings still open, both small:**

1. **`userId`-only strip on a `.passthrough()` schema is a forward-compat PII risk** *(Critical).* `GetChatResponseSchema` still uses `.passthrough()`, so any future daemon-side column (`ownerEmail`, `accountId`, `sessionToken`, …) will silently flow into shared `chat.json` exports the moment it's added. Whitelist the fields you want exported instead of blacklisting `userId`. Inline comment with a suggested shape.

2. **No upstream timeout on daemon fetches** *(Important — regression from the old route's `EXPORT_TIMEOUT_MS = 10_000`).* A hung daemon now pins the request indefinitely. `AbortSignal.any([c.req.raw.signal, AbortSignal.timeout(10_000)])` covers it; map a timeout to 504 distinct from 499.

**Two smaller follow-ups:**

3. *Silent dropped artifacts* still have no recipient signal — with the HTML viewer gone there's not even a broken-link hint. An `export-manifest.json` (or top-level `errors[]` in `chat.json`) with dropped ids + reasons is cheap and replaces the affordance the viewer used to carry. Inline.

4. *Orphan import* — `chat-message-list.svelte:11` still has `import { formatMessageTimestamp } from "@atlas/core/chat/export/render";` with zero remaining call sites after the `{#if isExport}` branch was removed. Drop it. Separately worth checking the no-unused-imports lint config — CI should have caught this. (Not inline since the line isn't in the diff hunk.)

**Process note (non-blocking):** the compiled-binary smoke-test box in the PR description is still unchecked. That's the whole reason this PR exists — without it, "fixes the bug" stays unverified. Tick it before merge or wire up a static-server assertion that `/api/export/*` is in the compiled bundle.

Leaving as comments, not change requests.

ljagiello (approved):

PR #335, opened 2026-05-15
fix(credential-resolver): env-var fallback before declaring provider missing

ljagiello:
## Summary

A user uploading a workspace.yml via Studio's "Add space → Upload File" gets blocked with \`error: \"missing_providers\", providers: [\"hubspot\"]\` even when \`HUBSPOT_ACCESS_TOKEN\` is set in \`~/.friday/local/.env\`. The bundled hubspot agent reads from env at runtime, so once you work around the validator the job runs fine — the validator was the only thing rejecting the upload.

This is the daemon-side fix for the chain documented in [bucketlist-cs#7](https://github.com/friday-platform/bucketlist-cs/pull/7) discovery + the QA round on the fresh-installer VM.

## The chain

1. \`POST /api/workspaces/create\` calls \`injectBundledAgentRefs\` (\`apps/atlasd/routes/workspaces/inject-bundled-agents.ts:8\`) — auto-injects a Link credential ref env var on every \`type: atlas\` agent declared in the workspace, based on the bundled-agents registry's \`requiredConfig\` field
2. For hubspot the registry declares \`requiredConfig: [{ from: 'link', envKey: 'HUBSPOT_ACCESS_TOKEN', provider: 'hubspot', key: 'access_token' }]\`, so \`env.HUBSPOT_ACCESS_TOKEN: { from: 'link', provider: 'hubspot', key: 'access_token' }\` gets injected
3. \`extractCredentials\` finds the injected ref → \`uniqueProviders = ['hubspot']\`
4. \`resolveCredentialsByProvider('hubspot')\` calls Link's \`/v1/summary?provider=hubspot\` → in dev mode (and apparently in some prod configs too), Link doesn't know about hubspot → returns empty \`providers\` list
5. \`isKnownProvider\` is false → throws \`InvalidProviderError\`
6. Call site at \`apps/atlasd/routes/workspaces/index.ts:557-562\` collects the provider into \`invalidProviders\` → returns 400 \`missing_providers\`

## Fix

In \`resolveCredentialsByProvider\`, before throwing \`InvalidProviderError\` or \`CredentialNotFoundError\`, look up the env-var name the bundled atlas agents register for this provider via \`bundledAgentsRegistry\`, then check \`process.env\`. If the var is set, return a synthetic \`CredentialSummary { id: 'env:<KEY>', type: 'env', isDefault: true, ... }\`.

Why \`id: 'env:...'\` prefix: lets downstream callers short-circuit Link refresh logic — you can't refresh an env-supplied credential.

## Why not fix \`injectBundledAgentRefs\` instead?

Two cleaner reasons to fix at the resolver layer rather than the injector:

- The resolver is the single chokepoint where "is this provider configured?" actually gets decided. Skipping injection would mean the workspace's runtime config no longer carries the credential ref at all, which removes the declarative requirement that the setup page uses to render Connect buttons.
- Other call sites (MCP server requirement validators, etc.) ask the same "configured?" question. Fixing the resolver fixes them all in one place.

## Tests

Added 4 new cases to \`credential-resolver.test.ts\` under a new \`env-var fallback\` describe block:

- env-var present, would-have-been InvalidProviderError → returns synthetic
- env-var present, would-have-been CredentialNotFoundError → returns synthetic
- env-var absent → still throws InvalidProviderError (no false-positive)
- unknown provider with no bundled-agent declaration → no fallback (random env vars don't get treated as credentials)

317 / 317 in \`packages/core/src/mcp-registry/credential-resolver.test.ts\`. Type-check clean.

## Test plan

- [ ] On a fresh installer setup with \`HUBSPOT_ACCESS_TOKEN\` in \`~/.friday/local/.env\` but NO Settings → Connections HubSpot configured: \"Add space → Upload File → bucketlist-cs/workspace.yml\" succeeds (no \`missing_providers\` block)
- [ ] When \`HUBSPOT_ACCESS_TOKEN\` is unset: same upload still blocks with \`missing_providers: [\"hubspot\"]\` (no false-positives)
- [ ] When a real Link credential exists for hubspot: that one wins (env-var fallback only kicks in when Link returns zero credentials)

### DIFF
--- a/packages/core/src/mcp-registry/credential-resolver.test.ts
+++ b/packages/core/src/mcp-registry/credential-resolver.test.ts
@@ -219,6 +219,85 @@ describe("resolveCredentialsByProvider", () => {
     expect((error as Error).message).toContain("not a registered provider");
   });
 
+  // Env-var fallback: a bundled atlas agent declares a Link credential
+  // requirement (e.g. hubspot's HUBSPOT_ACCESS_TOKEN), the user has set
+  // that env var in `~/.friday/local/.env` instead of configuring a Link
+  // credential via Settings → Connections, the runtime works because the
+  // bundled agent reads from env directly. The validator path used to
+  // reject the workspace upload with `missing_providers: ["hubspot"]`
+  // even though everything works at runtime — the fallback below makes
+  // resolveCredentialsByProvider accept the env var as a configured
+  // credential and synthesize a CredentialSummary for it.
+  describe("env-var fallback (~/.friday/local/.env)", () => {
+    const ENV_KEY = "HUBSPOT_ACCESS_TOKEN";
+    let originalValue: string | undefined;
+
+    beforeEach(() => {
+      originalValue = process.env[ENV_KEY];
+    });
+
+    afterEach(() => {
+      if (originalValue === undefined) delete process.env[ENV_KEY];
+      else process.env[ENV_KEY] = originalValue;
+    });
+
+    it("returns a synthetic 'env:' credential when InvalidProviderError would have fired but env is set", async () => {
+      // Link service has no record of hubspot AND no credentials —
+      // would normally throw InvalidProviderError ("not a registered
+      // provider"). With env-var fallback + HUBSPOT_ACCESS_TOKEN set,
+      // we synthesize a CredentialSummary instead.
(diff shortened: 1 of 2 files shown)


ljagiello:
Pivoting to a different approach per discussion — skipping the missing_providers validation entirely instead of fallback. Will open a fresh PR.

PR #336, opened 2026-05-15
fix(/api/workspaces/create): drop missing_providers hard-fail

ljagiello:
## Summary

Workspace upload via Studio's "Add space → Upload File" was returning `400 missing_providers: ["hubspot"]` (or similar) on a fresh installer setup, even when the user had `HUBSPOT_ACCESS_TOKEN` set in `~/.friday/local/.env`. The bundled atlas agent reads from `process.env` at runtime — the workspace ran fine the moment the validator was bypassed. The validator was the only thing rejecting the upload.

This supersedes the closed PR #335 (env-var fallback in `resolveCredentialsByProvider`); we agreed the cleaner fix is to skip the missing-providers hard-fail at the import layer rather than synthesize fake credentials at the resolver layer.

## Change

`apps/atlasd/routes/workspaces/index.ts` — in the `/create` handler, treat `InvalidProviderError` the same as `CredentialNotFoundError`:

```diff
-if (invalidProviders.length > 0) {
-  return c.json({ error: "missing_providers", ... }, 400);
-}
-if (unresolvedProviders.length > 0) {
-  unresolvedCredentialPaths = providerOnlyRefs
-    .filter((ref) => unresolvedProviders.includes(ref.provider))
-    .map((ref) => ref.path);
-}
+const allUnresolvedProviders = [...unresolvedProviders, ...invalidProviders];
+if (allUnresolvedProviders.length > 0) {
+  unresolvedCredentialPaths = providerOnlyRefs
+    .filter((ref) => allUnresolvedProviders.includes(ref.provider))
+    .map((ref) => ref.path);
+}
```

The workspace imports successfully with `requires_setup: true`. The per-workspace integrations sidebar can still render Connect buttons for Link credentials. If the agent genuinely needs a credential at runtime and neither env nor Link have one, the runtime surfaces a clear error there — the import path is the wrong place for that check.

## Test

Added `creates workspace with requires_setup when providers are unknown to Link (InvalidProviderError)` to `apps/atlasd/routes/workspaces/import-credentials.test.ts`. Mirrors the existing `CredentialNotFoundError` coverage but uses `InvalidProviderError` to exercise the new code path. Asserts:

- response is 201 (was 400 before)
- workspace metadata gets `requires_setup: true`
- response body's `unresolvedCredentials` includes the provider path
- written config preserves the provider-only ref so the setup page can still render a Connect button

## Test plan

- [ ] Fresh installer with `HUBSPOT_ACCESS_TOKEN` set in `~/.friday/local/.env` but no Link credential: upload `bucketlist-cs/workspace.yml` via Studio's drop zone → succeeds
- [ ] Workspace's `Reindex Knowledge Corpus` and `Answer HubSpot Ticket` jobs run end-to-end (env var picked up at runtime)
- [ ] When neither env nor Link have a credential and the agent actually runs: get a clear runtime error, not a silent skip

### DIFF
--- a/apps/atlasd/routes/workspaces/import-credentials.test.ts
+++ b/apps/atlasd/routes/workspaces/import-credentials.test.ts
@@ -1,6 +1,7 @@
 import {
   CredentialNotFoundError,
   type fetchLinkCredential,
+  InvalidProviderError,
   LinkCredentialExpiredError,
   LinkCredentialNotFoundError,
   type resolveCredentialsByProvider,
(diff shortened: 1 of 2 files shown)


ljagiello:
Pivoting again — user is going to declare HUBSPOT_ACCESS_TOKEN with a from-environment ref in workspace.yml directly, no daemon change needed.

PR #340, opened 2026-05-15
Daemon: plug four FD/memory leaks under the RLIMIT_NOFILE ceiling

ljagiello:
Follow-up to #332 (the launcher RLIMIT_NOFILE bump). That PR raised the
ceiling from 256/10240 to 65536; this PR fixes the leaks *underneath* so
we're not just postponing the next EMFILE.

## What's in this PR

Each leak was measured with a one-off microbench (results below) before
the fix landed. The bench scripts themselves are dropped from this PR
per reviewer request — numbers are reproducible from the patterns alone.

### #1 — HTTP fetch body undrained on `!response.ok`  *(highest-impact)*

**Bench (2k reqs @ concurrency 64, 1 MiB bodies):**
- BROKEN peak: **341 TCP sockets**
- FIXED peak:  **211 TCP sockets**
- **Δ: −130 sockets (−38%)** per burst

Every site that drops a Response without consuming the body holds its
socket in hyper's "borrowed" state. Under concurrency the pool grows
unbounded.

**Fix:** new `discardBody()` helper in `@atlas/utils`, threaded through
17 production files (~45 fix sites): compose-blocks, model-catalog,
credential-fetcher (with retry — each retry was leaking), upstream-client,
webfetch-handler, memory/env/workspace-chat tool suites, bundled-agent
HTTP paths, skills-sh-client, and the MCP readiness-probe loop. Sites
that already drain via `.text()/.json()/readResponseBody` or operate on
Result-typed wrappers (Hono `parseResult`, `@atlas/utils` Result) were
left alone.

### #2 — `agents/register` subprocess: missing `on("exit")` + unconsumed stdout

**Bench (200 spawns):**
- BROKEN peak: **17 pipe FDs**
- FIXED peak:  **12 pipe FDs**

The route spawned with `stdio:"pipe"` and SIGTERM'd as fire-and-forget.
Node held the ChildProcess (with its 3 pipe FDs) until GC visited it.
Worse, stdout was never read — any agent that prints would block on write
past ~64 KiB.

**Fix:** `stdio:["ignore","ignore","pipe"]`, `proc.once("exit")` awaited
in `finally`, SIGTERM→SIGKILL escalation after 2 s. Mirrors the lifecycle
`ProcessAgentExecutor` already had.

### #3 — `createHttpFetchHandler`: ReadableStream reader leaked on throw

**Bench: no measurable delta on a quiet system.** Honest call-out — Deno's
GC reclaims dropped Response bodies aggressively when v8 isn't under
pressure. The fix matters under memory pressure where v8 GC stalls past
hyper's idle reclaim window.

**Fix:** `try/finally` around the read loop in `agent-executor-utils.ts`
with `reader.cancel()` on every exit path (clean EOF, size-overflow throw,
mid-body network error). No regression risk: cancel is a no-op on the
clean-EOF path.

### #4 — Shared MCP child stderr accumulator unbounded

**Bench (5 s of stderr output):**
- BROKEN accumulator: **31.3 MiB held** / ΔRSS **+59 MiB**
- FIXED accumulator:  **0.81 MiB held** / ΔRSS **+34 MiB**
- **Δ: −97% accumulator, −42% RSS**

`process-registry.ts` attached a stderr `on("data")` handler that appended
every byte to a string for the diagnostic envelope of the
"spawn exited before ready" error — and never removed it. A chatty
uvicorn child (workspace-mcp logs every request) accumulated forever.

**Fix:** keep the handler attached for the child's whole lifetime (so the
kernel pipe buffer never back-pressures the child) but flip an
`accumulating` flag to false once the readiness probe passes. Bytes after
that point are drained and dropped.

## Test plan

- [x] `deno check` on every modified file — clean
- [x] 215 vitest tests across 14 test files in touched packages — all green (process-registry, utils, agent-context, credential-fetcher, upstream-client, model-catalog, skills, skills-sh-client)
- [ ] Manual: run the daemon for ≥48 h with a chatty MCP and confirm RSS + lsof FD count don't grow past steady-state

## Files

- `packages/utils/src/fetch-drain.ts` (new — `discardBody` helper)
- `packages/utils/mod.ts` (re-export)
- 18 production .ts files patched to call `discardBody` on early-return / throw paths (including readiness-probe + skills-sh-client added post-review)
- `apps/atlasd/routes/agents/register.ts` (subprocess lifecycle fix)
- `apps/atlasd/routes/agents/register.test.ts` (mock updated for new lifecycle)
- `packages/workspace/src/agent-executor-utils.ts` (reader try/finally)
- `packages/mcp/src/process-registry.ts` (stderr accumulator bounded at readiness + readiness-probe body drain)


packages/mcp/src/process-registry.ts
```diff
@@ -385,10 +385,30 @@ class ProcessRegistry {
     // Capture stderr for error messages only — never pattern-matched. uvicorn
     // fails silently on bind errors so the buffer is unreliable for control
     // flow.
+    //
+    // Three constraints driving the shape below:
+    //   1. We need stderr text for the "spawn exited before ready" error
+    //      thrown at line 411 — only useful BEFORE readiness, never after.
```
basedfriday: Stale line ref already at write time — this comment is part of a hunk that inserts +20 lines above the throw, so the `throw new MCPStartupError(...)` actually lands at line 435 in the new file, not 411. Either drop the line number or write "thrown in the readiness loop below" so it can't rot further.

basedfriday (commented):
Solid PR — four real leak classes, accurately diagnosed, fixed at the right layer with reproducible bench scripts. Three findings; the substantive ones are in inline threads, plus one finding on a file outside the diff that has to live here:

**Missed sites in `packages/skills/src/skills-sh-client.ts` (not in this diff):**
- L132 — `if (!response.ok) throw new Error(...)` with no drain. Not Result-typed (raw `fetchFn`).
- L166 — same shape on the download path.

Both need `await discardBody(response)` before throwing. Low traffic, but matches exactly the rule this PR codifies.

**Bench numbering:** files are `leak{1,2,3,5}` while the commit numbers them 1–4. Either rename `leak5` → `leak4` or note the gap so future-you doesn't go hunting for the missing one.

_What's good:_ `discardBody` shape (small, defensive, well-documented), the synchronous-attach for `once("exit")` in `register.ts` (closes the exit-already-fired race), the switchable-handler pattern in `process-registry.ts`, and the honest disclosure on leak #3. Spot-checked ~25 un-fixed `!res.ok` sites and they all correctly drain via `.text()`/`.json()` or operate on Result envelopes.

packages/mcp/src/process-registry.ts
```diff
@@ -443,6 +463,10 @@ class ProcessRegistry {
             status: response.status,
             elapsedMs: Date.now() - startTime,
           });
+          // Stop accumulating once we're past the failure-diagnostic window.
+          // The handler stays attached (kernel pipe buffer would otherwise
+          // back-pressure a chatty child) but now drops bytes on the floor.
+          accumulating = false;
```
basedfriday, line 479: Adjacent gap in the same file: the readiness-probe `Response` at line 440 (`deps.fetch(readyUrl, { method: "GET" })`) is never drained on either branch — the 5xx case logs and falls through, the success case just `return { child }`s. This is a polling loop, so a slow-to-bind child drops N un-drained Responses per startup. Worth a `await discardBody(response)` before the `if (response.status >= 500)` check (or `await response.text()` if you want to log the 5xx body). Same file, same class of leak as the one you're already fixing.

ljagiello:
Thanks for the careful review @basedfriday — all three findings addressed in `2120550b7`:

1. **skills-sh-client.ts** — added `discardBody(response)` before both throws (search L132 + download L166). Same pattern, same fix.
2. **process-registry readiness probe leak** — good catch, this was a real per-poll leak. Single `await discardBody(response)` immediately after the fetch resolves, covers both the 5xx not-ready branch and the success branch. Was definitely going to bite under slow-to-bind workspace-mcp / OAuth bootstrap.
3. **Stale line ref** — rewritten to "thrown in the readiness loop below" so it can't rot further.

Also renamed `leak5-stderr-accumulator.ts` → `leak4-...` to match the PR's 1–4 numbering. The "5" was a leftover from my original incident-analysis numbering (where #4 was the unrelated launcher KeepAlive trap that didn't land here).

Verified: `deno check` clean, 215 vitest tests in the touched packages green, leak4 bench numbers reproduce post-rename (BROKEN 31.3 MiB accumulator / FIXED 0.81 MiB).

packages/mcp/src/process-registry.ts
ljagiello: Done in 2120550b7 — replaced the line ref with "thrown in the readiness loop below" so it can't drift on future edits to this hunk.

packages/mcp/src/process-registry.ts
ljagiello: Good catch — fixed in 2120550b7. Single `await discardBody(response)` immediately after the probe fetch resolves, before any branching, so both the 5xx not-ready and the success paths drain through one call site. Workspace-mcp's OAuth-bootstrap 5xx pages would have been the canonical victim — multi-page bodies × N poll iterations × every cold spawn.

PR #352, opened 2026-05-15
Allow users to set secret values for environment variables in chat

Vpr99:
## Summary

- Lets the user type secret-looking env values directly into the confirmation card so they never enter chat history. Card sends a `varsOverride` map on `/answer`; the daemon merges it over the proposed vars at commit time, only for keys already in the proposal (no smuggling new keys), and re-applies `env_set`'s POSIX-key / no-newline validation to the override values.
- Updates the `env_set` tool spec: agents pass `""` for secret-bearing keys so the user fills the real value via the card. Carve-out: if the user has already pasted the literal value in chat, the agent passes it through — privacy is already gone and forcing a retype is pure friction.
- Aligns the confirmation card to project tokens and patterns: `--surface-dark` recessed surface (was: 8% blue wash), inset `--color-info` ring on pending (was: `--blue-primary` halo), status badges matched to `MemoryEntryTable`'s `.badge` convention (sentence case, brand `--green-primary` / `--red-primary` for applied / denied), buttons right-justified, disabled-Confirm hint and Expired explanation moved into `Tooltip`s. Terminal text removed for applied/denied/expired — the badge carries the state.

## Card polish

A second pass focused on compactness, fewer font registers, and consistent rendering across turns:

- **Title slot.** Dropped the monospace-uppercase `ENVIRONMENT WRITE` eyebrow + separate scope line in favor of a single sentence-case title ("Set N variable(s) in workspace .env"). Monospace stays for identifiers only (keys, `.env`, input values), leaving the card on two typographic registers instead of three.
- **Var-row as one envelope.** Key label slot + vertical divider + transparent input + optional reveal — all share one bordered container. `:focus-within` highlights the whole envelope (border + divider) as a unit.
- **Footer row.** Credential-bearing hint moved inline with the action buttons (hint left, buttons right via `margin-inline-start: auto`). Falls back to a standalone line in terminal states.
- **Definitive width.** Card sized with explicit `inline-size: var(--size-128)` + `align-self: flex-start` so the same elicitation renders identically regardless of the accompanying text bubble's width. (`max-inline-size` alone wasn't enough — `align-items: stretch` on the parent took the card right up to the cap based on what's beside it.)
- **Deny button** → `variant="none"` (ghost) so Confirm reads as the primary action and Deny as quiet-cancel.
- **Disabled-Confirm tooltip** copy → "Fill in any blank values." (was a three-clause sentence with "secret-looking key" hedging).
- **Bundled card height** on the 1-variable common case: roughly halved.

## Chat message layout

Reworked the surrounding message row so the actions row (time, duration, usage, ellipsis) aligns with the bubble's right edge instead of the wider message column when a tool card is present:

- `.message-content` text bubble: `inline-size: fit-content` with `max-inline-size: 100%` so the bubble paints only as wide as its text.
- `.message.assistant`: `width: fit-content` so the column sizes to the widest thing the agent said.
- New `.text-actions-group`: when an assistant turn has both a tool card and text, the last text bubble + actions row get wrapped in one `fit-content` block so the trailing-edge ellipsis tracks the bubble, not the card.
- `.message.user` gets `align-items: flex-end` so small bubbles aren't stranded on the left of a right-anchored column when the actions row content is wider than the bubble.
- Timestamps drop seconds inline ("2:44 PM"), with full date+time+seconds available on hover via the `@atlas/ui` `Tooltip` and on Alt-hold inline (kept for scanning many timestamps at once).
- Inline custom ellipsis SVG replaced with `<Icons.TripleDots />` from `@atlas/ui`; `.message-actions` switches to `align-items: center` so the dots line up with the timestamp x-height.

CleanShot 2026-05-15 at 15 23 18

| Before / After |
| --- |
| CleanShot 2026-05-15 at 14 38 19@2x |
| CleanShot 2026-05-18 at 14 59 16@2x |


basedfriday:
UI needs to be more compact. Also, too many font faces IMO.

Vpr99:
> UI needs to be more compact. Also, too many font faces IMO.

CleanShot 2026-05-18 at 13 29 52@2x


Thoughts?

ljagiello (commented):
Review complete — **Needs Work**, but light. Smuggling defense, schema validation, and commit-before-answer ordering are sound and test-protected. The substantive items below are: (1) applied non-secret values display the stale proposal after page refresh; (2) test hygiene around `vi.clearAllMocks()` and a missing `declined` case in the terminal-state test; (3) a small TOCTOU window between the new `status !== "pending"` pre-check and the CAS in `transitionPending`. Full review doc with triage and `Worth doing: yes/no` calls landed at `docs/reviews/2026-05-18-pr-352-env-write-vars-override.md`.

**Important (worth doing before merge):**
- Applied card shows stale proposal for non-secret keys after refresh — `fetchAppliedValue` is wired only to the secret-key reveal. See inline.
- `vi.clearAllMocks()` in `index.test.ts:138-139` doesn't drain `mockResolvedValueOnce` queues; switch to `mockReset()` on each hoisted mock. Today's ordering happens to be safe; next inserted test silently picks up a leftover stub. (Outside the diff, can't inline.)
- Terminal-state test omits `declined` — see inline.
- `missingSecretValue` is the only client gate preventing empty-secret confirm; untested. See inline.

**Minor / Needs Decision:**
- `varsOverride` is always sent on confirm (full payload, even when nothing was edited) — could trim to user-changed delta and let the dead `hasOverride ? ... : ...` branch in `answer()` go away.
- `env_set` carve-out for "user already typed it in chat" relies on LLM judgment; the load-bearing assumption comment already calls this out. Drop the carve-out (force retype) or document it explicitly — either is fine; today's middle ground is the riskiest one.
- PR bundles ~300 lines of unrelated chat-layout polish with the security-relevant ~150 lines. Not worth splitting now, but worth noting for next time.

No blockers — most of these are 1–2 line fixes.

tools/agent-playground/src/lib/components/chat/env-set-tool-card.svelte
```diff
@@ -111,6 +129,37 @@
...
 
+  /**
+   * Pull the current value of a secret-bearing key from the daemon's env
+   * endpoint (which returns the real value — only the agent-facing
+   * `env_get` tool masks). Used to reveal what the user typed into an
+   * applied card after a page refresh, without persisting the value on
+   * the elicitation answer record. The value travels: card → daemon →
+   * `.env` → daemon → card, never touching the chat message stream.
+   */
+  async function fetchAppliedValue(key: string): Promise<void> {
```
ljagiello, line 142: **Stale display on applied non-secret values.** `fetchAppliedValue` is only invoked from the secret-key reveal button (line ~243). For a non-secret key the user edited and confirmed (e.g. `LOG_DIR` corrected from `/var/log` to `/srv/log`), a page refresh rehydrates `userValues` from the proposal, not from disk. The card then shows `/var/log` while `.env` holds `/srv/log` — the card lies about what was committed.

Cheap fix: lazily fetch every applied value on mount when status is `applied`, not just on reveal-click. Non-secret values aren't masked so there's no privacy reason to defer them.

**Worth doing:** yes — the card is supposed to reflect what was written; silently showing the proposal undermines auditability.

tools/agent-playground/src/lib/components/chat/env-set-tool-card.svelte
```diff
@@ -86,20 +79,45 @@
...
+  $effect(() => {
+    for (const [key, value] of entries) {
+      if (!(key in userValues)) {
+        userValues = { ...userValues, [key]: value };
+      }
+    }
+  });
+
+  /** Confirm is blocked while any secret-looking key is empty (or whitespace-only). */
+  const missingSecretValue = $derived(
```
ljagiello: **`missingSecretValue` is load-bearing but untested.** The server schema accepts `z.string()` (empty strings included), so this derived is the *only* gate preventing confirm-with-empty-secret. A typo (`!isSecretKey(k)` flipped, or `.trim().length` losing the negation) would silently allow blank-confirm of a credential-bearing key.

The new `env-set-tool-card.test.ts` exercises a 6-line pure helper but skips this branch. Worth a 2–3-line direct test: assert `missingSecretValue === true` for `[["API_KEY", ""]]` with empty `userValues`, and `=== false` once a real value lands.

apps/atlasd/routes/elicitations/index.test.ts
```diff
@@ -348,6 +355,28 @@ describe("POST /:id/answer", () => {
       expect(mockElicitationStorage.answer).not.toHaveBeenCalled();
     });
 
+    test.each([
+      "answered",
```
ljagiello, line 370: **`declined` is missing from this `test.each`.** The implementation guard is `got.data.status !== "pending"` — so it already catches `declined`, but the test only pins `answered` and `expired`. `declined` is a reachable production state (POST `/decline` flips it), and replaying a confirm after declining is exactly the kind of state-machine attack the guard exists to prevent.

One-line fix:

```diff
-    test.each([
-      "answered",
-      "expired",
-    ] as const)
+    test.each([
+      "answered",
+      "expired",
+      "declined",
+    ] as const)
```

apps/atlasd/routes/elicitations/index.test.ts
```diff
@@ -414,6 +443,152 @@ describe("POST /:id/answer", () => {
...
+        "real-secret-from-card",
+      );
+      // Non-overridden key keeps its proposed value.
+      expect(mockSetEnvFileVar).toHaveBeenCalledWith("/tmp/ws_1/.env", "LOG_DIR", "/var/log");
+      const answerInput = mockElicitationStorage.answer.mock.calls[0]?.[0] as {
+        answer: Record<string, unknown>;
+      };
+      expect(answerInput.answer).toEqual(expect.objectContaining({ value: "confirm" }));
+      expect(answerInput.answer).not.toHaveProperty("varsOverride");
+      expect(JSON.stringify(answerInput.answer)).not.toContain("real-secret-from-card");
```
ljagiello: **Substring-on-JSON is fragile for this invariant.** This works because the test secret is plain ASCII. A secret containing `"`, `\`, or other JSON-escapable bytes would be encoded in `JSON.stringify` (`foo\"bar`) and a literal substring check would miss it. The stronger assertion already exists one line above: `not.toHaveProperty("varsOverride")`.

Either drop the substring check, or swap the test secret to one containing a `"` to prove escaping doesn't sneak the secret through. Today the assertion's truth value depends on the chosen string, not on the structure under test.

apps/atlasd/routes/elicitations/index.ts
```diff
@@ -289,15 +330,25 @@ elicitationApp.post(
...
+      // accepted asymmetry: the write IS done, and a stuck-`pending`
+      // elicitation is a far better failure than a silently-lost write. There
+      // is no compensating rollback.
+      //
+      // Note: `userValues` on the confirmation card is in-memory only — on
+      // commit failure + page refresh, the user has to retype any secret
+      // they entered. Acceptable because commit failure is rare.
       if (got.data.kind === "env-write" && body.value === "confirm") {
-        const commit = await commitEnvWriteElicitation(c, got.data);
+        if (got.data.status !== "pending") {
```
ljagiello, line 356: **Small TOCTOU window worth a comment.** This `status !== "pending"` is a pre-check; the real CAS happens inside `transitionPending` in `ElicitationStorage.answer` (after the commit on line 351). If two `/answer` requests for the same id race past this check, both call `commitEnvWriteElicitation` and both write to disk. The CAS lets exactly one envelope update win — but the *loser's* `setEnvFileVar` may have run second and overwritten disk. The answered envelope can record values A while `.env` holds values B.

In practice this needs network retries or a double-click to manifest, so I'd rate it minor — but worth one sentence appended to the existing asymmetry comment block (lines 333-344) so the next person doesn't have to re-derive it.

basedfriday (approved):

PR #354, opened 2026-05-16
feat(webhook-tunnel): drop hardcoded bitbucket/jira providers + ship wiring skill

ljagiello:
## Summary

Three threads of work landed on this branch, each with promptfoo eval coverage:

### 1. Webhook-tunnel: drop `bitbucket` + `jira` providers — `raw` is the only path

The embedded webhook-tunnel now ships exactly one provider: `raw`. Workspace agents own parsing, event filtering, and (optional) signature verification for every upstream.

- `raw` forwards the whole request body to the workspace agent — agents can handle every GitHub/Bitbucket/Jira/custom event, do whatever transform makes sense for their domain.
- The tunnel does no HMAC verification, no event allowlisting, and reads no provider-specific env vars. The previous embedded `github` / `bitbucket` / `jira` paths are gone.
- New `wiring-external-webhooks` skill teaches the raw-only path: URL shape, `/status` discovery, signal-schema `additionalProperties: true` requirement, and the upstream UI steps for Bitbucket and GitHub.
- New Tier 0 example in `writing-friday-python-agents` covering `ctx.input.config or ctx.input.raw` — agents were mis-using `parse_input(prompt)` for signal-direct invocations.

### 2. Signal-trigger: `?nowait=true` publish-ack mode + SSE follow

The HTTP signal-trigger endpoint now supports three response modes:

| Mode | When |
|---|---|
| **`?nowait=true`** → 202 + `{message, status:"accepted", workspaceId, signalId, correlationId}` | Fire-and-forget — caller publishes a signal but doesn't need the cascade output. Atlasd publishes to JetStream and returns in <100ms; the cascade runs async on the CASCADES consumer. The webhook-tunnel uses this internally now, fixing the 30s-deadline timeouts on long cascades. |
| **Synchronous JSON** (default) → 200 + `output` + `summary` + `sessionId` | Caller consumes the cascade result programmatically (CLI tools, RPC handlers, scripts). |
| **SSE** (via `Accept: text/event-stream` on the trigger POST) | Caller watches cascade events live (UIs, log tails). Subscribe-then-publish happens in one handler — there is no separate `GET /signals/stream/<correlationId>` endpoint, because the cascade response is published to a core-NATS subject with no replay. |

- Webhook-tunnel forwarder appends `?nowait=true` automatically + has clearer timeout error (`atlasd publish ACK timeout (30s) — bus normally returns in <100ms; check /health + JetStream`).
- `friday-cli` SKILL.md teaches all three modes so non-webhook callers (CLI, MCP server, agent playground) also pick the right one.
- Three callers updated to discriminate on `status === "completed"` before destructuring `sessionId`/`output` (the response shape is now a union): `workspace-chat`'s `job-tools.ts`, `mcp-server`'s `signals/trigger.ts`, `agent-playground`'s `run-job-dialog.svelte`.

### 3. Skill content: Anthropic best-practices audit pass

Audited the changed skills against [Anthropic's prompt-engineering](https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/claude-prompting-best-practices) and [skill-authoring](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices) docs. Three fixes:

- `friday-cli` description was imperative ("Interact with..."); flipped to third-person ("Interacts with...") per the discovery-injection rule.
- `wiring-external-webhooks` had baked-in date references in body text; reframed as present-tense facts so the skill doesn't read as stale within months.
- HMAC-secret rule led with a negative enumeration; flipped to lead with the positive rule (verification lives in the agent, not the tunnel), then list the silently-ignored variants.

## Eval coverage — 12 scenarios, all passing stably

The promptfoo eval (`webhook-setup-bitbucket`) grew to 12 scenarios covering every behavior change:

| # | Scenario | What it locks in |
|---|---|---|
| 1 | `with-skill` | Positive baseline — all setup-guidance checks pass with skill loaded |
| 2 | `without-skill` | Control — must fail without the skill (proves the skill is the carrier) |
| 3 | `corrects-wrong-env-var` | User mentions `BITBUCKET_WEBHOOK_SECRET` → agent corrects without blaming the restart |
| 4 | `points-at-status-first` | Webhook isn't firing → agent surfaces `/status` first, not "restart the daemon" |
| 5 | `test-webhook-no-bypass` | User wants to test the webhook → agent offers real test (curl through tunnel, Bitbucket Test connectivity, `atlas signal trigger`) without recommending a sidecar/bypass |
| 6 | `recommends-nowait-pattern` | User doesn't need output → agent recommends `?nowait=true` + 202 + correlationId, not raise timeout / long-poll / sidecar |
| 7 | `recommends-sync-for-output` | Programmatic caller needs the cascade output → agent recommends default sync mode, not nowait, not polling |
| 8 | `recommends-sse-for-progress` | UI watching cascade progress → agent recommends SSE, not polling, not parsing the sync output |
| 9 | `lists-current-providers` | Agent names `raw` as the only built-in, does NOT list bitbucket/jira/github, does NOT recommend `WEBHOOK_MAPPINGS_PATH` to register others |
| 10 | `no-atlasd-url-fallback` | Agent emits `/hook/raw/` tunnel URL or asks the user to fetch `/status`; never synthesizes `/api/workspaces/.../signals/` |
| 11 | `webhook-signal-schema-passes-body` | Signal schema is `type:object, additionalProperties: true` (or bare top-level) — never nested `type: object` without `additionalProperties`, which Zod would silently strip |
| 12 | `github-setup-walkthrough` | GitHub UI steps emit `/hook/raw/` URL + `application/json` content type, name `issue_comment` / `pull_request` / `push` events, point at `/status`, no atlasd-direct URL, no hand-rolled tunnel install |

Implementation uses the same lightweight `drive()` pattern as `secret-input-guard.ts` — direct Anthropic API calls, no daemon spawn, ~15s per scenario.

## Migration

Any webhook still pointed at `/hook/bitbucket/...`, `/hook/jira/...`, or `/hook/github/...` returns `400 Unknown provider: <name>. Available: raw`. Update the URL in the upstream system to `/hook/raw/{workspaceId}/{signalId}` and let the workspace agent handle parsing + (if needed) HMAC verification. The new `wiring-external-webhooks` skill walks through this end-to-end.

`?nowait=true` is opt-in; existing callers using the default sync mode are unaffected.

## Byte-for-byte forwarding (added during review)

`webhook-tunnel` is now a true reverse proxy: `httputil.ReverseProxy` rewrites only host + path and streams body bytes + headers verbatim. atlasd's `/api/workspaces/:ws/signals/:sig` discriminates on body shape (no envelope keys → webhook mode) and preserves the raw bytes + lowercased headers on the JetStream envelope (`webhookBody` base64, `webhookHeaders`). They thread through the cascade into the FSM context and surface to the agent as `ctx.input.raw["body"]` (decode → bytes) and `ctx.input.raw["headers"]`. The agent can recompute HMAC against the exact upstream-signed bytes — verified end-to-end by registering a webhook on `tempestteam/atlas`, posting a PR comment, and observing `hmac_valid: true` in the agent output.

## Follow-ups (not in this PR)

- **Agent-side HMAC docs + eval coverage** — `wiring-external-webhooks` is intentionally the connect-the-webhook surface only (URL shape, `/status` discovery, upstream UI steps). HMAC verification is an agent-implementation concern and belongs in the agent-authoring docs or a dedicated skill — not here. The `webhook-setup-bitbucket` promptfoo eval currently pins Bitbucket-flavored connection guidance only; GitHub + Jira eval scenarios should land alongside the agent-side HMAC docs in a separate task.

## Test plan

- [x] `go test ./tools/webhook-tunnel/...` — provider tests updated, all pass
- [x] `/status` reports `providers: [raw]` after daemon restart
- [x] Promptfoo eval `webhook-setup-bitbucket` passes stably across consecutive full runs
- [x] End-to-end: natural-language prompt creates a workspace with a `provider: http` signal + FSM job + Python user agent + loop guard; an external webhook POSTs to `/hook/raw/{wsId}/{signalId}` → tunnel forwards with `?nowait=true` → workspace agent runs → ACK reply posted back upstream in ~5s. Run 20× consecutively against both Bitbucket and GitHub with 100% pass.
- [x] Loop containment verified: atlasd's `concurrency: skip` default blocked the ACK→webhook→ACK cycle; agent-side marker guard is the belt-and-braces second layer
- [x] Three union-discriminator call sites verified through the UI (run-job-dialog) + via MCP signal trigger + via chat job-tools


tools/webhook-tunnel/forwarder/forwarder.go
```diff
@@ -84,25 +84,41 @@ func buildTransport(caCertPath string) (http.RoundTripper, error) {
...
+// genuinely sick and a clearer error than `context deadline exceeded`
+// is what callers want.
 func (f *Forwarder) Forward(workspaceID, signalID string, payload map[string]any) (string, error) {
 	body := map[string]any{"payload": payload}
 	encoded, err := json.Marshal(body)
 	if err != nil {
 		return "", fmt.Errorf("marshal: %w", err)
 	}
-	endpoint := fmt.Sprintf("%s/api/workspaces/%s/signals/%s",
+	endpoint := fmt.Sprintf("%s/api/workspaces/%s/signals/%s?nowait=true",
```
Vpr99: `?nowait=true` here is the central new behavior of this PR. Delete it from this line and every test still passes. `TestForwardHappyPath` asserts `r.URL.Path` but not `r.URL.RawQuery`. The pattern is local: `TestProxyPathRewrite` already captures `RawQuery` in the same file. Two-line fix:

```go
if r.URL.RawQuery != "nowait=true" {
    t.Errorf("expected nowait=true, got %q", r.URL.RawQuery)
}
```

Without it, a regression that re-couples the upstream webhook to the full cascade duration ships green.

tools/webhook-tunnel/forwarder/forwarder.go
```diff
@@ -120,8 +151,11 @@ func (f *Forwarder) Forward(workspaceID, signalID string, payload map[string]any
 	}
 	respBody, _ := io.ReadAll(io.LimitReader(resp.Body, 1<<20))
 	var sr SignalResponse
-	_ = json.Unmarshal(respBody, &sr) // session ID is optional
-	return sr.SessionID, nil
+	_ = json.Unmarshal(respBody, &sr) // both id fields are optional
+	if sr.SessionID != "" {
+		return sr.SessionID, nil
+	}
+	return sr.CorrelationID, nil
```
Vpr99: Dual semantics in one return slot: `SessionID` (sync) or `CorrelationID` (nowait). The caller at `main.go:384-385` writes the result into `resp["sessionId"]` and the logger at `main.go:382` calls it `session_id`. Now that the tunnel forces `?nowait=true`, the default path returns a correlationId labeled as a sessionId — ops tailing the log and HTTP clients reading `resp.sessionId` both get something that isn't one. Suggest returning `(id, kind, error)` and renaming the response field to match what's actually there.

apps/atlasd/routes/workspaces/index.ts
```diff
@@ -2136,6 +2136,68 @@ const workspacesRoutes = daemonFactory
...
+      // arriving after the response was published would silently miss it.
+      // Callers that need to watch progress must use the in-handler SSE path
+      // (Accept: text/event-stream on the POST below) which subscribes
+      // BEFORE publishing.
+      const nowait =
+        c.req.query("nowait") === "true" ||
+        c.req.query("nowait") === "1" ||
+        c.req.query("wait") === "false" ||
+        c.req.query("wait") === "0";
+      if (nowait) {
```
Vpr99, line 2304: Topology problem with the three-mode contract: the SSE middleware at line 1712 short-circuits before reaching this branch. A request with `?nowait=true` and `Accept: text/event-stream` together routes to SSE; this `if (nowait)` block never runs. No documented precedence and no test pinning the choice. Pick one (reject the combo with 400, or document that Accept wins) and add one test.

apps/atlasd/routes/workspaces/index.ts
```diff
@@ -2136,6 +2136,68 @@ const workspacesRoutes = daemonFactory
...
+      // webhook-tunnel that don't need cascade output to respond to upstream.
+      //
+      // No follow-by-correlationId endpoint: the cascade response is published
+      // to a core-NATS subject (no JetStream replay), so a follow-up GET
+      // arriving after the response was published would silently miss it.
+      // Callers that need to watch progress must use the in-handler SSE path
+      // (Accept: text/event-stream on the POST below) which subscribes
+      // BEFORE publishing.
+      const nowait =
+        c.req.query("nowait") === "true" ||
```
Vpr99: Four spellings for one toggle: `nowait=true`, `nowait=1`, `wait=false`, `wait=0`. The webhook-tunnel forwarder only sends `nowait=true`, and no test covers the other three. Drop them. Two lines saved, and you avoid the future "why does `wait=0` work but `wait=no` doesn't?" question.

packages/mcp-server/src/tools/signals/trigger.ts
```diff
@@ -149,6 +149,23 @@ export function registerSignalTriggerTool(server: McpServer, ctx: ToolContext) {
 
         const response = result.data;
 
+        // The route can return a 202 "accepted" envelope (?nowait=true)
+        // or the default "completed" envelope. This MCP tool only triggers
+        // synchronously, so we expect "completed" — but narrow defensively
+        // so a future default flip doesn't crash here.
+        if (response.status !== "completed") {
```
Vpr99: This narrowing has no test. `job-tools.test.ts` exercises both branches of the same union for the chat path; mirror that pattern here. MCP tools are reached by LLM clients, which makes silent shape drift on the response envelope highest-stakes among the four discriminator call sites. CLI and the Svelte dialog are lower stakes and the inline comments cover them — but this one warrants a Vitest case.

tools/webhook-tunnel/main_test.go
```diff
@@ -134,22 +123,13 @@ func TestHookValidSignature(t *testing.T) {
...
 	}
-	defer func() { _ = resp.Body.Close() }()
-	if resp.StatusCode != http.StatusUnauthorized {
-		t.Fatalf("expected 401, got %d", resp.StatusCode)
+	// The forwarder wraps the body as {"payload": <body>}; just sanity-check
+	// that the actor field survived round-trip.
+	if !bytes.Contains(seenBody, []byte("alice")) {
+		t.Errorf("body did not reach atlasd intact: %s", seenBody)
 	}
 }
```
Vpr99, line 149: The adjacent `TestHookUnknownProvider` covers `/hook/martian/...`, but the realistic failure mode after this PR is a user POSTing to a pre-PR `/hook/github/...`, `/hook/bitbucket/...`, or `/hook/jira/...`. Table-test those three names; assert the body contains both the bad provider name and `raw` in the available list. Locks the migration UX so a future refactor of the diagnostic string at `main.go:326` trips a test.

packages/system/skills/wiring-external-webhooks/SKILL.md
```diff
@@ -0,0 +1,147 @@
+---
```
Vpr99, line 1: The skill now owns the HMAC reminder for every provider. The promptfoo eval scenario pins the Bitbucket version. GitHub and Jira have no equivalent. Either extend the eval matrix, or call out in the PR description that GitHub/Jira HMAC coverage is a follow-up. Failure mode if an agent skips verification: anyone with the tunnel URL can fire signals.

Vpr99 (commented):
Inline review notes. Direction is right; the seven items below are worth fixing before merge. Full notes in `docs/reviews/2026-05-18-pr354-webhook-tunnel-raw-provider.md` on the branch.

tools/webhook-tunnel/forwarder/forwarder.go
ljagiello: Obsolete after the byte-for-byte refactor in 9ef4ad50f — `Forward()` is deleted, the tunnel is now a pure `httputil.ReverseProxy` that doesn't add any query params (atlasd auto-forces nowait when webhook mode is discriminated by body shape). The test it referenced (`TestForwardHappyPath`) is replaced by `TestHookForwardsByteForByte` which locks the byte-identical body + headers invariant.

tools/webhook-tunnel/forwarder/forwarder.go
ljagiello: Obsolete — `SignalResponse` struct is deleted in 9ef4ad50f along with the rest of the JSON-parsing path. The reverse proxy streams the atlasd response back to the upstream verbatim; no field labeling happens on the tunnel side.

apps/atlasd/routes/workspaces/index.ts
ljagiello: Fixed in 8c826956d. Documented Accept-header-wins precedence inline at the nowait branch + locked with a vitest: a request with both `Accept: text/event-stream` and `?nowait=true` reaches the SSE handler (returns 400 from its envelope guard), never the JSON handler's webhook fast-path (which would return 202). Chose to lock the existing behavior rather than reject the combo — Accept-header-wins is the natural Hono routing order.

apps/atlasd/routes/workspaces/index.ts
ljagiello: Fixed in 8c826956d. Collapsed to `?nowait=true` only. Test asserts non-canonical spellings (`nowait=1`, `wait=false`, `wait=0`, `nowait=yes`) do NOT take the nowait fast-path.

packages/mcp-server/src/tools/signals/trigger.ts
ljagiello: Fixed in 8c826956d. Extracted `mapSignalTriggerResponse(workspaceId, signalId, response)` as a pure function covering the full `completed | accepted | failed | cancelled` union; 6 vitest cases lock every branch including missing-error/missing-reason fallbacks. A schema change (renamed `sessionId`, dropped a variant) trips the test.

tools/webhook-tunnel/main_test.go
ljagiello: Fixed in 8c826956d. `TestHookUnknownProvider` now iterates `{github, bitbucket, jira, martian}` and asserts (a) 400, (b) response body names the rejected provider, (c) response body names `raw` so the operator sees the migration path without grepping docs.

packages/system/skills/wiring-external-webhooks/SKILL.md
ljagiello: Called out in the updated PR body under "Follow-ups" — GitHub + Jira HMAC eval scenarios should land as a separate task. Failure mode (anyone with tunnel URL can fire signals) noted there.

ljagiello: Correction on my earlier reply — `wiring-external-webhooks` is the connect-the-webhook surface only (URL shape, /status discovery, upstream UI steps). HMAC verification is an agent-implementation concern and explicitly out of scope for this skill. PR body Follow-ups section updated to reframe the eval matrix item as agent-side concern (separate task), not a skill-owned reminder.

Issue #330, opened 2026-05-15
env tools: thread auth header for non-local daemon installs (env_get/list/delete 401)


basedfriday:
From PR #323 review (Vpr99).

**Finding:** `env_get`, `env_list`, and `env_delete` all call `fetch(daemonUrl/...)` with no `Authorization` header. In local/dev mode the daemon auto-mints a user so this works. In any non-local install (or once auth is tightened), every call silently 401s — the env read/delete tools are completely non-functional.

`env_set` is unaffected (it writes via ElicitationStorage directly, not HTTP).

**Fix options:**
- Add `apiKey` to `ToolContext`, pass `FRIDAY_KEY` from daemon construction, send `Authorization: Bearer <key>` header in each fetch call.
- Better: move env tools to call the workspace facade directly (same-process call, no HTTP hop, no auth dance). Same pattern improvement applies to memory tools which have the same inherited wart.

**Scope:** Not a concern in current single-local-user Studio; becomes a breakage when auth is tightened or multi-user is introduced.

Refs: `packages/mcp-server/src/tools/env/get.ts:37`, `list.ts:33`, `delete.ts:38`

LissaGreense:
Poked at this before picking it up and two things confused me — wanted to check before going further.

The first option (Bearer `FRIDAY_KEY`) — does that actually authenticate today? Everywhere else in the codebase `FRIDAY_KEY` is used outbound (Link, Persona, gateway), and the session middleware validates inbound tokens against the `SESSIONS` KV bucket. Sending it as a Bearer header would land the plumbing but still 401, no? Unless I'm missing a path where it gets accepted as a service token.

For the second option — when you say "no auth dance," do you mean in-process callers stay unauthenticated by construction (anything running in the daemon's own process is inside the trust boundary already)? Or did you have something else in mind?

Also stumbled into `packages/system/agents/workspace-chat/tools/env-tools.ts` — the chat agent has its own `env_get` / `env_set` factory that mirrors the MCP ones, and its `env_get` does the same bare `fetch` with no auth. Its docstring says chat doesn't connect to the platform MCP server, so the MCP env tools never reach a chat LLM. Should this issue cover that file too, or is that a separate concern?

PR #395, opened 2026-05-19
fix(playground): accept workspace zip in sidebar Add Space dialog

### LINKED ISSUE #210: [BUG] Can't upload a workspace zip from the "Add Space" button
### What happened?

The plus button next to the Spaces label in the sidebar only allows a single `workspace.yml` file to be uploaded. In order to upload a full zip, you have to go to Settings. This can cause confusion and should be consistent.

### Version

_No response_

### Logs / screenshots

_No response_

LissaGreense:
## Summary

The sidebar's "Add Space" plus button → Upload File tab only accepted `.yml`/`.yaml`. To import a workspace zip you had to detour through Settings, even though the user intent is identical: "add one space here." This PR makes both surfaces accept both formats.

Closes #210.

## Behavior

| Input | Path |
|-------|------|
| `.yml` / `.yaml` | `POST /api/workspaces/create` (unchanged) |
| `.zip` (single-workspace export) | `POST /api/daemon/api/workspaces/import-bundle` (multipart `bundle` field) |
| `.zip` (full-export, many workspaces) | _Not handled here._ Stays in Settings — that's a bulk-migration flow, not "add a space." |

On success: invalidate the workspace query, close the dialog, navigate to the imported / created workspace's platform page. Mirrors the existing yml flow exactly.

## Implementation

- `tools/agent-playground/src/lib/components/workspace/workspace-loader.svelte`
  - `accept=".yml,.yaml"` → `accept=".yml,.yaml,.zip,application/zip"`
  - Drop-validator + drop-zone copy updated to mention zip
  - Splits the loadFile body so YAML and zip code paths are readable side-by-side (`createFromYaml(file)` / `importBundle(file)`) instead of nested branches
  - Zod-parses the import-bundle response (`workspaceId?`, `error?`) so the error fallback is explicit

## Risks

- **Endpoint:** `/api/workspaces/import-bundle` exists at `apps/atlasd/routes/workspaces/index.ts:1233` and is the same endpoint Settings has been using successfully for single-workspace imports. Confirmed shape: multipart `bundle` field, returns `{ workspaceId, path, name, ... }` or `{ error }`.
- **Auth:** Local dev daemon auto-mints a user; same as Settings flow. Future tightening would affect both surfaces identically (separate concern, tracked in #330).
- **Behavioral regression for YAML path:** None — that code path is byte-equivalent, just extracted into `createFromYaml()`.
- **Misclassification:** Files are classified by extension (with `application/zip` as a positive confirmation for the MIME type). Safari sometimes reports zips with `type: ""` so the extension check has to be the canonical signal.

## Test plan

- [x] `npx svelte-check` clean on the touched file (0 errors / 32 pre-existing unrelated warnings)
- [x] Manual: drop a `.yml` → workspace creates, navigates correctly (existing path, untouched)
- [x] Manual: drop a `.zip` exported from Settings > "Download a workspace" → workspace imports, sidebar refreshes, lands on the new workspace's page
- [x] Manual: drop a `.txt` or other garbage → friendly error, no request fires
- [x] Manual: drop a malformed zip → error surfaces from the daemon's `error` field, dialog stays open

### DIFF
--- a/tools/agent-playground/src/lib/api/workspace-import.ts
+++ b/tools/agent-playground/src/lib/api/workspace-import.ts
@@ -58,5 +58,3 @@ export const ImportBundleAllResponseSchema = z
 export const ImportBundleErrorSchema = z.object({ error: z.string() }).passthrough();
 
 export type ImportedEntry = z.infer<typeof ImportedEntrySchema>;
-export type ImportBundleResponse = z.infer<typeof ImportBundleResponseSchema>;
-export type ImportBundleAllResponse = z.infer<typeof ImportBundleAllResponseSchema>;
--- a/tools/agent-playground/src/lib/api/workspace-import.ts
+++ b/tools/agent-playground/src/lib/api/workspace-import.ts
@@ -1,0 +1,62 @@
+/**
+ * Response schemas for the daemon's workspace-import surface.
+ *
+ * `POST /api/workspaces/import-bundle` accepts a single workspace zip
+ * (`multipart/form-data` field `bundle`) and returns one imported
+ * workspace. `POST /api/workspaces/import-bundle-all` accepts a
+ * full-export archive and returns a per-entry list plus aggregate
+ * `errors[]` and optional `globalSkills` status. Both endpoints share
+ * the same `{ error: string }` failure body. The playground proxies
+ * `/api/daemon/*` to the daemon, so callers prefix `/api/daemon`.
+ */
+
+import { z } from "zod";
+
+const ImportedMemorySchema = z
+  .object({ kind: z.string(), path: z.string().optional(), reason: z.string().optional() })
+  .passthrough();
+
+export const ImportedEntrySchema = z
+  .object({
+    workspaceId: z.string(),
+    name: z.string().optional(),
+    path: z.string().optional(),
+    memory: ImportedMemorySchema.optional(),
+    agentsInstalled: z.number().optional(),
+    agentsSkipped: z.number().optional(),
+  })
+  .passthrough();
+
+/** Success body for `POST /api/workspaces/import-bundle`. */
+export const ImportBundleResponseSchema = z
+  .object({
+    workspaceId: z.string(),
+    name: z.string().optional(),
+    path: z.string().optional(),
+    memory: ImportedMemorySchema.optional(),
+    agentsInstalled: z.number().optional(),
+    agentsSkipped: z.number().optional(),
+  })
+  .passthrough();
+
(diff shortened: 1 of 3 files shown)


Vpr99:
Reviewed — verdict: **Clean**. Endpoint path, response shape, error decoding, and `bundle` form field all match Settings and the daemon route. YAML branch is byte-equivalent. Drop label + `accept` attribute pair handles the Safari `type: ""` quirk correctly.

One thing worth fixing in this PR:

### Picker path skips the extension validation the drop path enforces
`workspace-loader.svelte:101-108` — `handleFileInput` calls `loadFile` directly without the `isYamlFile(file) || isZipFile(file)` gate that `handleDrop` runs. `<input accept>` is only a hint (Linux/Firefox in particular lets users pick anything), so a `.txt` selected via the picker falls through to `createFromYaml` and surfaces a raw YAML parser error instead of the friendly "Please drop a .yml, .yaml, or .zip file" string. The gap pre-existed, but the new "Drop a workspace.yml or .zip here, or click to browse" copy now advertises the picker as a first-class path, so the inconsistency is more visible.

Two-line fix using the helpers already defined right above:
```ts
async function handleFileInput(e: Event) {
  const input = e.target as HTMLInputElement;
  const file = input.files?.[0];
  if (!file) return;
  error = null;
  if (!isYamlFile(file) && !isZipFile(file)) {
    error = "Please drop a .yml, .yaml, or .zip file";
    input.value = "";
    return;
  }
  await loadFile(file);
  input.value = "";
}
```

### Needs decision (not blocking)
`ImportBundleSchema` (lines 168-171) decodes the same endpoint that Settings reads inline via `typeof body.error === "string"` (`src/routes/settings/+page.svelte:684-721`). Both work; Zod isn't catching anything `typeof` wouldn't here. Pick whichever should be the convention going forward — happy to defer to your call.

### Notes
- No tests added — acceptable for a one-file UX fix with the manual test plan you ran. If the error-decoding in `importBundle` gets extracted into a pure helper later, it'd be cheap to unit-test.
- `isYamlFile`/`isZipFile` extraction is on the bubble for the "third use" rule, but the Safari-quirk comment on `isZipFile` earns its keep — keep.

LissaGreense:
Both fixed in b4289fb 🙏

Picker path: applied your snippet verbatim.

Decoding: went with Zod. Pulled schemas into `lib/api/workspace-import.ts` and rewrote Settings' `postBundle` to use them too — single + bulk now parse their own schemas instead of one `typeof`-branchy function.

Smoke-tested e2e: sidebar zip, Settings single, Settings full archive, and the bogus-zip error path. All good.

Vpr99 (approved):

PR #402, opened 2026-05-19
fix(playground): use ?nowait=true for Run button + poll for session

ljagiello:
😏 Senryū

Click Run, close the tab.
Cascade lives in the daemon —
Not in your browser.

## Summary

- Run dialog POSTed in **sync mode**: modal closed immediately so it looked fire-and-forget, but the fetch was alive in the background. Any navigation/refresh/tab-close aborted the request, the daemon treated that as `Client disconnected`, and used it as the cue to cancel the spawned session. Long jobs (e.g. fix-a-Jira-ticket, minutes to tens of minutes) always lost the race — the customer log that prompted this had three back-to-back cancellations of the same `ai-fix-ticket` at 23s / 58s / 141s.
- Switched the dialog to `?nowait=true`: daemon publishes to JetStream and returns 202 immediately, **cascade is fully decoupled from the browser**. No daemon edits — 5 other signal-trigger callers (CLI, MCP server, workspace-chat job tools, platform client) untouched.
- The 202 response has no `sessionId` (consumer assigns one when it picks up the message), so to preserve the land-on-the-session-view UX we poll the **same** `/api/sessions` endpoint that already powers the Recent Runs sidebar (`sessionQueries.list`), find the newest session whose `startedAt ≥ click timestamp`, and `goto()` it. ~6s budget, 400ms intervals, silent timeout.

## Test plan

QA'd via `deno task dev:playground` + Chrome (claude-in-chrome). 4/4 cases pass:

- [x] Run sends `POST …/signals/<sig>?nowait=true → 202` (`generate-release-notes`)
- [x] Modal closes immediately, no premature nav
- [x] Cascade **survives navigation** — spawned session ran 9s past a full SPA navigation away and ended on a natural workspace failure, no `Client disconnected` in the daemon log
- [x] Auto-nav lands on the spawned session view — `count-and-sleep` in `braised_oregano` → `/platform/braised_oregano/sessions/5660ccf5-…` showing `RUNNING` within ~3s of the click

Typecheck: 0 errors (33 pre-existing warnings unchanged).

### DIFF
--- a/tools/agent-playground/src/lib/components/workspace/run-job-dialog.svelte
+++ b/tools/agent-playground/src/lib/components/workspace/run-job-dialog.svelte
@@ -14,7 +14,7 @@
 <script lang="ts">
   import { Button, Dialog } from "@atlas/ui";
   import { goto } from "$app/navigation";
-  import { getDaemonClient } from "$lib/daemon-client";
+  import { getDaemonClient, PROXY_BASE } from "$lib/daemon-client";
   import {
     getFieldRendering,
     humanizeFieldName,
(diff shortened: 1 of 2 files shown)


Vpr99:
Nit on the comment block at `run-job-dialog.svelte:129-158` — the safety claim at L151-153 is overstated:

> Filtering by `jobName === jobId` also makes concurrent triggers for *other* jobs in the same workspace (cron, another tab) safe — they can't be misidentified as ours.

True only for *other* jobs. The realistic case — same user, same job, two tabs, or a job whose cron schedule overlaps a manual click — still races: this dialog snapshots, a concurrent source spawns a session for the same job, `findSpawnedSessionId` returns the first non-snapshot match, and the wrong tab wins the navigation. Worth re-framing as a known limitation rather than as safe, so future readers don't trust an invariant that doesn't hold.

(The "clock-free" rationale at L149-151 is fine — it correctly explains why this rev moved away from `startedAt >= since`. Only the safety line is misleading.)

Vpr99 (approved):

ljagiello:
Addressed in 49fcb1390 — reworded the comment block at `run-job-dialog.svelte:143-163` to drop the "safe" claim and call out the same-job same-workspace race explicitly as a known limitation (cron-overlapping-manual or two-tab cases). Failure mode is documented as benign (lands on a real running session of the same job, not an error). Noted the structurally clean fix (thread the daemon's `correlationId` into `SessionSummary` and match on that) as out of scope for this PR. No behavior change.

PR #413, opened 2026-05-20
feat(ui): add HotkeyRegistry and DragDropContext

davidwoolf:
## Summary
- Adds `HotkeyRegistry` and `DragDropContext` to `@atlas/ui` — generic, lifecycle-gated, stack-based.
- Migrates the playground's chat hotkeys + chat drag-drop out of `user-chat.svelte` (~95 lines off the file).
- Migrates 11 more keydown handlers across 7 files to the shared registry; removes 6 `<svelte:window onkeydown>` listeners and 2 ad-hoc `window.addEventListener("keydown")` effects.

## Behavior

**`HotkeyRegistry`** — single window-level `keydown` listener. Stack-based dispatch (top of stack wins; falls through when `when()` returns false). Modifier matching: `cmdOrCtrl` (⌘ on Mac, Ctrl elsewhere), `ctrl` (strict, never ⌘), `meta` (strict ⌘). Single-character keys match case-insensitively so a Shift+D binding declared as `key: "d"` actually fires. Listener attaches lazily and detaches when the last binding is disposed.

**`DragDropContext`** — body-level `dragenter`/`over`/`leave`/`drop`. Lifecycle-gated: route components register in a `\$effect`, unmounting disposes — no path matching. Counter-based nesting (no overlay flicker on child dragenter/leave). Files-only (`dataTransfer.types` includes `"Files"`). Filtering, accept rules, and uploads stay at the call site.

## Side effect

The chat inspector hotkey (**⌘/Ctrl+Shift+D**) now works. The previous handler compared `e.key === "d"`, which fails when Shift produces \`"D"\` — the registry's matcher lowercases single-char keys.

## Test plan
- [ ] Chat: **Ctrl+F** toggles fullscreen; **⌘F** on Mac still opens browser find
- [ ] Chat: **⌘/Ctrl+Shift+D** toggles inspector
- [ ] Chat: **Esc** exits fullscreen only when in fullscreen
- [ ] Chat: drag a file over the chat → overlay appears, drop → upload starts
- [ ] Chat: drag/drop overlay doesn't flicker when dragging across child elements
- [ ] Navigate off chat route → drag/drop no longer responds
- [ ] Command palette: **⌘K** and **⌘/** work anywhere; bare **/** works outside text fields, suppressed inside inputs/textareas/contentEditable
- [ ] **⌘/** inside CodeMirror does NOT open palette (toggle-comment still works)
- [ ] Workspace: **Esc** returns to idle from agents/jobs/signals; does nothing on overview
- [ ] Chat route: **Ctrl+B** toggles chat-list sidebar; state persists across reload; **⌘B** on Mac still opens browser bookmarks bar
- [ ] Memory page: **r** refetches; suppressed when typing in any text field
- [ ] Inspector: **Esc** closes a selected block; does nothing otherwise
- [ ] Jobs detail: **Esc** clears selected skill; **/** focuses search (suppressed in text fields)
- [ ] Model picker: **Esc** closes the picker (intentionally fires even inside the search input)
- [ ] Stack ordering: with model picker open over inspector with a selected block — first Esc closes picker, second Esc closes block panel
- [ ] Regressions: Alt-held chat timestamps still expand to full date; chat tool-call menus still close on Esc; command-palette inputs still handle ⌘P / ⌘/ / ⌘K locally; agent-workbench textarea ⌘+Enter still submits

### DIFF
--- a/packages/ui/src/lib/drag-drop-zone.svelte
+++ b/packages/ui/src/lib/drag-drop-zone.svelte
@@ -8,6 +8,12 @@
   layout, positioning, and scoped-CSS context. Override with `class`
   or `style` if a real layout box is needed.
 
+  Callers that overlay absolutely-positioned children inside the zone
+  must set `position: relative` on the child themselves — `display:
+  contents` on the wrapper removes it as a containing block, so the
+  overlay would otherwise anchor to the nearest positioned ancestor
+  outside the zone.
+
   Multiple zones can coexist on the same page; each is scoped to its
   own subtree, so a drop on one never fires the other.
 
--- a/packages/ui/src/lib/drag-drop-zone.svelte
+++ b/packages/ui/src/lib/drag-drop-zone.svelte
@@ -1,0 +1,46 @@
+<!--
+  Element-scoped file drop zone.
+
+  Wraps a region with dragenter/over/leave/drop handlers and exposes a
+  live `dragOver` flag through the children snippet. The wrapping
+  `<div>` defaults to `display: contents` so it adds drop listeners
+  without introducing a layout box — children keep their existing
+  layout, positioning, and scoped-CSS context. Override with `class`
+  or `style` if a real layout box is needed.
+
+  Multiple zones can coexist on the same page; each is scoped to its
+  own subtree, so a drop on one never fires the other.
+
+  `onFiles` is read through the props proxy on every drop, so callers
+  can pass a fresh arrow each render without losing the latest closure.
+-->
+<script lang="ts">
+  import type { Snippet } from "svelte";
+  import { createDragDropState } from "./drag-drop.svelte.ts";
+
+  interface Props {
+    /** Called with dropped files. Filtering is the caller's job. */
+    onFiles: (files: File[]) => void;
+    /** Optional class on the wrapper element. */
+    class?: string;
+    /** Inline style override. Appended after the default `display: contents`. */
+    style?: string;
+    /** Children receive the live `dragOver` state. */
+    children: Snippet<[{ dragOver: boolean }]>;
+  }
+
+  const props: Props = $props();
+  const state = createDragDropState((files) => props.onFiles(files));
+</script>
+
+<div
+  class={props.class}
+  style="display: contents;{props.style ? ` ${props.style}` : ''}"
+  ondragenter={state.onDragEnter}
+  ondragover={state.onDragOver}
+  ondragleave={state.onDragLeave}
+  ondrop={state.onDrop}
+  role="presentation"
+>
+  {@render props.children({ dragOver: state.dragOver })}
+</div>
--- a/packages/ui/src/lib/drag-drop.test.ts
+++ b/packages/ui/src/lib/drag-drop.test.ts
@@ -98,9 +98,12 @@ describe("createDragDropState — dragOver lifecycle", () => {
     state.onDragLeave(asDragEvent(mockEvent({ files: [] })));
     state.onDragLeave(asDragEvent(mockEvent({ files: [] })));
     state.onDragEnter(asDragEvent(mockEvent({ files: [] })));
-    // After one matched enter we should still see dragOver true —
-    // proves the counter clamped at 0, not -2.
-    expect(state.dragOver).toBe(true);
+    state.onDragLeave(asDragEvent(mockEvent({ files: [] })));
+    // With the Math.max(0, …) clamp the counter rides 0 → 0 → 1 → 0, so
+    // the final leave clears dragOver. Without it the counter goes
+    // negative and the matched enter never brings it back to 0 — dragOver
+    // would stay stuck true.
+    expect(state.dragOver).toBe(false);
   });
 });
 
@@ -117,6 +120,15 @@ describe("createDragDropState — browser-required side effects", () => {
     const e = mockEvent({ files: [] });
     state.onDragOver(asDragEvent(e));
     expect(e.defaultPrevented).toBe(true);
+  });
+
+  it("ignores dragover for non-file drags", () => {
+    const state = createDragDropState(vi.fn());
+    const e = mockEvent({ nonFile: true });
+    state.onDragOver(asDragEvent(e));
+    // Non-file drags (text/url) must not be preventDefault'd, or the
+    // browser's own drag preview gets hijacked.
+    expect(e.defaultPrevented).toBe(false);
   });
 
   it("sets dropEffect = copy on dragenter and dragover", () => {
--- a/packages/ui/src/lib/drag-drop.test.ts
+++ b/packages/ui/src/lib/drag-drop.test.ts
@@ -1,0 +1,161 @@
+/**
+ * @vitest-environment happy-dom
+ */
+import { describe, expect, it, vi } from "vitest";
+import { createDragDropState } from "./drag-drop.svelte.ts";
+
+interface MockDataTransfer {
+  types: readonly string[];
+  files: ArrayLike<File> & Iterable<File>;
+  dropEffect: string;
+}
+
+interface MockEvent {
+  dataTransfer: MockDataTransfer | null;
+  defaultPrevented: boolean;
+  preventDefault(): void;
+}
+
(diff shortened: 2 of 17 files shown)


packages/ui/src/lib/hotkeys.svelte.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  function handle(e: KeyboardEvent) {
+    for (let i = bindings.length - 1; i >= 0; i--) {
+      const b = bindings[i];
+      if (!b) continue;
+      if (!matches(b, e)) continue;
+      if (b.when && !b.when(e)) continue;
+      if (b.preventDefault !== false) e.preventDefault();
+      b.handler(e);
```
Vpr99: A handler that throws propagates out of `handle()`, drops the current event silently (`preventDefault` already ran), and floods the console. Wrap, log, treat as consumed:

```ts
try { b.handler(e); } catch (err) { console.error("hotkey handler threw", err); }
```

Add a test mirroring the fallthrough style at `hotkeys.test.ts:39-50`: A (throws) on top of B, dispatch combo, assert B does NOT fire and the next event still dispatches. Three lines + one test — eliminates a class of "one buggy route breaks Esc everywhere" regressions.

packages/ui/src/lib/drag-drop.test.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  it("dragLeave does not underflow past zero", () => {
+    const state = createDragDropState(vi.fn());
+    state.onDragLeave(asDragEvent(mockEvent({ files: [] })));
+    state.onDragLeave(asDragEvent(mockEvent({ files: [] })));
+    state.onDragEnter(asDragEvent(mockEvent({ files: [] })));
+    // After one matched enter we should still see dragOver true —
+    // proves the counter clamped at 0, not -2.
+    expect(state.dragOver).toBe(true);
+  });
```
Vpr99, line 107: Named "does not underflow past zero" but the assertion passes either way. Without `Math.max(0, ...)` the counter goes to -2, then `enter` flips `dragOver = true` unconditionally — green.

Restructure as a differential: after `leave, leave, enter, leave` assert `dragOver === false`. With the clamp the counter rides at 0 → 1 → 0; without it, `leave` decrements to -1 and `dragOver` stays true. Same line count, actually catches the bug.

packages/ui/src/lib/drag-drop.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    state.onDragEnter(asDragEvent(e));
+    expect(e.defaultPrevented).toBe(true);
+  });
+
+  it("preventDefault on dragover (required for drop to fire)", () => {
+    const state = createDragDropState(vi.fn());
+    const e = mockEvent({ files: [] });
+    state.onDragOver(asDragEvent(e));
+    expect(e.defaultPrevented).toBe(true);
+  });
```
Vpr99, line 123: The `onDragOver` files-only filter (`drag-drop.svelte.ts:57`) is unguarded by tests. Remove the `hasFiles` check there and every test still passes — but behavior breaks subtly (text/url drags get `preventDefault`'d, hijacking drag previews). Five lines:

```ts
it("ignores dragover for non-file drags", () => {
  const state = createDragDropState(vi.fn());
  const e = mockEvent({ nonFile: true });
  state.onDragOver(asDragEvent(e));
  expect(e.defaultPrevented).toBe(false);
});
```

Vpr99 (commented):
Re-review after a Karpathy-triage pass on the [2026-05-21 review](https://github.com/friday-platform/friday-studio/blob/abstract-dragdrop-keyboard-registry/docs/reviews/2026-05-21-abstract-dragdrop-keyboard-registry.md). Six inline comments below — the strongly-worth-doing subset (in-scope, surgical, minimum code). The full list (plus three Needs-Decision items: `onDragLeave` files filter, `display: contents` browser floor, `register()` HMR-fragile `lastIndexOf`) lives in the prior review doc and the [2026-05-25 follow-up](https://github.com/friday-platform/friday-studio/blob/abstract-dragdrop-keyboard-registry/docs/reviews/2026-05-25-abstract-dragdrop-keyboard-registry.md).

packages/ui/src/lib/hotkeys.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    const removeSpy = vi.spyOn(window, "removeEventListener");
+    const reg = createHotkeyRegistry();
+    expect(addSpy).not.toHaveBeenCalledWith("keydown", expect.any(Function));
+    const dispose = reg.register({ key: "k", cmdOrCtrl: true, handler: () => {} });
+    expect(addSpy).toHaveBeenCalledWith("keydown", expect.any(Function));
+    dispose();
+    expect(removeSpy).toHaveBeenCalledWith("keydown", expect.any(Function));
+    addSpy.mockRestore();
+    removeSpy.mockRestore();
+  });
```
Vpr99, line 211: Confirms attach-on-first and detach-on-last but never re-registers after dispose. A bug where `detachListener()` forgets to null `listener` would still pass — `removeEventListener` fires, but the next `register()` would skip `ensureListener()` and silently no-op. Three-line append:

```ts
addSpy.mockClear();
track(reg.register({ key: "j", cmdOrCtrl: true, handler: () => {} }));
expect(addSpy).toHaveBeenCalledWith("keydown", expect.any(Function));
```

packages/ui/src/lib/hotkeys.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    const a = vi.fn();
+    const b = vi.fn();
+    track(reg.register({ key: "k", cmdOrCtrl: true, handler: a }));
+    track(
+      reg.register({ key: "k", cmdOrCtrl: true, when: () => false, handler: b }),
+    );
+    dispatch({ key: "k", metaKey: true });
+    expect(a).toHaveBeenCalledOnce();
+    expect(b).not.toHaveBeenCalled();
+  });
```
Vpr99, line 50: Fallthrough test has exactly two bindings. A refactor that turns `continue` into `break` inside the dispatch loop at `hotkeys.svelte.ts:71-79` wouldn't be caught. One new case pins the loop semantics:

```ts
it("walks past multiple disabled bindings to the first enabled match", () => {
  const reg = createHotkeyRegistry();
  const a = vi.fn(); const b = vi.fn(); const c = vi.fn();
  track(reg.register({ key: "k", cmdOrCtrl: true, handler: a }));
  track(reg.register({ key: "k", cmdOrCtrl: true, when: () => false, handler: b }));
  track(reg.register({ key: "k", cmdOrCtrl: true, when: () => false, handler: c }));
  dispatch({ key: "k", metaKey: true });
  expect(a).toHaveBeenCalledOnce();
  expect(b).not.toHaveBeenCalled();
  expect(c).not.toHaveBeenCalled();
});
```

packages/ui/src/lib/drag-drop-zone.svelte
```diff
@@ -0,0 +1,46 @@
+<!--
+  Element-scoped file drop zone.
+
+  Wraps a region with dragenter/over/leave/drop handlers and exposes a
+  live `dragOver` flag through the children snippet. The wrapping
+  `<div>` defaults to `display: contents` so it adds drop listeners
+  without introducing a layout box — children keep their existing
+  layout, positioning, and scoped-CSS context. Override with `class`
+  or `style` if a real layout box is needed.
```
Vpr99, line 9: Worth documenting the trap this PR itself hit: `display: contents` removes the wrapper as a containing block, so any absolutely-positioned overlay inside the zone won't anchor to it. Migrating chat's drop overlay outward forced `user-chat.svelte:1517-1525` to add `position: relative` to `.user-chat` and delete the previously-conditional `.user-chat.chat-drag-over { position: relative; }` rule.

One-line JSDoc addition so the next caller doesn't re-discover this:

> Callers that overlay absolutely-positioned children inside the zone must set `position: relative` on the child themselves — `display: contents` on the wrapper removes it as a containing block.

PR #416, opened 2026-05-21
feat: per-chat conversational model picker

LissaGreense:
## What

Adds a conversational-model picker pill to the chat input toolbar. Click it to pick a model for the current workspace; the next message uses that model. Reuses the existing settings `ModelPicker` modal unchanged.

When no override is set, the daemon's resolved global default applies (current behavior).

## How

- **Storage** — browser `localStorage`, keyed per workspace. No server-side persistence, no schema changes, no daemon restart.
- **Transport** — chat-send body carries an optional `model: "<provider>:<modelId>"`. The field is renamed at the SDK boundary into the agent session as `modelOverride`.
- **Agent** — `workspace-chat` resolves the override via a new `resolveModelFromString` helper that reuses the existing registry/credential machinery. Falls back to `platformModels.get("conversational")` when absent. A bad spec surfaces a clear error rather than silently falling back.
- **Picker reuse** — the modal's "save & unlock API key" flow is rejected here; per-chat overrides target already-credentialed providers, credential management stays in Settings.

## Scope

- Conversational role only
- Single model, no fallback chain (chain stays in global settings)
- Delegate calls within the turn still use the global model
- No cross-device sync — set the global default in Settings if you want that

## Test plan

**Automated coverage**

- `packages/llm/src/platform-models.test.ts` → `describe("resolveModelFromString")` covers real resolver behavior (format parsing, unknown providers, missing credentials, LITELLM_API_KEY fallback).
- `packages/system/agents/workspace-chat/workspace-chat.handler.test.ts` → "Per-turn model override" are **wiring tests** with `resolveModelFromString` mocked — they only assert that `session.modelOverride` flows into `streamText` and the message-metadata tagger. They are not end-to-end resolver tests.

**Manual smoke**

- [ ] Fresh workspace shows resolved global default on the pill
- [ ] Selecting a model updates the pill, persists, next message uses it
- [ ] Switching workspaces reflects that workspace's override (or its absence)
- [ ] "Use default chain" clears the override
- [ ] Bad spec produces an error, not a silent fallback


LissaGreense:
## QA cases tested

- Pill renders the global default in a fresh chat
- Click pill → settings `ModelPicker` modal opens, scoped to Conversational
- Select a model → pill updates, `localStorage["model-override-<wsId>"]` persisted
- Send a message with override → daemon logs confirm `modelOverride` flows to the agent, assistant UI badge shows the override model
- "Use default chain" → override cleared, pill back to default, localStorage key removed
- Override survives a fresh tab / reload of the same chat
- localStorage key scoped per workspace (verified by inspection)
- Registry-form provider (`openai.chat`) renders as catalog short-form (`OpenAI`) in the pill tooltip
- Bad spec via devtools → no error visible in chat UI (known gap, intentionally deferred — only reachable via tampering or stale credentials; client-side stale-detection in the pill is the cleaner future fix)
- localStorage unavailable → helper degrades silently (structural check)

8 pass, 1 known gap deferred.

LissaGreense:
image

odk- (approved):

PR #417, opened 2026-05-21
feat(chat): cross-workspace @-mentions

odk-:
Cross-workspace `@workspaceId/chatId` mentions in the chat composer, plus a `summarize_chat` tool so a new chat can compactly continue a context-maxed prior one.

### @-mentions
- **`read_chat` tool** — wraps `ChatStorage.getChat`. Both the platform MCP server (`packages/mcp-server/src/tools/chat/read.ts`) and the workspace-chat agent (`packages/system/agents/workspace-chat/tools/read-chat.ts`) expose it under the same name. The agent draws from its own AtlasTools registry rather than the MCP list, so both wrappers are needed. `friday-studio-vp0` / `friday-studio-om5`
- **Server-side resolver** (`apps/atlasd/src/chat-sdk/mention-resolver.ts`) — parses `@ws/chat` refs in incoming user messages, ACLs against `WorkspaceMemberStorage`, builds a frozen snapshot (title + first/last excerpt), persists it as a `data-mention-resolved` part, injects a hidden `<atlas-mention-context>` text part the model sees, and unions source workspaces into `foreground_workspace_ids`. Dedupes on `(workspaceId, chatId)` so client-side optimistic placeholders are replaced by the canonical server data on persist. `friday-studio-ivt`
- **Composer autocomplete** — Svelte popover anchored above the textarea (`mention-autocomplete.svelte`), fuzzy title match with recency tiebreak, sourced from a new `GET /api/me/chats` cross-workspace listing. Inserts the friendly `@Title` into the textarea; `expandMentionDisplayText` swaps it back to `@workspaceId/chatId` on submit. The composer also ships a client-side `data-mention-resolved` part with the picked title so the optimistic bubble renders the link immediately. `friday-studio-c7j`
- **Empty-state composer** on the workspace overview (`/platform/<ws>`) — same MentionAutocomplete wiring. Seed payload in sessionStorage is now JSON-shaped (`{ text, mentions }`) when mentions are picked; plain string otherwise for backward compat.
- **History rendering** — `splitMentions` renders each resolved token as a styled link to `/platform/<wsId>/chat/<chatId>`. Renderer skips the synthetic `mention-expansion` text part using the same `providerMetadata.atlas.kind` marker pattern as `attachment-expansion`.
- **Defensive ACL** in the JetStream backend so `ChatMetadata.workspaceId` is the source of truth, not the KV key prefix — paves the way for the storage refactor in the `friday-studio-1z9` chain. `friday-studio-be9`

### summarize_chat (continuation across context limits)
- **`POST /api/workspaces/:wid/chat/:cid/summarize`** — server-side map-reduce LLM summarization. Project messages → role-prefixed ledger, chunk by char budget (~6k tokens, with oversized-message split fallback), map each chunk via `smallLLM` (labels role), then reduce per-chunk summaries into a single bounded-output document (cap 1500 tokens). Single-chunk chats skip the reduce step. `friday-studio-6dq`
- **`CHAT_SUMMARIES` KV cache** — keyed on `workspaceId.chatId.updatedAt.focusHash`. A new message append advances `updatedAt` and naturally invalidates without an explicit purge. Write failures are non-fatal.
- **`summarize_chat` tool** — agent variant in `packages/system/agents/workspace-chat/tools/summarize-chat.ts` and MCP variant in `packages/mcp-server/src/tools/chat/summarize.ts`. Tool descriptions explicitly nudge the agent to pick `summarize_chat` (bounded output) over `read_chat` (raw transcript) for continuation use cases. `friday-studio-g6q` / `friday-studio-5te`

## Tests

51 new tests, all passing:
- `mention-resolver` — parse / snapshot / apply / merge-foreground / orchestrator + dedup (17)
- `read_chat` MCP tool (3) + `summarize_chat` MCP tool (3)
- `mention-text` shared parser — split / count / detect-query / score / expand (16)
- `/api/me/chats` route — merge / empty / filter / cap / 401 (5)
- `summarize-chat` pipeline — projectMessages / chunkLedger / orchestrator (9)
- `POST /:chatId/summarize` route — cache-miss / cache-hit / 404 / 400 / 503 (5)
- `storage` defensive ACL — three poisoned-KV cases (3)
- All existing `chat-sdk-instance` + `storage` + `workspace-chat` tests still green

## Deferred / not in this PR

- `friday-studio-mth` (P3) — inline banner when a send completes with skipped (unauthorized / not-found) mentions. Backend logs failures today but the UI doesn't surface them.
- `friday-studio-1z9` chain — drop `workspaceId` from JetStream stream/subject/upload names so reassigning a chat between workspaces is a metadata flip. Gated on a separate review checkpoint before backfill runs against persisted data.
- `friday-studio-qj6` — E2E continuation test for `summarize_chat` (manual QA tracked separately).

## Test plan

- [x] `deno task dev:playground`
- [x] Open a chat, type `@`, verify autocomplete shows accessible chats with workspace label
- [x] Pick a chat; verify the optimistic user bubble renders the link with the friendly title
- [x] Send; verify the agent reply references the snapshot
- [x] Mention a chat in a different workspace; verify that workspace's skills/agents become callable for the turn
- [x] Reload the page; verify the link still renders with the canonical server-side title
- [x] Empty-state composer at `/platform/<ws>`: type `@`, pick a chat, send — verify the optimistic bubble in the new chat renders the link with the friendly title
- [x] `curl -k -X POST https://localhost:5200/api/daemon/api/workspaces/<wid>/chat/<cid>/summarize -H "Content-Type: application/json" -d '{}'` — verify bounded output, `cached: false` on first call, `cached: true` on second
- [x] Append a message to the source chat; next summarize call: `cached: false` (updatedAt invalidates)
- [x] In a new chat, `@-mention` a long source chat and ask the agent to continue — verify it picks `summarize_chat` over `read_chat` and the response references the prior conversation coherently

### DIFF
--- a/apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
+++ b/apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
@@ -921,6 +921,7 @@ export function createMessageHandler(
       requesterUserId: userId,
       currentWorkspaceId: workspaceId,
       foregroundWorkspaceIds: initialForegroundWorkspaceIds,
+      exposeKernel: options?.exposeKernel ?? false,
     });
 
     const appendResult = await ChatStorage.appendMessage(chatId, storedMessage, workspaceId);
--- a/apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
+++ b/apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
@@ -916,9 +916,23 @@ export function createMessageHandler(
     // this turn. Unauthorized / not-found refs are logged + dropped so a
     // bad reference doesn't break the send. See friday-studio-ivt /
     // friday-studio-4j7.
+    // For the web adapter, `userId` (= message.author.userId) is the
+    // Atlas userId set via `X-Atlas-User-Id` on the inbound request.
+    // For Slack/Telegram/Discord/Teams/WhatsApp it's the *platform*
+    // user id (a Slack U… id, etc.), which never matches a row in
+    // WORKSPACE_MEMBERS — every mention would fail as 'unauthorized'.
+    // Fall back to the workspace owner (the user who connected the
+    // platform integration). Single-owner workspaces (the common
+    // case) see no behavior change; in shared workspaces this is a
+    // permissive proxy that still respects per-workspace membership
+    // (the owner has to be a member of the referenced workspace for
+    // the resolution to succeed). See friday-studio-phq.
+    const mentionRequesterUserId =
+      adapterName === "atlas" ? userId : (options?.ownerUserId ?? userId);
+
     const { message: storedMessage, foregroundWorkspaceIds } = await applyMentions({
       message: initialStoredMessage,
-      requesterUserId: userId,
+      requesterUserId: mentionRequesterUserId,
       currentWorkspaceId: workspaceId,
       foregroundWorkspaceIds: initialForegroundWorkspaceIds,
       exposeKernel: options?.exposeKernel ?? false,
--- a/apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
+++ b/apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
@@ -33,7 +33,11 @@ import {
   type PlatformCredentials,
 } from "./adapter-factory.ts";
 import { ChatSdkNotifier } from "./chat-sdk-notifier.ts";
-import { applyMentions } from "./mention-resolver.ts";
+import {
+  applyMentions,
+  extractMentionsFromMessage,
+  type MentionResolutionFailure,
+} from "./mention-resolver.ts";
 
 const logger = createLogger({ component: "chat-sdk-instance" });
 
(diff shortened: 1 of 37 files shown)


apps/atlasd/src/chat-sdk/mention-resolver.ts
```diff
@@ -0,0 +1,337 @@
...
+  for (const part of message.parts) {
+    if (part.type === "text" && typeof (part as { text?: unknown }).text === "string") {
+      const text = (part as { text: string }).text.trim();
+      if (text.length > 0) return text;
+    }
+  }
+  return "";
+}
+
+export function buildSnapshot(chat: Chat): {
```
Vpr99, line 138: **[Critical] Sanitize title + excerpts before embedding in `<atlas-mention-context>`.**

A hostile source-chat title or message body containing `</atlas-mention-context>` followed by override instructions lands inside the model's hidden context across workspace boundaries. The same risk applies in `summarize-chat.ts:199` and is already sanitized there via `replace(/[<>` + "`" + `]/g, "")`. The mention path carries user-controlled bytes cross-workspace and should be held to the same bar.

Mirror the strip on `title`, both excerpts, and the `id` interpolated into the opening tag in `renderMentionContextText`. Add one regression test in `mention-resolver.test.ts` with a hostile title.

Vpr99 (commented):
Review of cross-workspace @-mentions + summarize_chat. Architecture is sound; 8 actionable items below (full review at `docs/reviews/2026-05-25-chat-move-experiment.md`). One Critical (prompt-injection sanitization), the rest are small test/observability additions.

Also worth a separate decision before merge: the communicator-adapter `mentionRequesterUserId` fallback to `options.ownerUserId` at `chat-sdk-instance.ts:947` is a real ACL hole for shared Slack/Telegram channels — owner's foreign workspaces resolve under any teammate's send. Not commented inline because the fix isn't surgical (needs a real Atlas userId mapping or workspace flag). Worth a call: log-and-drop now, ship web-only mentions, or ship-as-is and gate behind a config?

apps/atlasd/src/chat-sdk/mention-resolver.ts
```diff
@@ -0,0 +1,337 @@
...
+  // server-side resolver is the sole authority on which mentions are
+  // valid; client placeholders are a render-time convenience that
+  // must not survive persistence. After stripping, we re-append a
+  // canonical part for each resolved ref below. See friday-studio-1ev:
+  //   - empty server-resolved set + client placeholders ⇒ all
+  //     placeholders dropped (forged refs / unauthorized refs / refs
+  //     whose @-token the user erased can no longer persist).
+  //   - non-empty set + matching placeholder ⇒ canonical replaces it
+  //     (one part out, one part in, same key).
+  const filteredParts = message.parts.filter((part) => !isMentionResolvedPart(part));
```
Vpr99, line 257: **[Test gap] Add a regression test for forged `data-mention-resolved` rejection.**

This filter is the only gate preventing clients from forging mention metadata that bypasses ACL. A future refactor that narrowed the predicate (e.g. only stripping parts matching a resolved ref) would break the security model silently. One test: ship a `data-mention-resolved` for an unauthorized chatId in the input message; assert the output contains zero `data-mention-resolved` parts.

apps/atlasd/src/chat-sdk/mention-resolver.test.ts
```diff
@@ -0,0 +1,369 @@
...
+    });
+
+    expect(out.resolved).toHaveLength(1);
+    expect(out.failures).toHaveLength(0);
+    expect(out.foregroundWorkspaceIds).toEqual(["ws-b"]);
+    expect(out.message.parts.some((p) => p.type === "data-mention-resolved")).toBe(true);
+  });
+
+  it("drops an unauthorized mention but keeps the message sendable", async () => {
+    getMemberMock.mockResolvedValue({ ok: true, data: null } as never);
```
Vpr99, line 396: **[Test gap] Swap the null stub for a role-mismatch stub.**

This tests `if mock returns null → unauthorized`, but the actual ACL gate is `MEMBER_ROLES.has(result.data.role)`. A regression that added `"viewer"` to the role set, or removed the role gate, would still pass. Add (or replace) with `getMemberMock.mockResolvedValue({ ok: true, data: { role: "viewer" } } as never)` and assert `unauthorized`. Pins the role set, not the null check.

apps/atlasd/src/chat-sdk/mention-resolver.test.ts
```diff
@@ -0,0 +1,369 @@
...
+
+    const out = await applyMentions({
+      message: userMessage("peek @ws-x/c-9"),
+      requesterUserId: "u-1",
+      currentWorkspaceId: "ws-a",
+      foregroundWorkspaceIds: undefined,
```
Vpr99, line 402: **[Test gap] Assert `foregroundWorkspaceIds` stays empty on unauthorized mention.**

The unauthorized test asserts `failures` and that the message is unaltered, but never asserts `foregroundWorkspaceIds` doesn't leak the unauthorized workspaceId. A regression where `mergeForegroundWorkspaceIds` was fed the failures list (or the unparsed refs) would leak `ws-x` into the turn's foreground without any test catching it. One additional `expect(out.foregroundWorkspaceIds).toBeUndefined()` line.

packages/core/src/chat/storage.test.ts
```diff
@@ -304,6 +305,69 @@ describe("ChatStorage (JetStream-backed)", () => {
...
   });
 
+  // Defense-in-depth ACL: ChatMetadata.workspaceId is the source of
+  // truth for chat ownership. Today the KV key prefix also encodes
+  // workspaceId, but the storage refactor (beads friday-studio-1z9)
+  // will drop that prefix. These tests pin the metadata-based check so
+  // the rejection path survives the refactor.
+  describe("metadata-based workspace ACL", () => {
+    const enc = new TextEncoder();
+    const poisonedMeta = (chatId: string, metaWorkspace: string) => ({
```
Vpr99, line 315: **[Test gap] Cross the defensive-ACL cases with `newNamingEnabled: true`.**

These poisoned-KV tests are well-constructed but use the default singleton backend. The whole point of the `1z9` chain that motivates this defense is the refactor that drops `workspaceId` from the KV key — so the defense MUST work under `newNamingEnabled: true`. Currently no test combines those two configs. Add 1-3 cases extending the existing poison pattern with a backend created via `createJetStreamChatBackend(nc, { newNamingEnabled: true })`.

packages/core/src/chat/summaries-storage.ts
```diff
@@ -0,0 +1,122 @@
+/**
```
Vpr99, line 1: **[Test gap] Add `summaries-storage.test.ts` against a real NATS test server.**

The route tests at `routes/workspaces/chat.test.ts:856-976` mock `ChatSummariesStorage.get/put` directly, so they only assert "mock returns null → compute; mock returns value → return." Nothing pins that two appends with different `updatedAt` produce different keys, that `focusHash` participates, or that chatIds that previously collided under the old sanitizer (e.g. `team.demo` vs `team_demo`) hash distinctly via SHA-256 — the literal invariants the PR description names.

Follow the pattern in `storage.test.ts`. 3-4 cases: (1) same key tuple = hit, (2) bump `updatedAtMs` = miss, (3) different `focusHash` = miss, (4) previously-colliding chatIds hash distinctly.

packages/system/agents/workspace-chat/tools/read-chat.ts
```diff
@@ -0,0 +1,90 @@
+/**
```
Vpr99, line 1: **[Test gap] Mirror the MCP-server tests against the agent-side `read_chat` and `summarize_chat` tools.**

This file (and `summarize-chat.ts` next to it) duplicates the trim/total/truncated logic from `packages/mcp-server/src/tools/chat/read.ts`. The MCP side has tests; the agent-side parallel wrapping a different transport (`fetch(getAtlasDaemonUrl())`) has none. Every sibling tool in this directory has a `.test.ts`. A future fix to one path silently leaves the other broken.

3 cases each (happy path, route-trim with `totalMessageCount`, 404).

apps/atlasd/src/chat-sdk/chat-sdk-instance.ts
```diff
@@ -908,10 +909,51 @@ export function createMessageHandler(
...
+    // Fall back to the workspace owner (the user who connected the
+    // platform integration). Single-owner workspaces (the common
+    // case) see no behavior change; in shared workspaces this is a
+    // permissive proxy that still respects per-workspace membership
+    // (the owner has to be a member of the referenced workspace for
+    // the resolution to succeed). See friday-studio-phq.
+    const mentionRequesterUserId =
+      adapterName === "atlas" ? userId : (options?.ownerUserId ?? userId);
+
+    const { message: storedMessage, foregroundWorkspaceIds } = await applyMentions({
```
Vpr99: **[Observability] Surface mention-resolution failures as structured logs.**

`applyMentions` returns `failures` (unauthorized / not_found / internal_error) but only `{ message, foregroundWorkspaceIds }` is destructured here — `failures` is dropped silently. The PR description correctly defers the user-facing `mth` UI work, but ship a structured log here (workspace, chatId, reason, count) so we have a signal to measure how often this fires in prod before the UI work starts. Cheap, doesn't expand scope, gives the deferred UX work a target.

PR #422, opened 2026-05-22
feat(evals): multi-model promptfoo suites via LiteLLM proxy

LissaGreense:
## Summary

- **New eval root at `tools/evals/promptfoo/`** runs daemon-free, parallel, multi-provider suites against a single LiteLLM proxy on `localhost:4000`. Promptfoo only knows alias names (`friday-sm` / `friday-md` / `friday-lg`); the proxy resolves them to backends. Adding a model is one line in `litellm_config.yaml` + one entry in `shared/providers.yaml` — suites need no edits.
- **Seven daemon-free promptfoo suites** end-to-end:
  - Text-only (stock providers): `progress-line`, `prompt-interpolation`, `agent-config-prompt`, `title-generation`.
  - Tool-capturing (custom Deno-worker provider): `workspace-chat-elicitation`, `workspace-chat-bundled-agent`, `workspace-chat-agent-type`. Migrated from the custom Deno harness without losing tool-call assertion fidelity.
- **Cross-runtime bridge for tool-capturing suites** — `shared/providers/deno-worker.cjs` (Node) spawns one long-lived Deno worker per handler path, JSON-Lines protocol on stdin/stdout. Each suite's `handler.ts` imports the real `@atlas/llm` registry + real `prompt.txt` + real `bundledAgents` registry + real tool descriptions and runs `streamText` with mock tool execute() closures that capture the model's decisions. No daemon, no real side effects.
- **Custom Deno harness in `tools/evals/lib/` survives only for `agents/web/web.eval.ts`** — it needs `TraceEntry[]` inspection for `snapshotBeforeInteract` / `errorRecovery` / `stepEfficiency` metrics promptfoo can't natively model.
- **Function-wrapping suites** (`prompt-interpolation`, `agent-config-prompt`) call the real Friday prompt-builder once at generation time and write a committed `tests.generated.yaml`. Regenerate with `deno task evals:render-promptfoo`.
- **Single entry point**: `deno task evals:promptfoo` runs every suite in parallel — each suite gets its own `promptfoo eval` invocation (per-suite is required because promptfoo's multi-config merge breaks `file://` var resolution). Per-suite JSON outputs are aggregated into a pass-rate table; the wrapper exits 100 only if the **aggregate** pass rate < `PROMPTFOO_PASS_RATE_THRESHOLD`.
- **Picking models**: `EVAL_TIER=small|medium|large` picks a provider subset via `--filter-providers` regex. Adding a tier is just tagging a new provider's label.

## Architecture (tool-capturing suites)

```
promptfoo (Node)
  ↓  callApi(prompt, context)
shared/providers/deno-worker.cjs (Node shim)
  ↓  spawn ONCE per handler path, JSON-Lines on stdin/stdout
shared/providers/worker.ts (Deno dispatcher)
  ↓  dynamic-import + per-request dispatch
suites/<name>/handler.ts (Deno)
  ↓  imports real @atlas/llm + real prompt.txt + real registry
@atlas/llm registry → (when LITELLM_API_KEY set) → LiteLLM proxy → real provider
```

Same production code path the real workspace-chat uses, plus mock tool execute() closures that capture call args instead of running real side effects.

## Live validation (full matrix, medium-tier)

```
suite                                pass/total   %
──────────────────────────────────── ──────────── ────────
agent-config-prompt                  9/9          100.0
progress-line                        33/36        91.7
prompt-interpolation                 14/15        93.3
title-generation                     13/15        86.7
workspace-chat-agent-type            8/10         80.0
workspace-chat-bundled-agent         4/6          66.7
workspace-chat-elicitation           4/4          100.0
──────────────────────────────────── ──────────── ────────
AGGREGATE                            85/95        89.5
```

Real signals surfaced by the multi-model matrix that the original Haiku/Sonnet-only evals couldn't produce:
- **Groq llama-3.1-8b is unreliable on tool-use** in elicitation (returns empty response, no tool calls).
- **Haiku falls back to `type:"llm"`** on bundled-agent decisions where Sonnet picks the right `type:"atlas"`.
- **csv-streaming is non-deterministic** even on Sonnet (the original eval's docstring already flagged this).

## How to run locally

```bash
# 1. Install LiteLLM (Python 3.13 required — 3.14 breaks uvloop)
uv tool install --python 3.13 'litellm[proxy]'

# 2. Start the proxy with whatever provider keys you have
export ANTHROPIC_API_KEY=...   # for friday-md, friday-lg
export GROQ_API_KEY=...        # for friday-sm
export LITELLM_MASTER_KEY=sk-friday-evals-dev
cd tools/evals/promptfoo/litellm && ./start.sh

# 3. Tell the @atlas/llm registry to route through the proxy
export LITELLM_API_KEY=sk-friday-evals-dev

# 4. Run every suite in parallel, see the pass-rate table
deno task evals:promptfoo

# 5. Or just the cheap CI tier (text-only suites; workspace-chat-* drop small)
EVAL_TIER=small deno task evals:promptfoo

# 6. CI shape — fail the run if aggregate pass rate < 85
PROMPTFOO_PASS_RATE_THRESHOLD=85 deno task evals:promptfoo

# 7. Regenerate tests.generated.yaml after editing cases in any render.ts
deno task evals:render-promptfoo

# 8. Custom harness still runs as before for the remaining web.eval.ts
deno task evals run
```

## Generic-config audit (no machine-local wiring)

- No hardcoded paths (`/Users/<name>`, `$HOME`, etc.) anywhere in `tools/evals/promptfoo/`.
- All path resolution is relative (`import.meta.dirname` / `__dirname`) or git-based (`git rev-parse --show-toplevel`).
- All API keys come from env vars; documented in `litellm/README.md`.
- `localhost:4000` is the documented LiteLLM proxy convention; overridable for the registry via `LITELLM_BASE_URL`.
- `LITELLM_MASTER_KEY` defaults to `sk-friday-evals-dev` in `litellm/start.sh` — overridable via env. Local-dev convenience only, never a real secret.

## Picking a model tier

Every provider entry in `shared/providers.yaml` (and each workspace-chat suite's inline providers) carries a `tier:<small|medium|large>` tag in its label:

```yaml
- id: openai:chat:friday-sm
  label: 'groq-8b (openai) | tier:small'
```

`EVAL_TIER` is forwarded as `--filter-providers 'tier:(<value>)'` (regex). The workspace-chat suites omit `tier:small` — Groq llama-3.1-8b's free-tier 6k TPM cap can't fit the 9k+-token workspace-chat system prompt. Re-enable when a higher-context small-tier provider lands.

## Gotchas documented inline

- Python 3.14 + uvloop incompat → pin to 3.13 via `uv tool install --python 3.13`.
- Promptfoo's `cost` assertion can't read LiteLLM's `x-litellm-response-cost` header → kept out of shared `defaultTest.yaml`; opt-in per-suite.
- Promptfoo passes assertion `value` through Nunjucks → bare `{{` / `}}` crash; use `not-regex` with escaped braces.
- JS `RegExp` has no inline `(?i)` flag → use character classes (`[Ii]`) for case-insensitivity in `regex`/`not-regex`.
- Promptfoo `-c configA -c configB` merge uses the **first** config's basePath for ALL configs' `file://` var resolution → multi-config invocation breaks suites that reference per-suite prompt files. That's why `run-all.sh` iterates per-suite.
- `--filter-providers` is a regex against id + label — the `tier:<size>` tag in each provider's label is what makes `EVAL_TIER=small` work.

## What is NOT in scope (deliberate)

- Migrating `web.eval.ts` — its scoring is built on `TraceEntry[]` inspection (`snapshotBeforeInteract`, `stepEfficiency`, `errorRecovery`) that promptfoo's stock output model doesn't expose. Would be a reimplementation, not a migration.
- GitHub Actions workflow wiring — env knobs (`EVAL_TIER`, `PROMPTFOO_PASS_RATE_THRESHOLD`) are real but no `.github/workflows/*.yml` references them yet.
- Re-enabling cost assertions — depends on either a different provider that surfaces cost in-band or a custom JS provider that reads the LiteLLM header.
- Re-enabling `tier:small` on workspace-chat suites — needs a higher-context small-tier model (Groq paid, OpenRouter, etc.).

## Test plan

- [x] `deno task test tools/evals/lib/` passes
- [x] `deno check` clean on `tools/evals/lib/*.ts`, `tools/evals/run.ts`, all `render.ts`, all workspace-chat `handler.ts`, `shared/providers/worker.ts`
- [x] `deno task evals list` discovers only `agents/web/web.eval.ts` (3 workspace-chat .eval.ts files migrated, dirs cleaned)
- [x] `deno task evals:render-promptfoo` regenerates both `tests.generated.yaml` files cleanly
- [x] `npx promptfoo validate config -c <each suite>` returns valid for all 7 suites
- [x] Live full-matrix run reproduces the 85/95 pass-rate table above


tools/evals/promptfoo/litellm/start.sh
```diff
@@ -0,0 +1,30 @@
...
+
+export LITELLM_MASTER_KEY="${LITELLM_MASTER_KEY:-sk-friday-evals-dev}"
+
+docker run --rm -i \
+  -p 4000:4000 \
+  -v "$PWD/litellm_config.yaml:/app/config.yaml:ro" \
+  -e ANTHROPIC_API_KEY \
+  -e GROQ_API_KEY \
+  -e LITELLM_MASTER_KEY \
+  ghcr.io/berriai/litellm:main-stable \
```
Vpr99: **`:main-stable` is a moving tag.** The recent `x-litellm-response-cost` → `x-litellm-response-cost-original` rename (fixed in 7db2f994) is exactly the failure mode this introduces. The handler now reads both spellings as a hedge — the next rename silently zeroes out cost.

Pin to a concrete tag (e.g. `:v1.86.0-stable`) or a `sha256:` digest, and document the bump cadence in `litellm/README.md`. Removes the load-bearing reason for the dual-header read.

tools/evals/promptfoo/shared/providers/deno-worker.cjs
```diff
@@ -0,0 +1,227 @@
...
+        "run",
+        "--allow-all",
+        "--unstable-worker-options",
+        "--unstable-kv",
+        "--unstable-raw-imports",
+        workerScript,
+        "--handler",
+        this.handlerAbsPath,
+      ],
+      { cwd: repoRoot(), env: process.env, stdio: ["pipe", "pipe", "inherit"] },
```
Vpr99, line 75: **Worker import failures surface as generic "exited unexpectedly."** With `stdio: [..., "inherit"]`, a Deno import error (bad path, missing dep, syntax error) sends its stack trace to the parent's stderr — but the rejection promptfoo sees is just `Deno worker exited unexpectedly (code=1 signal=null)`. Under concurrency, N pending calls all reject with identical noise and the operator has no signal pointing at the real cause.

Switch stderr to `pipe`, buffer the last ~100 lines, and prepend them to `exitErr` when rejecting pending. Still tee to `process.stderr` for live visibility. This is the single most likely "what just broke" debugging surface.

tools/evals/promptfoo/shared/providers/deno-worker.cjs
```diff
@@ -0,0 +1,227 @@
...
+    );
+
+    this.proc.stdout.on("data", (chunk) => {
+      this.buffer += chunk.toString("utf8");
```
Vpr99: **UTF-8 multi-byte boundary split.** `chunk.toString("utf8")` decodes per-chunk. If model output containing em-dashes / emoji crosses a chunk boundary mid-codepoint, you get `U+FFFD` and a corrupted JSON parse. The Deno side uses `TextDecoder` correctly; the Node side doesn't.

Use `require("node:string_decoder").StringDecoder("utf8")` and `decoder.write(chunk)`. Two-line fix, low-probability but the captured strings include arbitrary model output.

tools/evals/promptfoo/shared/providers/deno-worker.cjs
Vpr99, line 75: **`env: process.env` + `--allow-all` forwards every env key to the worker.** Local-eval calculated risk, but worth naming. The handler imports real `@atlas/llm` with full host privileges and inherits every env key the parent had (`ANTHROPIC_API_KEY`, `GROQ_API_KEY`, etc). A malicious or buggy prompt template flowing through real prompt-builder code into a handler-imported dep with `fs`/`fetch` access sees the lot.

Minimum: add a README warning that evals run with full host privileges; never paste untrusted prompt templates into a case file. Stretch: scope to `--allow-net=$LITELLM_HOST,<provider hosts>` + `--allow-read=<repo root>` and forward only an eval-key allowlist.

tools/evals/promptfoo/scripts/run-all.sh
```diff
@@ -0,0 +1,287 @@
...
+if ! curl -sf -m 2 http://localhost:4000/health \
+    -H "Authorization: Bearer ${LITELLM_MASTER_KEY}" >/dev/null; then
+  echo "✗ proxy not responding at :4000 (run tools/evals/promptfoo/litellm/start.sh)" >&2
+  echo "  see tools/evals/promptfoo/litellm/README.md for setup" >&2
+  exit 1
+fi
+
+FILTER_ARGS=()
+if [[ -n "${EVAL_TIER:-}" ]]; then
+  FILTER_ARGS+=(--filter-providers "tier:(${EVAL_TIER})")
```
Vpr99: **Unanchored regex over-matches.** `tier:(small)` substring-matches any future label containing `tier:small` (e.g. `tier:small-fast`). The README explicitly invites adding tiers; this is a contributor footgun.

Anchor against label whitespace: `'(^| )tier:(${EVAL_TIER})( |$)'`. Or validate `EVAL_TIER` against an allow-list before invoking promptfoo.

tools/evals/promptfoo/scripts/run-all.sh
```diff
@@ -0,0 +1,287 @@
...
+const trippedSuites = rows.filter(suiteFailed).map((r) => r.suite);
+const skippedSuites = rows.filter((r) => r.skipped).map((r) => r.suite);
+const ranSuites = rows.length - skippedSuites.length;
+// `requireSuites=0` opts out of the gate entirely (ad-hoc tier filtering).
+const requireGateFailed = requireSuites > 0 && ranSuites < requireSuites;
+if (trippedSuites.length > 0 || aggregateFailed || requireGateFailed) {
+  console.log("");
+  if (requireGateFailed) {
+    console.log(
+      `✗ only ${ranSuites}/${requireSuites} suites ran; SKIPPED: ${
```
Vpr99, line 306: **`EVAL_TIER` typo silently passes every suite as SKIPPED.** `EVAL_TIER=smell` filters every provider out; every suite reports zero results → all SKIPPED → aggregate is 0/0. The `PROMPTFOO_REQUIRE_SUITES` floor catches the count, but the error message doesn't hint at the filter typo.

When `EVAL_TIER` is set and `ranSuites === 0`, append: `hint: EVAL_TIER='${EVAL_TIER}' matched no provider labels (expected one of: small, medium, large)`. Cheap dev-experience improvement.

tools/evals/promptfoo/suites/workspace-chat-elicitation/handler.ts
```diff
@@ -0,0 +1,137 @@
...
+  // companion `-discount-amount`/`-margin-*` variants); earlier versions used a
+  // flat `x-litellm-response-cost`. Check both so the bridge keeps working
+  // across upstream upgrades. The AI SDK exposes HTTP response headers on
+  // `result.response.headers` for any HTTP-based provider that talks to
+  // LiteLLM. Forward to deno-worker.cjs which surfaces it as top-level `cost`
+  // on the promptfoo ProviderResponse.
+  const responseMeta = await result.response;
+  const costHeader =
+    responseMeta.headers?.["x-litellm-response-cost-original"] ??
+    responseMeta.headers?.["x-litellm-response-cost"];
```
Vpr99, line 133: **Cost-header read is shape-fragile + only wired in this suite.** Two issues stacked:

1. `responseMeta.headers?.["x-..."]` assumes plain-object access. AI SDK v6's `result.response.headers` is typed `Record<string, string>` for some providers but `Headers` for raw-fetch paths — bracket access on a `Headers` instance returns `undefined`. The NaN guards in 9d6d9428 catch bad strings but not the wrong-shape case.
2. Only this suite reads cost. `workspace-chat-bundled-agent` and `workspace-chat-agent-type` run `stepCountIs(20)` tool loops — exactly where a tool-thrash regression would 10× cost — and have no cost assertion. The threshold gate is armed on the cheapest suite.

Extract a `readCostHeader(headers)` helper into `shared/providers/` that handles both `Headers` and plain-object shapes, and add a `cost` assertion to the other two suites' `tests.yaml`. Surface a one-time stderr warning when both header names are absent so a third rename fails loud.

tools/evals/promptfoo/suites/workspace-chat-agent-type/assertions/check.js
```diff
@@ -0,0 +1,226 @@
...
+  if (upsertAgents.length === 0) return "no upsert_agent calls";
+  return upsertAgents
+    .map((u) => `${u.id}(type=${u.config.type || "?"}, agent=${u.config.agent || "?"})`)
+    .join(", ");
+}
+
+// 1. At least one upsert_agent matched the expected type (and agent, if given).
+function correctTypeChosen(output, context) {
+  const { captures, expectedType, expectedAgent } = parseContext(output, context);
+  const matched = captures.upsertAgents.find(
```
Vpr99, line 53: **`correctTypeChosen` passes if *any* upsert is right.** `.find(u => u.type === expectedType)` returns true even when the model emitted five agents with the right type plus four with garbage. `noWrongType` only catches explicitly-forbidden types (and only when `forbiddenTypesJson` is set). For most rows this is "at least one is right" — a weak quality signal that will keep passing as the model regresses.

Where case-vars specify a single expected agent, use `.every(...)` or add a `singleAgentEmitted` metric. May have been inherited from the original eval — the migration is the place to tighten it.

tools/evals/promptfoo/suites/workspace-chat-agent-type/promptfooconfig.yaml
```diff
@@ -0,0 +1,52 @@
...
+evaluateOptions:
+  showProgressBar: true
+  # csv-streaming-user is acknowledged non-deterministic (the same prompt
+  # produces opposite type-selection outcomes across sessions; see tests.yaml
+  # and the original agent-type-default.eval.ts docstring). `repeat` is only
+  # valid at the config level — there is no per-test `options.repeat` field,
+  # so we repeat the entire suite 3× and let per-metric aggregates absorb the
+  # flake. All 5 cases × 2 providers now run 3×, so expect ~3× spend on this
+  # suite vs. a single pass.
+  repeat: 3
```
Vpr99, line 52: **Suite-wide `repeat: 3` for one flaky case.** A single acknowledged-flaky case forces 3× repeat across the entire suite — 30 rows instead of 10, ~3× cost. A previously-stable case regressing from 3/3 → 2/3 averages into the metric and the regression is invisible.

Either move `csv-streaming-user` to a separate quarantine suite (scoped `repeat: 3`), or add a per-case PASS-rate to the `run-all.sh` aggregator so 3/3 → 2/3 surfaces. Per-case visibility is the cheaper option; suite split can wait for the second flaky case.

tools/evals/promptfoo/shared/assertions/not-a-refusal.ts
```diff
@@ -0,0 +1,33 @@
...
+ * `not-regex` for genuine alternations. JS RegExp has no inline `(?i)` flag,
+ * so case-insensitivity rides on character classes (`[Ii]`).
+ *
+ * @param metric  Promptfoo metric name to attribute these assertions to.
+ *                Defaults to "NotARefusal" — change it if a suite wants to
+ *                bucket separately in reports.
+ */
+export function notARefusalAsserts(metric = "NotARefusal"): Array<Record<string, unknown>> {
+  return [
+    { type: "not-regex", value: "missing (required )?input", metric },
```
Vpr99: **Pattern is case-sensitive on the most-likely-uppercased branch.** Other patterns in this file use explicit `[iI]` / `[nN]` classes; this one is bare `"missing (required )?input"`. The canonical refusal phrasing starts capitalized ("Missing required input…") and won't match — making the assertion dead code on its primary input.

`"[mM]issing (required )?input"`. One character.

tools/evals/promptfoo/suites/progress-line/assertions/status-line.js
```diff
@@ -0,0 +1,36 @@
...
+// Mirrors statusLineScore() in tools/evals/agents/small-llm/small-llm.eval.ts.
+//
+// Promptfoo calls this with a default export receiving `(output, context)`.
+// Return a {pass, score, reason} object; promptfoo records the score under the
+// `metric:` name set on the assert.
+
+const FAIL_PATTERNS = [
+  /^I('m| am| cannot| can't| don't)/i,
+  /^(Some|Here|The|There|This|These|Those)\s/i,
+  /\d+\.\s+\*?\*?[A-Z]/i,
```
Vpr99: **Two over-broad fail patterns reject legit output.**

1. Line 10's `/^(Some|Here|The|There|This|These|Those)\s/i` fails "These are titanium gravel bike options" on a leading determiner alone, regardless of whether the rest is on-spec.
2. Line 11's `/\d+\.\s+\*?\*?[A-Z]/i` is positional-agnostic *and* redundant — any `1. X` requires a newline, which already fails on line 13.

Combined with the 50-char ceiling, these over-fit the rubric to the original eval's phrasing rather than its intent. Drop the redundant numbered-list pattern; soften the determiner check to require a non-progress-verb continuation. Test against the four research cases that currently bypass `StatusLine` entirely (see other comment on `tests.yaml`).

Vpr99 (commented):
Drive-by code review. 14 inline comments for findings worth acting on — full writeup in `docs/reviews/2026-05-27-evals-multi-model-litellm.md` on the branch. No critical bugs; mostly operability/observability polish and a couple of assertion landmines. Verdict: **Needs Work**.

Top-of-list: `:main-stable` LiteLLM tag is the failure mode the cost-header dual-read is hedging against — pinning to a concrete tag removes the load-bearing reason for that hedge.

tools/evals/promptfoo/suites/progress-line/tests.yaml
```diff
@@ -0,0 +1,155 @@
...
+# ─── Research — progress updates ─────────────────────────────────────────────
+
+- description: 'research — AI safety'
+  vars:
+    system_prompt: file://prompts/research.txt
+    input: 'stage: analyzing, context: AI safety developments'
+  assert:
+    - type: icontains-all
+      value: [ai, safety]
+      metric: Relevance
```
Vpr99, line 75: **Research cases skip the StatusLine assert entirely.** All four research cases (lines 66-100) assert `Relevance` only. They claim to test progress updates but don't enforce the ≤50-char / -ing-verb contract. A model emitting "Today is sunny and my favorite color is blue. anthropic" passes the Anthropic-meeting case.

Add `StatusLine` to the four research cases, or rename them to clarify they're keyword-relevance only.

tools/evals/promptfoo/suites/workspace-chat-bundled-agent/tests.yaml
```diff
@@ -0,0 +1,52 @@
...
+# `output.captures.upsertAgents`) instead of re-parsing in every expression.
+
+- description: 'scrape-hacker-news — daily HN headlines should pick bundled web agent'
+  vars:
+    user_prompt: 'I want a workspace that scrapes the top headlines from Hacker News every morning.'
+  options: &parse_output
+    transform: 'JSON.parse(output)'
+  assert:
+    - type: javascript
+      value: "output.captures.upsertAgents.some(u => u.config.type === 'atlas' && u.config.agent === 'web')"
```
Vpr99, line 19: **No upper bound on emitted agents.** All three row-level asserts here (and at lines 36 and 51) use `upsertAgents.some(u => ...)`. A model emitting `atlas/web` + `atlas/slack` + `atlas/gh` + a spurious `llm` agent passes every metric. The agent-type suite has `noSpuriousUserAgent`; bundled-agent has no equivalent over-emission guard — which is the failure mode this suite is most likely to expose at 66.7%.

Add a `noExtraneousAgents` metric (`upsertAgents.length <= N` per case-var). More likely to *help* surface the model-behavior gap than to add false negatives.

tools/evals/run.ts
```diff
@@ -63,7 +63,7 @@ const runCommand = define({
 evals run
 
 # Run a specific eval file
-evals run -t tools/evals/agents/small-llm/small-llm.eval.ts
+evals run -t tools/evals/agents/workspace-chat/elicitation-behavior.eval.ts
```
Vpr99: **Example references a deleted eval.** `agents/workspace-chat/elicitation-behavior.eval.ts` was deleted in eed49e7b. Next user copy-pasting the example gets file-not-found.

Change to `agents/web/web.eval.ts` (the only surviving custom-harness eval).

LissaGreense: Fixed in 2d45629 — repointed to `tools/evals/agents/web/web.eval.ts` (and fixed the same stale ref in `tools/evals/CLAUDE.md`). <3

tools/evals/promptfoo/shared/assertions/not-a-refusal.ts
LissaGreense: Fixed in 2fff827 — `[mM]issing (required )?input`. Thanks for catching, that branch was dead coverage on the capitalized phrasing.

tools/evals/promptfoo/shared/providers/deno-worker.cjs
LissaGreense: Fixed in c9eb037 — switched the Node reader to `StringDecoder`. Also fixed the Deno side (`worker.ts`), which had the same bug: `decoder.decode(chunk)` was missing `{ stream: true }`. Thanks for catching.

tools/evals/promptfoo/scripts/run-all.sh
LissaGreense: Fixed in c46e321 - anchored the filter with `$` so it can't substring-match a longer label, plus added up-front `EVAL_TIER` validation against the small/medium/large allow-list (splitting the `|` alternation). Thanks!

tools/evals/promptfoo/scripts/run-all.sh
LissaGreense: Done in c46e321 - went fail-fast instead of a post-hoc hint: an unknown `EVAL_TIER` token now exits immediately before any suite runs, naming the bad value and the valid tiers (small/medium/large).

tools/evals/promptfoo/litellm/start.sh
LissaGreense: Pinned in d0cf735 — `ghcr.io/berriai/litellm:v1.86.2`. That's the exact release `:main-stable` resolves to right now (verified same manifest digest `sha256:cae1ac3…`), so it's zero behavioral change today while making eval cost/behavior reproducible. Documented the bump procedure + the cost-header re-check in litellm/README.md. Kept the dual-header read — it's still load-bearing for older cached images and the `friday-local`/Ollama path, so it's not coupled to the tag.

tools/evals/promptfoo/shared/providers/deno-worker.cjs
LissaGreense: Added the README warning in e8d2f7c (promptfoo/README.md) — calls out that evals run with full host privileges, inherit the shell env / API keys, and that tool-capturing suites exec `handler.ts` under `--allow-all`. Went with the warning rather than scoping the Deno permissions: the untrusted `prompt` is fed to the model as data, not into `fs`/`fetch` args, so there's no untrusted-input→exec path — and `@atlas/llm`'s provider registry can hit several hosts + the Deno cache, so an allowlist would be a maintenance treadmill for near-zero gain on a local-only tool. Happy to revisit the scoping if you disagree.

tools/evals/promptfoo/suites/progress-line/assertions/status-line.js
LissaGreense: Dug into this — landed split: kept the numbered-list pattern, cut the determiner check (and its sibling openers). Done in 30cc640.

**Numbered-list pattern — kept, it's load-bearing.** The premise "any `1. X` requires a newline, which already fails on line 13" isn't true:
- `/\d+\.\s+\*?\*?[A-Z]/i.test("1. Buy milk")` → `true`
- `/\n/.test("1. Buy milk")` → `false`

Line 13 doesn't catch a single-line list item; this pattern does. It also catches `Reviewing 2. Options`, which would otherwise **pass** the `^[A-Z][a-z]+ing\b` gate. Redundant with neither — keeping it.

**Determiner check — cut, though not for the reason given.** It doesn't reject *legit* output: the positive -ing gate already fails `These are titanium…` regardless (`/^[A-Z][a-z]+ing\b/` → `false`). But you're right it wasn't earning its place — with no `threshold` on the assert, `pass` gates and the score is cosmetic, so the opener blocklist (this plus the sibling `^I'm…` / `^Unfortunately…` patterns) only moved an already-doomed output's reported score 0 → 0.7 without changing pass/fail. Dropped all three; kept the two patterns (numbered list, newline) that catch -ing-starting-but-still-broken output. Net contract: leading -ing verb + ≤50 chars + no list/newline.

tools/evals/promptfoo/shared/providers/deno-worker.cjs
LissaGreense: Fixed in e38186e. Went lighter than pipe+buffer: the stack already reaches the terminal via inherited stderr — the real gap was *correlation* (under `-j 20` every pending row rejected with the same causeless message). So on a non-zero exit the rejection now appends a hint pointing at the Deno stack above ("worker died before responding; see the stack trace on stderr above — usually a handler import/syntax error"). Keeps the free live-stderr that `inherit` gives, no ring-buffer/tee/interleaving surface. If per-row stderr capture turns out to be worth it later we can layer the pipe+buffer on top.

tools/evals/promptfoo/suites/progress-line/tests.yaml
LissaGreense: Fixed in 3a9f658 — but not by adding `StatusLine`. The four research cases use `prompts/research.txt`, which mandates *"Maximum 4 words, start with an active verb"* — **not** the leading `-ing` verb `status-line.js` enforces. So reusing StatusLine would false-fail valid output like `Analyze AI safety progress` (bare verb) or `Investigating quantum computing trends` (38 chars, fine by word-count). Instead added a research-tailored `research-format.js` assert (single line, ≤4 words, leading capitalized verb) under a `ResearchFormat` metric, alongside the existing keyword Relevance check. That closes the hole you flagged — your `"Today is sunny … anthropic"` example now fails (10 words) — without over-fitting the research cases to a contract they don't have.

(History note: research cases were Relevance-only in the original `small-llm.eval.ts` too; only web-search had `statusLineScore`. The migration added the -ing gate to the claude-code tool cases, which is why the asymmetry looked accidental.)

tools/evals/promptfoo/suites/workspace-chat-agent-type/promptfooconfig.yaml
LissaGreense: Validated this against the pinned promptfoo source (`0.121.12`). The cost math is exact, but the proposed fixes hit a tooling wall and "invisible" isn't quite right.

**Per-case / scoped `repeat` isn't possible in promptfoo 0.121.12.** `repeat` is only valid at `EvaluateOptionsSchema` (config-level) and as the CLI `--repeat` flag. The per-test `TestCase.options` schema is `…spread… .catchall(z.any())`, so a per-case `options.repeat: 3` is silently swallowed — no error, no effect. In the evaluator it's a global multiplier:
```js
for (let repeatIndex = 0; repeatIndex < (options.repeat || 1); repeatIndex++)
  for (const vars of varCombinations) { … }
```
It wraps every row; there's no per-test scoping hook. That silent `.catchall` swallow is exactly why we hoisted it to `evaluateOptions` in c3464eb (the per-case form had been a dead field). So "scoped `repeat: 3`" can't be done per-case — a quarantine means a whole separate suite dir.

**Cost: confirmed exact** — 5 cases × 2 providers × 3 = 30 rows vs 10.

**"Regression is invisible" — overstated for this aggregator.** `run-all.sh` gates on row-level `results.stats.successes/failures/errors`, not per-metric `namedScores`. A case regressing 3/3 → 2/3 moves the gated pass rate 30/30 → 29/30 = **−3.3pp** — visible to the `PROMPTFOO_SUITE_FLOOR`/`AGGREGATE_CEILING` gate (just not enough for one row to trip 70/85). The "averages into the metric" concern is real for `namedScores`, but the aggregator doesn't gate on those.

That said — your underlying worry is legit: a single-row regression is under-weighted in a 3×-repeated suite. The clean fix isn't a quarantine suite (we already weighed + rejected the comparable `tags:flaky` + exclusion route in c3464eb for the maintenance cost), it's a **per-case PASS-rate in the run-all.sh aggregator**: the per-row JSON already carries `vars` + `namedScores`, so grouping by case to surface 3/3 → 2/3 directly is buildable and low-effort. Tracking that as a follow-up rather than splitting the suite.

tools/evals/promptfoo/suites/workspace-chat-elicitation/handler.ts
LissaGreense: Split outcome — pushed back on sub-claim 1, implemented sub-claim 2 (in 5b208de).

**Sub-claim 1 (Headers-vs-object helper): declining.** In `ai@6.0.184`, `result.response.headers` is typed `Record<string, string>` (`dist/index.d.ts:156`) and `@ai-sdk/provider-utils` does `Object.fromEntries([...headers])` before exposing it — it's *always* a plain object, never a `Headers` instance. A `readCostHeader` helper with a Headers branch would be dead code, so I left the bracket access as-is.

**Sub-claim 2 (cost gate on the heavier suites): done, but not by mirroring elicitation.** Mirroring it would've been hollow: `result.response.headers` is the **last step's** cost only, so on a `stepCountIs(20)` loop a tool-thrash regression (whose final step can be cheap) wouldn't show. Instead I price `result.totalUsage` (cumulative across all steps) via `@atlas/llm` `tokensToCost` in a shared `litellm-cost.ts` helper, and added a `cost` assert to both agent-type and bundled-agent. Bonus: it sidesteps the churny `x-litellm-response-cost` header name entirely. Thresholds (2.00 / 1.20) are conservative placeholders sized for friday-lg — flagged in-code to tighten after a calibration run; they catch a gross blowup today, not a fine regression.

tools/evals/promptfoo/suites/workspace-chat-agent-type/assertions/check.js
LissaGreense: Fixed in de2a63c. Added an opt-in `agentSetMatches` metric driven by a new `expectedAgentsJson` case-var: it fails on an unexpected-type extra (the "right + garbage" case you flagged) or a missing expected type. It no-ops when a case doesn't declare a set, so the acknowledged non-deterministic `csv-streaming-user` case stays loose. Matching is by type (+ agent slug when given), so several agents of an expected type still pass — only genuine wrong-type/garbage fails, which avoids `.every()`'s false-fails on legitimate multi-agent rows. Kept `correctTypeChosen` as a soft "at-least-one" sub-signal for the per-metric report.

tools/evals/promptfoo/suites/workspace-chat-bundled-agent/tests.yaml
LissaGreense: Fixed in 6b56446. Added the same opt-in `agentSetMatches` metric (mirrors the agent-type suite) to the two targeted cases via `expectedAgentsJson` — `atlas/web` and `atlas/slack` — so the right bundled agent plus a spurious extra (e.g. a stray `type:llm`) now fails on the unexpected element. The generic `fetch-and-summarize` smoke case stays opted out (still `AnyAtlasAgent`) since its exact set isn't inferable. Existing `Bundled*AgentChosen` asserts kept as soft signals.

Vpr99 (approved):

PR #442, opened 2026-05-25
feat(llm): local OpenAI-compatible provider (LM Studio / Ollama / vLLM)

LissaGreense:
## Summary

Adds a `local` provider to `@atlas/llm` for self-hosted OpenAI-compatible servers — LM Studio, Ollama, vLLM, llama.cpp. Same pattern as OpenRouter (`@ai-sdk/openai` with a custom baseURL), but gated on a base URL rather than an API key because for localhost servers the URL **is** the credential.

- **Backend (`packages/llm`)** — new `local.ts` provider, registered in both direct and LiteLLM-mode registries (LiteLLM passthrough is intentionally skipped — localhost can't route through a hosted proxy). `hasCredential("local")` checks `LOCAL_BASE_URL`. Catalog fetcher hits `${LOCAL_BASE_URL}/models` for live model discovery against whatever LM Studio currently has loaded. Trailing slashes are normalized.
- **UI (`tools/agent-playground`)** — model-picker locked banner branches on `provider === "local"` to render a text URL input (not a password input) with a `http://localhost:1234/v1` placeholder and LM Studio / Ollama help copy. Backend `saveApiKey(envVar, value)` is generic and unchanged.
- **Tests** — catalog fetcher tests cover LM Studio happy path, trailing-slash normalization, unset env var, server-down error surfacing, and the "LITELLM_API_KEY does not unlock local" semantic. Plus a construction smoke test for `createLocalWithOptions`. 116/116 tests passing in `packages/llm`. Workspace typecheck and Svelte typecheck both clean.

## How it works

Set `LOCAL_BASE_URL` (via the daemon `.env` file or through the Settings → Models picker's "Save & unlock" banner) and Friday will:

1. Show a `Local` provider pill in the model picker, credentialed.
2. Hit `${baseURL}/models` to discover whatever model LM Studio / Ollama currently has loaded — no hardcoded model list.
3. Route any `local:<model-id>` selection through `@ai-sdk/openai` pointed at the local server, using the Chat Completions API (LM Studio/Ollama don't speak the Responses API).

`LOCAL_API_KEY` is optional and defaults to the literal string `"local"` — most local servers ignore auth entirely, but `@ai-sdk/openai` requires a non-empty key string.

## Test plan

Manually verified end-to-end against a real LM Studio instance (`google/gemma-4-e4b` loaded on port 1234):

- [x] **Case 1 — Locked Local pill renders in catalog.** `GET /api/config/models/catalog` shows `local` with `credentialConfigured: false`, `credentialEnvVar: "LOCAL_BASE_URL"`, empty models, no error.
- [x] **Case 2 — Banner shows text URL input (not password).** Picker → locked banner renders `type="text"` input with `placeholder="http://localhost:1234/v1"`, copy reads "Local models are locked until we know where your OpenAI-compatible server is running", help mentions LM Studio + Ollama.
- [x] **Case 3 — Paste URL → save → catalog refreshes.** "Save & unlock" flow flipped the Local pill to "connected", catalog discovered `google/gemma-4-e4b` (chat) and `text-embedding-nomic-embed-text-v1.5` (embedding). `LOCAL_BASE_URL` persisted to daemon `.env`.
- [x] **Case 4 — Trailing slash normalized.** Saving `http://localhost:1234/v1/` (trailing slash) discovers the same models — no double-slash artifacts in the outbound `/models` request.
- [x] **Case 5 — Chain picker writes `local:` to friday.yml.** Selecting `google/gemma-4-e4b` for the planner role and clicking "Save models" wrote `planner: 'local:google/gemma-4-e4b'` to `friday.yml`. The slash in the model id round-trips through `parseModelId` correctly (split at first colon only).
- [x] **Case 7 — End-to-end chat round-trip.** Direct registry call (`registry.languageModel("local:google/gemma-4-e4b")` → `generateText`) round-tripped through `@ai-sdk/openai` → LM Studio in ~5s. Response: _"I am Gemma 4, an open weights Large Language Model developed by Google DeepMind."_ Token usage parsed correctly (25 input / 232 output / 210 reasoning).
- [x] **Case 8 — Bad URL surfaces actionable error.** Setting `LOCAL_BASE_URL=http://localhost:9999/v1` (nothing running) surfaced `error: "tcp connect error: Connection refused (os error 61)"` in the catalog entry, `modelCount: 0`, `credentialConfigured: true` (env IS set, fetch just failed).

### Not validated in this pass

- **Case 6 (per-chat model pill)** — N/A. Plan referenced UI from an unmerged PR; not in current `main`.
- **Case 9 (server-down during session) + Case 10 (model swap)** — deferred. Same underlying mechanism as case 8 (catalog re-fetch surfaces error).

### Pre-existing finding (not a regression, worth a follow-up)

`GET /api/config/models` shows `resolved` stuck on the default chain after live edits to `friday.yml` — `createPlatformModels` captures `userConfig` at daemon-boot time, so changes need a daemon restart to take effect. Affects any `models:` change, not just local. Filing separately.

## Notes / non-goals

- **Multi-endpoint not supported.** One local server at a time; if we ever want `local-1` / `local-2` we'd want to redesign around workspace variables, not add more env vars.
- **No model catalog hardcoding.** Discovery is dynamic — whatever LM Studio has loaded is what shows up. Means a model that's not currently loaded won't appear in the picker until the user loads it and refreshes the catalog.
- **Catalog cache TTL is 1h.** Mutating `LOCAL_BASE_URL` via the env editor calls `invalidateCatalog()` so refreshes are immediate after the env-PUT path. Manual model swaps in LM Studio require a forced refresh (no-op env-PUT or daemon restart).

### DIFF
--- a/packages/llm/src/local.test.ts
+++ b/packages/llm/src/local.test.ts
@@ -1,0 +1,51 @@
+import process from "node:process";
+import { afterEach, describe, expect, it } from "vitest";
+import { createLocalWithOptions } from "./local.ts";
+
+/**
+ * Smoke test for the constructor — purely shape-level. The actual HTTP
+ * traffic isn't exercised here (no MockAgent / no real server); we rely on
+ * `@ai-sdk/openai` (a third-party SDK) to honor the `baseURL` we pass.
+ * Matches the repo precedent: `openrouter.ts` ships with zero tests.
+ */
+describe("createLocalWithOptions", () => {
+  const envBackup: Record<string, string | undefined> = {};
+  function withEnv(vars: Record<string, string | undefined>) {
+    for (const k of Object.keys(vars)) {
+      envBackup[k] = process.env[k];
+      if (vars[k] === undefined) delete process.env[k];
+      else process.env[k] = vars[k];
+    }
+  }
+  afterEach(() => {
+    for (const k of Object.keys(envBackup)) {
+      if (envBackup[k] === undefined) delete process.env[k];
+      else process.env[k] = envBackup[k];
+    }
+    for (const k of Object.keys(envBackup)) delete envBackup[k];
+  });
+
+  it("returns a provider when explicit options are passed", () => {
+    const provider = createLocalWithOptions({
+      baseURL: "http://localhost:1234/v1",
+      apiKey: "test",
+    });
+    // The provider is a callable function that also exposes `.chat()`.
+    expect(typeof provider.chat).toBe("function");
+    expect(typeof provider.languageModel).toBe("function");
+  });
+
+  it("reads baseURL and apiKey from env vars when options are omitted", () => {
+    withEnv({ LOCAL_BASE_URL: "http://localhost:11434/v1", LOCAL_API_KEY: "ollama" });
+    const provider = createLocalWithOptions();
+    expect(typeof provider.chat).toBe("function");
+  });
+
+  it("does not throw when LOCAL_BASE_URL is unset (falls back to LM Studio default)", () => {
+    withEnv({ LOCAL_BASE_URL: undefined, LOCAL_API_KEY: undefined });
+    // Construction must not throw — the registry instantiates this provider
+    // eagerly even for users who never use it. Credential gating elsewhere
+    // ensures it's only *invoked* when LOCAL_BASE_URL is set.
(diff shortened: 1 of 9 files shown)


LissaGreense:
image

odk- (approved):

PR #445, opened 2026-05-25
fix(playground): render run-detail results as markdown, not escaped JSON

### LINKED ISSUE #420: Run-detail page (`/platform/<wsId>/sessions/<sessionId>`) renders artifact output as raw JSON — markdown is shown as an escaped string
## What

When an FSM action's output is an `LLMResult` carrying markdown (or any artifact ref), the run-detail page shows it only as a syntax-highlighted JSON blob. Markdown headings, lists, and inline emphasis arrive as escaped characters inside a JSON string — users read the structural shape of the result instead of the result itself.

## Where

`tools/agent-playground/src/routes/platform/[workspaceId]/sessions/[sessionId]/+page.svelte:165-168` — the "Complete" block dumps `query.data.results` straight through `JsonHighlight`:

```svelte
{:else if query.data.results && Object.keys(query.data.results).length > 0}
  <FormattedData copyText={JSON.stringify(query.data.results, null, 2)} maxLines={7}>
    <JsonHighlight code={JSON.stringify(query.data.results, null, 2)} />
  </FormattedData>
{/if}
```

No branch checks the result type. No artifact refs are followed. "See output" just expands the same JSON.

## Places to investigate / reuse

The chat surface already renders this content well; the pieces are in tree:

- `artifactQueries.byId(id)` — `tools/agent-playground/src/lib/queries/artifact-queries.ts:48-69` — fetches metadata + inline contents.
- `ArtifactCard` — mounted from `tool-call-card.svelte:473-484` for the `display_artifact` tool — handles `text/markdown` natively (`artifact-card.svelte:73,140-141,220-229`), plus tabular previews and download fallback.
- `MarkdownBody` — `lib/components/chat/markdown-body.svelte` — the bare renderer if the card chrome isn't wanted on this surface.

LissaGreense:
## Summary

- The final "Complete"/"Failed" roll-up on `/platform/<wsId>/sessions/<sessionId>` dumped `query.data.results` through `JsonHighlight` regardless of shape. LLM markdown landed as an escaped JSON string — users read the structural shape of the result instead of the result itself.
- Now branches per entry in the results map based on what the FSM session reducer at `packages/core/src/session/session-reducer.ts:90` actually emits:
  - `{ text: "..." }` (workspace-chat / `"unknown"` agent fall-through — the dominant real-world shape) → markdown via `splitMarkdownByTables` + `MarkdownRendered` / `TableView`
  - `{ response: "...", data?: ... }` (Agent SDK `complete({ response, data })`) → response as markdown, optional `data` as a `JsonHighlight` block below
  - anything else → the original `JsonHighlight` fallback
- Renderer is wrapped in `FormattedData` (the same chrome used by `Input`, tool call request/response panels, and the existing fallback) so it has a Copy button that copies the raw markdown source, a 20-line clamp + Expand, and consistent padding against the StepBlock header.
- Per-agent label is only shown when there's more than one entry, so single-result roll-ups stay clean.

Bare-string and `{ error }` branches that were in the first cut were dropped — the session reducer doesn't emit either into this map (errors flow through `query.data.error` to the parent page's `sessionError` branch, which is untouched).

Scope intentionally limited to the roll-up at lines 165-168 named in the issue. The per-block `AgentBlockCard` output renderer at `agent-block-card.svelte:213-214` has the same JSON-dump shape but is a separate surface — flagged for follow-up, not bundled.

Closes #420

## Test plan

- [ ] Daemon + playground: load a finished `handle-chat` run (`results.unknown = { text: "..." }`). Markdown headings / lists / inline code render as styled prose; Copy button is present top-right; no escaped `\n` in the output.
- [ ] Load a run whose result is `{ response, data }` (Agent SDK `complete()` with both fields). Response renders as markdown, `data` collapses into a separate JsonHighlight block below.
- [ ] Load a run whose result contains an embedded GFM table inside a `{ text }` or `{ response }` markdown payload. Table renders via `TableView` between markdown prose chunks.
- [ ] Load a run with multiple agents in the results map (FSM with multiple completing agents). Each gets a small uppercase agent-name header above its content.
- [ ] Load a run with an unrecognised object shape (neither `{response}` nor `{text}`). Falls back to the existing `JsonHighlight` view.
- [ ] Failed-session error rendering (`sessionError` branch in the parent page) is unchanged.

## QA notes

- Live verified on `melted_onion` sessions `b459e082` (`{text}`) and `846c24a3` (`{response}`).
- SSR tests in `session-results.test.ts` cover 9 cases: empty results, `{text}`, `{response, data}`, `{response}` alone, embedded GFM table, unrecognised shape, multi-agent, single-agent header suppression, null value.

### DIFF
--- a/tools/agent-playground/src/lib/components/session/index.ts
+++ b/tools/agent-playground/src/lib/components/session/index.ts
@@ -1,3 +1,4 @@
 export { default as AgentBlockCard } from "./agent-block-card.svelte";
 export { parseError } from "./parse-error.ts";
+export { default as SessionResults } from "./session-results.svelte";
 export { StepBlock } from "./step-block/index.ts";
--- a/tools/agent-playground/src/lib/components/session/session-results.svelte
+++ b/tools/agent-playground/src/lib/components/session/session-results.svelte
@@ -1,16 +1,25 @@
 <!--
   Renders the final per-agent results map that lands on the run-detail
   "Complete"/"Failed" roll-up. `query.data.results` is typed as
-  `Record<string, unknown>` because agents can emit any shape via
-  `complete()`, FSM `outputTo`, or the chat-tool's implicit `{ text }`
-  fall-through. This component does the runtime branching so markdown
-  prose, structured `{response, data}` payloads, `{text}` envelopes,
-  error envelopes, and raw JSON fallbacks all land in the right
-  renderer.
+  `Record<string, unknown>` but in practice the FSM session reducer
+  (`packages/core/src/session/session-reducer.ts:90`) only ever assigns
+  `block.output` — an *object* — keyed by `block.agentName`. Two shapes
+  carry markdown prose in real sessions, the rest fall back to a JSON
+  view:
 
-  Lifts the splitMarkdownByTables + MarkdownRendered + TableView
-  pattern from the fullscreen artifact markdown view
-  (`/artifacts/[id]/markdown/+page.svelte`).
+    - `{ response, data? }` — Agent SDK `complete({ response, data })`
+                              per `writing-workspace-jobs/SKILL.md:715`.
+    - `{ text }`            — workspace-chat / "unknown" agent
+                              fall-through (the dominant shape).
+
+  Bare strings and `{error}` envelopes are intentionally NOT handled
+  here: the reducer never produces a string output, and session-level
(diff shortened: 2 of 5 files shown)


LissaGreense:
Before
image


After
image

odk- (approved):

PR #448, opened 2026-05-28
feat(launcher): one-click diagnostic log export

### LINKED ISSUE #324: Friday launcher: one-click compress & export diagnostic logs
## Problem

When something goes wrong in Friday, grabbing logs for diagnostics is a manual hunt — finding the right directories, zipping them up, and attaching to a bug report. Friction discourages users from sharing logs at all.

## Proposal

Add a menu bar item in the Friday launcher (next to existing actions like **Restart all** / **Open in browser**) that compresses the relevant log directories into a single archive and reveals it in Finder (or copies the path to clipboard).

Something like:

- **Export diagnostic logs…** → produces e.g. `~/Downloads/friday-diagnostics-2026-05-14-1530.zip` and opens Finder with the file selected.

## Scope (rough)

- Bundle daemon logs, agent logs, recent workspace state, and version info into a single archive.
- Strip or redact obvious secrets where feasible (env values, tokens) — minimum viable: include a README in the zip noting what was collected.
- Surface in the launcher menu bar with the same UX pattern as Restart all / Open in browser.

## Why

Lowers the bar to filing useful bug reports — one click instead of "go find the logs, then zip, then attach."

LissaGreense:
## Summary

Adds a one-click diagnostic log export to the Friday launcher (Closes #324). A new tray item builds a single zip — logs, launcher state, pids, and a structured manifest — and reveals it in the OS file browser, so users filing bug reports don't have to hunt for log directories. Power users can optionally embed their workspace definition bundles.

Implemented as four layers behind a clean `diagnostics.Export` seam:

- **Daemon `/api/version` route** (`apps/atlasd/routes/version.ts`) — exposes `getVersionInfo()` over HTTP so the export can stamp the daemon build into the manifest.
- **`diagnostics` Go package** (`tools/friday-launcher/diagnostics/`) — `Export(opts) (zipPath, err)` builds the zip with atomic temp+rename, `~/Downloads` resolution with a writable-fallback, a YAML manifest, and graceful degradation: any daemon/HTTP failure records a skip token and the export still succeeds.
- **Workspace bundling** — when the user picks "Logs + workspaces" and the daemon is reachable, embeds `GET /api/workspaces/bundle-all?mode=definition` verbatim as `workspaces.zip`. Failures map to stable tokens (`daemon_unreachable`, `daemon_returned_5xx`, `auth_refused`, `bundle_all_timeout`, `bundle_all_too_large`).
- **Tray UI** (`tools/friday-launcher/tray.go`) — an `Export diagnostics` parent with two explicit sub-actions, `Logs only` and `Logs + workspaces`; `" Exporting…"` menubar title, live phase labels (logs → workspaces → packaging) on the parent, CAS-guarded against double-clicks, success/failure label flips, and reveal-in-Finder.

### Menu shape

```
Open in browser
Restart all
View logs
Export diagnostics ▸          ← new
   Logs only
   Logs + workspaces          ← greys out when daemon down
─────────
Start at login                ← unchanged
─────────
Quit
```

## Manifest

A YAML file describing **provenance + the negative space**, not a content listing — what's *in* the archive is read from the zip's own file tree (`unzip -l`), so the manifest can't drift from the bytes:

```yaml
# privacy header (logs are NOT redacted — review before sharing)
daemon_version: <string | unreachable>
os / arch / generated_at
include_workspaces_requested: <bool>
skipped:
  - what: workspaces
    why: <stable token>
```

`why` is a stable enum: `user_opted_out`, `daemon_unreachable`, `daemon_returned_5xx`, `auth_refused`, `bundle_all_timeout`, `bundle_all_too_large`, `downloads_unwritable`.

## Notable details

- **No persisted preference** — workspace inclusion is an explicit per-export choice (`Logs only` vs `Logs + workspaces`), not a remembered checkbox. `Logs + workspaces` is daemon-gated (greyed with a hint when the daemon is down); `Logs only` is always available.
- **Loopback TLS** — the daemon HTTP client mirrors the readiness probe's `InsecureSkipVerify` (private-CA s2s cert, loopback only).
- **Timeouts** — 5s for `/api/version`, 60s for `/bundle-all`, via `context.WithTimeout` covering connect+TLS+body.
- **Body cap** — the `/bundle-all` response is capped via `io.LimitReader`; an over-cap bundle records `bundle_all_too_large` and the export ships logs-only.

## Test plan

- [x] `go test -race ./tools/friday-launcher/...` — diagnostics package (manifest golden, zip layout, logs filter, Downloads fallback, ProgressFn order, all HTTP skip-reason paths incl. HTTPS via `httptest.NewTLSServer`, bundle byte-cap) + tray (CAS guard under blocking export, shutdown gate, label state machine, deadlock watchdog, menu-mutation race test)
- [x] `apps/atlasd/routes/version.test.ts` — 200 + `getVersionInfo()` shape (Zod-parsed)
- [x] `gofmt` / `go vet` clean; `deno check` / `deno lint` clean on changed TS
- [x] Headless end-to-end: built the launcher from-branch and drove `Export(opts)` against a live daemon — daemon up (workspaces embedded, real `dev-<sha>` version stamp), daemon down (`daemon_unreachable` skip, logs-only), opted-out (`user_opted_out`); confirmed the trimmed manifest + zip layout
- [x] Manual GUI: click each sub-action in the menu bar; confirm Finder reveal, the `" Exporting…"` progress UX, and `Logs + workspaces` greying out when the daemon is down

## Notes

Built via a parallel agent team; each layer is an independently buildable commit. Two refinements landed after review:

- The workspace opt-in moved from a persisted `Include workspaces` checkbox to an explicit `Export diagnostics ▸ {Logs only | Logs + workspaces}` submenu — deleting the persisted `IncludeWorkspaces` state (and the `--autostart` read-modify-write that existed only to protect it).
- The manifest dropped its redundant `included:` content listing; the zip's own file tree is the source of truth for what's present.

The "Start at login" item was briefly dropped during the tray rewrite and restored — it was never in scope for this feature.

### DIFF
--- a/tools/friday-launcher/diagnostics/export.go
+++ b/tools/friday-launcher/diagnostics/export.go
@@ -22,7 +22,6 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
-	"sort"
 	"strings"
 	"time"
 )
@@ -141,18 +140,15 @@ func writeBundle(tmpPath string, opts ExportOptions, outDirSkipped bool) (err er
 	zw := zip.NewWriter(f)
 
 	progress(opts.ProgressFn, "logs")
-	logNames, err := addLogs(zw, sourceLogsDir())
-	if err != nil {
+	if err := addLogs(zw, sourceLogsDir()); err != nil {
 		return err
 	}
 
-	stateIncluded, err := addStateJSON(zw, filepath.Join(sourceStateDir(), "state.json"))
-	if err != nil {
+	if err := addStateJSON(zw, filepath.Join(sourceStateDir(), "state.json")); err != nil {
 		return err
 	}
 
-	pidsIncluded, err := addPids(zw, filepath.Join(sourceStateDir(), "pids"))
-	if err != nil {
+	if err := addPids(zw, filepath.Join(sourceStateDir(), "pids")); err != nil {
 		return err
 	}
 
@@ -171,7 +167,7 @@ func writeBundle(tmpPath string, opts ExportOptions, outDirSkipped bool) (err er
 
 	progress(opts.ProgressFn, "packaging")
 
-	m := buildManifest(opts, logNames, stateIncluded, pidsIncluded, outDirSkipped, daemonVersion, wsResult)
+	m := buildManifest(opts, outDirSkipped, daemonVersion, wsResult)
 	body, err := marshalManifest(m)
 	if err != nil {
 		return fmt.Errorf("marshal manifest: %w", err)
@@ -188,7 +184,7 @@ func writeBundle(tmpPath string, opts ExportOptions, outDirSkipped bool) (err er
 // buildManifest assembles the manifest struct from inputs gathered
 // upstream. Pure assembly — every HTTP call has already happened by
 // the time we get here.
-func buildManifest(opts ExportOptions, logs []string, stateIncluded, pidsIncluded, outDirSkipped bool, daemonVersion string, ws workspaceResult) manifest {
+func buildManifest(opts ExportOptions, outDirSkipped bool, daemonVersion string, ws workspaceResult) manifest {
 	skipped := []manifestSkip{}
 	if !opts.IncludeWorkspaces {
 		skipped = append(skipped, manifestSkip{
@@ -213,13 +209,7 @@ func buildManifest(opts ExportOptions, logs []string, stateIncluded, pidsInclude
 		Arch:                       runtime.GOARCH,
 		GeneratedAt:                nowFn(),
 		IncludeWorkspacesRequested: opts.IncludeWorkspaces,
-		Included: manifestIncluded{
-			Logs:       logs,
-			StateJSON:  stateIncluded,
-			Pids:       pidsIncluded,
-			Workspaces: ws.body != nil,
-		},
-		Skipped: skipped,
+		Skipped:                    skipped,
 	}
 }
 
@@ -397,16 +387,15 @@ func joinDaemonURL(base, path string) (string, error) {
 
 // addLogs copies every live .log file from logsDir into zw's logs/
 // subdir. "Live" = filename ends in exactly .log (no .gz, no .log.N).
-// Returns the sorted list of basenames included (for the manifest).
-func addLogs(zw *zip.Writer, logsDir string) ([]string, error) {
+// A missing logs dir is not an error — the zip just has no logs/.
+func addLogs(zw *zip.Writer, logsDir string) error {
 	entries, err := os.ReadDir(logsDir)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return nil, nil
+			return nil
 		}
-		return nil, fmt.Errorf("read logs dir: %w", err)
+		return fmt.Errorf("read logs dir: %w", err)
 	}
-	var names []string
 	for _, e := range entries {
 		if e.IsDir() {
 			continue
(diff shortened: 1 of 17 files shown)


LissaGreense:
Screenshot 2026-05-29 at 02 51 30

tools/friday-launcher/diagnostics/export.go
```diff
@@ -0,0 +1,547 @@
...
+
+// writeBundle writes every entry into the zip at tmpPath.
+//
+// zip.Writer.Close MUST run before the file's Close — that's when the
+// central directory is flushed. The file Close happens via a named-
+// return defer so a real error always wins over a close error, but
+// a close-only failure still surfaces (instead of being silently
+// swallowed like a bare `defer f.Close()` would).
+func writeBundle(tmpPath string, opts ExportOptions, outDirSkipped bool) (err error) {
+	f, err := os.Create(tmpPath) //nolint:gosec // G304: launcher-resolved output dir (Downloads/TempDir/explicit), no user input
```
odk-: here it can use `os.Create(filepath.Clean(tmpPath))` instead of adding nolint exception.

tools/friday-launcher/diagnostics/export.go
```diff
@@ -0,0 +1,547 @@
...
+		name := e.Name()
+		if err := addFileFromDisk(zw, filepath.Join(pidsDir, name), "pids/"+name); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func addFileFromDisk(zw *zip.Writer, srcPath, zipName string) error {
+	src, err := os.Open(srcPath) //nolint:gosec // G304: launcher-controlled paths under $HOME/.friday/local (logs/state/pids)
```
odk-: same `filepath.Clean`

tools/friday-launcher/diagnostics/export.go
```diff
@@ -0,0 +1,547 @@
...
+		return false
+	}
+	probe, err := os.CreateTemp(dir, ".friday-write-probe-*")
+	if err != nil {
+		return false
+	}
+	_ = probe.Close()
+	_ = os.Remove(probe.Name())
+	return true
+}
```
odk-, line 532: This doesn't really check if file can be removed as error from removal is never checked.

tools/friday-launcher/diagnostics/manifest.go
```diff
@@ -0,0 +1,70 @@
...
+
+// marshalManifest prepends the privacy header to a yaml-encoded body.
+func marshalManifest(m manifest) ([]byte, error) {
+	body, err := yaml.Marshal(m)
+	if err != nil {
+		return nil, err
+	}
+	var buf bytes.Buffer
+	buf.WriteString(privacyHeader)
+	buf.Write(body)
```
odk-: silent error.

tools/friday-launcher/openers.go
```diff
@@ -50,3 +51,38 @@ func openInFileBrowser(path string) error {
...
+	// suppression has the same rationale.
+	var cmd *exec.Cmd
+	switch runtime.GOOS {
+	case "darwin":
+		cmd = exec.Command("open", "-R", path) //nolint:gosec // G204: launcher-produced path
+	case "windows":
+		cmd = exec.Command("explorer.exe", "/select,"+path) //nolint:gosec // G204: launcher-produced path
+	default:
+		cmd = exec.Command("xdg-open", filepath.Dir(path)) //nolint:gosec // G204: launcher-produced path
+	}
```
odk-, line 83: Most probably `filepath.Clean` allows to get rid of nolint comments

tools/friday-launcher/diagnostics/export.go
LissaGreense: Thanks, fixed in c178d114.

tools/friday-launcher/diagnostics/export.go
LissaGreense: Done in c178d114.

tools/friday-launcher/diagnostics/export.go
LissaGreense: Good catch, fixed in c178d114 - removal is now part of the writable() verdict.

tools/friday-launcher/diagnostics/manifest.go
LissaGreense: Thanks, addressed in c178d114.

tools/friday-launcher/openers.go
LissaGreense: These are G204 (subprocess), not G304 — filepath.Clean does not satisfy them (gosec still fires `Subprocess launched with variable`). Kept the nolint here.

odk- (approved):

PR #464, opened 2026-06-02
feat(settings): add configurable image-model picker (Image role)

LissaGreense:
## Why

Image generation was the only model layer users couldn't configure. The agent hardcoded `google:gemini-3.1-flash-image-preview` — a Google preview ID that can be retired without notice — and ignored capability differences between providers (DALL·E 3 is gen-only, GPT Image 1.5 needs a verified org, Imagen takes `aspectRatio` not `size`). Swapping the model meant editing source code; trying to edit an image on a gen-only model produced silent nonsense.

## What

- **Image picker in Settings.** New "Image" row alongside the four LLM roles, with primary + fallback chain. Each option shows capability badges (`gen ✓` always, `edit ✓` / `edit ✗`) and any provider note (e.g. "Requires a verified OpenAI organization." on GPT Image 1.5).
- **Stable default.** Replaces the `-preview` ID with `google:gemini-2.5-flash-image`. One-time migration banner for upgrade users; fresh installs see nothing.
- **Capability-aware agent.** When the chosen model can't satisfy the request (edit prompt on gen-only model), the agent returns a clean error pointing back to Settings instead of producing nonsense.
- **Fail-fast configuration.** A typo in `friday.yml`'s `models.image` now fails daemon boot with the list of known model ids, not at first image-gen attempt.
- **Validation harness.** `scripts/validate-image-models.ts` exercises every overlay entry against live providers (direct + LiteLLM proxy) with a cost preview and y/N gate — operators can confirm what actually works in their deployment.

## QA path

Spin up `deno task dev:playground` and walk through:

1. **Settings → Models** shows 5 role cards. Image card defaults to "Gemini 2.5 Flash Image".
2. **Override the default.** Click the Image tile → picker opens with 6 entries grouped by provider. Verify badges (gen/edit) render and the GPT Image 1.5 note appears.
3. **Locked provider.** If you don't have `OPENAI_API_KEY`, the OpenAI section shows the inline "Save & unlock" banner.
4. **Persist.** Pick DALL·E 3 → Save → reload the page → tile still shows DALL·E 3. Check `friday.yml` — `models.image: openai:dall-e-3` should be written.
5. **Capability mismatch.** From a chat, attach an image and ask to edit it. With DALL·E 3 selected, you should get a clean error mentioning "supports generation only" and pointing to Settings. Switch to GPT Image 1.5 (or Gemini), retry — the edit should succeed.
6. **Default flow.** Clear `models.image` from Settings ("Use default chain"). Generate an image from chat — output uses Gemini 2.5 Flash Image.
7. **Migration banner.** Delete the `image-picker-intro-seen` key from `localStorage` and reload Settings. Banner appears in the Image row. Dismiss → reload → banner stays gone.

## Notes

- Validation harness is manual / cost-gated; not wired into CI. Fixtures captured by the harness drive a `(model × transport)` test matrix that skips cleanly when fixtures are absent.
- Out of scope (deferred to follow-ups): mask/inpainting, multi-image generation, additional providers, per-workspace defaults overrides, quality evals.


LissaGreense:
image

